PR #1, opened 2015-03-04
sync with Kodi PR 6566

opdenkamp:
...and some fixes


opdenkamp:
removed `sstream` include and added avahi discovery (see https://github.com/xbmc/xbmc/pull/6597)

PR #6, opened 2015-03-08
Update included headers

AlwinEsch:
Here the same fault as on VNSI and also changed the "kodi/util/atomic.h" to "kodi/threads/atomics.h".


AlwinEsch:
Ups, has not seen before that it was already performed, becomes updated then to only included header update.

PR #8, opened 2015-03-12
add timer conflict state for latest tvheadend

Glenn-1990:


Jalle19:
Thanks. Will keep this around for a while in case someone submits something else, so we don't have to bump the version and changelog all the time.

If anyone else feels it's needed right now, feel free to merge.

ksooo:
I'm wondering what this PR fixes or what the functinal change in Kodi is. If I understand right, it makes that "conflicting" recordings do no longer show up in the list of completed recordings in Kodi, right?

But, parsing plain text - which is definitely not part of the public tvh api and therefore might change at any point in time - is always error-prone and we should only merge this if it adds any *real* value to Kodi.

tvh seems not to have a clean way to expose the "conflict" case to the clients. From my point of view this should be addressed in tvh first and our addon then adapted to support the new tvh feature.

Glenn-1990:
@ksooo 

so, the problem
older tvheadend: active timers that aren't currently recording are showed as "recording"
new tvheadend: active timers that aren't currently recording are missing in the "upcoming recordings" list
--> This happens when there is no free tuner, no signal, ... to record the timer.

Now, they shouldn't be hided from the list for newer tvheadend versions, they should show "error" or "conflict"
They are hided because of the new error field https://github.com/tvheadend/tvheadend/blob/master/src/htsp_server.c#L715

I agree that comparing the string is not ideal, but this has been done before, see "missing"
https://github.com/kodi-pvr/pvr.hts/blob/master/src/Tvheadend.cpp#L1426

And if the string changes in tvheadend, xbmc will still show these timers as "error" instead of "conflict error", that's still better than not showing these timers, right?

ksooo:
I consider this no clean solution, sorry. The fact that it has been done bad before does not make it better. 

Your PR actually changes two things:
1) https://github.com/kodi-pvr/pvr.hts/pull/8/files#diff-3d7c5d1c125334f96674ced9f9ac2cdfR1440: parsing the string and setting PVR_TIMER_STATE_CONFLICT_NOK - this is what I consider no good.

2) https://github.com/kodi-pvr/pvr.hts/pull/8/files#diff-3d7c5d1c125334f96674ced9f9ac2cdfR542: setting PVR_TIMER_STATE_ERROR whenever "error" field is not empty. - I'm fine with this.

2) alone lets the conflicting timer appear in the "upcoming recordings" list with status "error"
1) is just cosmetics and not really needed to solve the actual problem with the not appearing recording

Thus, my suggestion is: Remove 1) and keep 2) in this PR.

Glenn-1990:
@ksooo 

Now it will show conflicting timers as "error" instead of running or hiding,
so it's already an improvement ;-)

ksooo:
Cool. That's what I expected.

Now let me suggest one last thing: Move the code that sets PVR_TIMER_STATE_ERROR if error is not empty to CTvheadend::ParseRecordingAddOrUpdate(), like this

<pre>
/* Error */
if ((str = htsmsg_get_str(msg, "error")) != NULL)
{
  UPDATE(rec.error, str);

  if (!strcmp(str, "300"))
  {
    UPDATE(rec.state, PVR_TIMER_STATE_ABORTED);
  }
  else
  {
    UPDATE(rec.state, PVR_TIMER_STATE_ERROR);
  }
}
</pre>

The path checking for "missing" can be removed. No longer needed.

Glenn-1990:
@ksooo 

When doing that, Istimer() will return false
https://github.com/kodi-pvr/pvr.hts/blob/master/src/HTSPTypes.h#L161

Adding " PVR_TIMER_STATE_ERROR" to "istimer()" is not really an option too as "PVR_TIMER_STATE_ERROR" can be used also for recordings too.
So basically completed recordings (with error) will show up in the timer list.

ksooo:
Could you please explain a little more exactly, what the actual behavior with the patch I suggested is (what appers where or not) and how this differs from your expectation.

ksooo:
Good point about <code>IsTimer()</code>. I think the solution could be to implement this more clever, without the problematic PVR_TIMER_STATE. Something with time - a recording must be something in the past - or so. Not sure yet though...

ksooo:
@Glenn-1990 if #9 works for you, this pr could be closed, imo. Please report back.

ksooo:
Closing this pr in favour to #9

PR #9, opened 2015-03-14
Improve handling of timers and recordings with error.

ksooo:
@glenn-1990 could you please test. This is an alternative approach for #8

### DIFF
--- a/src/HTSPTypes.h
+++ b/src/HTSPTypes.h
@@ -25,7 +25,7 @@
 #include <algorithm>
 #include <vector>
 #include <map>
-#include "kodi/util/StdString.h"
+#include <string>
 #include "client.h"
 
 typedef enum {
@@ -142,7 +142,8 @@ struct SRecording
     startExtra = 0;
     stopExtra  = 0;
 
-    state   = PVR_TIMER_STATE_ERROR;
+    state      = PVR_TIMER_STATE_NEW;
+
     title.clear();
     description.clear();
     error.clear();
@@ -161,7 +162,8 @@ struct SRecording
   bool IsTimer () const
   {
     return state == PVR_TIMER_STATE_SCHEDULED ||
-           state == PVR_TIMER_STATE_RECORDING;
+           state == PVR_TIMER_STATE_RECORDING ||
+           state == PVR_TIMER_STATE_ERROR;
   }
 };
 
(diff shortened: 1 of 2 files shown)


src/HTSPTypes.h
```diff
@@ -151,17 +152,41 @@ struct SRecording
...
+    {
+      // determine type by state.
+      return state == PVR_TIMER_STATE_COMPLETED ||
+             state == PVR_TIMER_STATE_ABORTED   ||
+             state == PVR_TIMER_STATE_RECORDING;
+    }
+    else
+    {
+      // determine type by time (both timers and recordings may have errors).
+      return IsStopTimeInThePast();
```
Glenn-1990: Should be 'IsStartTimeInThePast()', otherwise running recordings will not show up in the "recordings" window.

ksooo: The problem with showing items with an error in the recordings listing is - at least in Confluence theme - that there is no "status" row. One cannot see that a recording has an error...

But hey, then we should make things even more simple and display everything that has state PVR_TIMER_STATE_ERROR in the timer listing - withdraw the time-based decisions in IsTimer() and IsRecording()? Everything that has an error will be a timer - as per definition. What do you think?

Glenn-1990: If using "PVR_TIMER_STATE_ERROR" only for timers, then "IsStopTimeInThePast()" is not needed anymore. I'm fine with that ;-)

ksooo: Okay. Will change that.

BTW, could you do me the favour and close #8?

ksooo:
@Glenn-1990 Changes are in as discussed. Please test and report back.

Glenn-1990:
keep in mind that "PVR_TIMER_STATE_ERROR" is the default 
https://github.com/kodi-pvr/pvr.hts/blob/master/src/HTSPTypes.h#L145

Glenn-1990:
Also "PVR_TIMER_STATE_ERROR" is used for missing and invalid recording (so not only timers)

Glenn-1990:
@ksooo 
this is working for me and based on your comments
https://github.com/kodi-pvr/pvr.hts/pull/12

ksooo:
@Glenn-1990 Changed this pr to reflect what I think you wanted to achieve with #12. Please test and report back. #12 may be closed imo.

ksooo:
Seeing no way to implement this in a clean way. tvheadend does not offer a suitable api at the moment. This should be addressed first. Closing this PR.

PR #12, opened 2015-03-17
add timer error state for latest tvheadend

Glenn-1990:


Glenn-1990:
"PVR_TIMER_STATE_ABORTED" is deleted as this doesn't have any effect.

ksooo:
Thanks. You're right. Setting PVR_TIMER_STATE_ABORTED has purely cosmetics effects in Kodi. Given, that this state can only be set by the addon via a hack (parsing a human readable error description), I agree to remove this hacky code here.

Please refer to my code comments. Some smaller changes are still nessecary get this pr in good shape.

src/Tvheadend.cpp
```diff
@@ -1419,18 +1415,12 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   /* Error */
   if ((str = htsmsg_get_str(msg, "error")) != NULL)
   {
-    if (!strcmp(str, "300"))
-    {
-      UPDATE(rec.state, PVR_TIMER_STATE_ABORTED);
-    }
-    else if (strstr(str, "missing") != NULL)
+    if (rec.IsTimer())
```
ksooo: Sorry, but this will not work. "File missing" error string is set by tvh for a finished recording (https://github.com/tvheadend/tvheadend/blob/master/src/htsp_server.c#L720). 

Setting PVR_TIMER_STATE_ERROR for the recording with this particular error (having state PVR_TIMER_STATE_COMPLETED in this particular situation) is important as it makes Kodi put this recording in the Timer list in the GUI, which is fine.

With your change, it will keep the bogus recording in PVR_TIMER_STATE_COMPLETED which makes it appear in the list of finished recordings, which is wrong per definition.

src/Tvheadend.cpp
```diff
@@ -1385,13 +1385,9 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
   }
-  else if (strstr(state, "missed") != NULL)
-  {
-    UPDATE(rec.state, PVR_TIMER_STATE_ERROR);
-  }
-  else if (strstr(state, "invalid")   != NULL)
+  else
   {
-    UPDATE(rec.state, PVR_TIMER_STATE_ERROR);
+    UPDATE(rec.error, state);
```
ksooo: Assigning the state provided by tvh ("completed", "missed", ...) to the string error makes no sense for me.

Please keep this lines as they are. NO change here, please.

src/Tvheadend.cpp
```diff
@@ -1419,18 +1415,12 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   /* Error */
   if ((str = htsmsg_get_str(msg, "error")) != NULL)
   {
-    if (!strcmp(str, "300"))
-    {
-      UPDATE(rec.state, PVR_TIMER_STATE_ABORTED);
-    }
```
ksooo: This looks good to me.

src/Tvheadend.cpp
ksooo: Instead of removing the lines dealing with "missed" and "invalid" you should add

<pre>
else
{
  UPDATE(rec.state, PVR_TIMER_STATE_ERROR);
}  
</pre>

as a last resort for unhandled/unkown states (maybe introduced by new tvh versions).

Even more, this "else" is needed as you changed init value of SRecording::state to PVR_TIMER_STATE_NEW which would lead all recordings with unknown/unhandled as NEW. Unfortunately there is no PVR_TIMER_STATE_UNKNOWN. Thus we need to translate those unknown states to PVR_TIMER_STATE_ERROR.

Glenn-1990: No, we can't use "PVR_TIMER_STATE_ERROR" for invalid or missed recordings, if we do, these recordings will show up between the timers with "IsTimer()" and those are no timers. 

Not setting a state for missing and invalid recordings will leave them as "PVR_TIMER_STATE_NEW"
Everything with "PVR_TIMER_STATE_NEW" won't be transferred to kodi as it's not in IsTimer() or IsRecording(), this is how it should be as we don't want them in kodi. 
I needed to change this to keep the old behavior after I added "PVR_TIMER_STATE_ERROR" to IsTimer().

Glenn-1990: I agree that it's not needed to add invalid and missed to the error field

ksooo: I'm lost. I don't get what exactly do you want t achieve, sorry.

Let's forget for a moment what we have changed so far.

Could you please describe as exact as possible what is buggy with current implementation in master and what exactly do you want change.

ksooo:
Please, not everything that "works" is correct. Let's do one step back and rethink the concept.

Not setting "missed" and "invalid" items to PVR_TIMER_STATE_ERROR just because they appear in the wrong list in the GUI sounds just plan wrong. This cannot be the right solution, sorry.

And again, I only can think about a solution if I know what the original problem is.

Again: Could you please describe as exact as possible what is buggy with current implementation in master and what exactly do you want change.

Glenn-1990:
@ksooo I made a small update

I'll try to explain it in a few clear points:
1) with new tvheadend, error field is set if there is no free tuner, no signal,... (state from tvh = recording)
2) kodi shows recording active in this case which is wrong
3) so we should PVR_TIMER_STATE_ERROR if our error field is not empty
4) but IsTimer() and IsRecording are not containing "PVR_TIMER_STATE_ERROR" so it will not show up in kodi, but we want to show it in the timer list as error.
5) Add "PVR_TIMER_STATE_ERROR" to istimer() so kodi shows our timers with conflict, no signal, ...
6) We have to change the init value to New as we only may use "PVR_TIMER_STATE_ERROR" for timers now and not for recordings!.
7) invalid and missed recordings where tagged as PVR_TIMER_STATE_ERROR before, but these are not timers and we don't want them in kodi so let the have the init state 
8) invalid and missed recordings where tagged as PVR_TIMER_STATE_ERROR before, but these are not timers and we don't want them in kodi so let the have the init state PVR_TIMER_STATE_NEW, all timers with this state won't be transferred to kodi. before this PR all timers with PVR_TIMER_STATE_ERROR won't be transferred to kodi, so the behavior stays the same because of the change in IsTimer().
9) we only set PVR_TIMER_STATE_ERROR for timers otherwise completed recordings will show up in the timer list.

ksooo:
Okay. Trying to forget about all the code changes done in this PR so far and all the subsequent errors caused by that: 

The only thing that you want to change is that Kodi shall not show recordings with an error (no free tuner, no signal,...) as "Active" in the Timers window, but as "Error" (or similar), right?

Technically, this situation is exposed by tvheadend using state "recording" and various "error" strings. As those strings cannot be parsed safely (they might change at any time), using this approach imo is no good solution for your problem.

Now, after going back and forth trying to solve the problem without string parsing, things go worse, code is somehow "strange" (missed and invalid items cannot have state error any more to avoid that they show up in the wrong window and such...).

My suggestion is: Extend tvhadend to expose errors for timers/recordings in a clean way. I know you already contributed to tvheadend... It would be great if you could fix the root cause of the trouble where it should actually be addressed.

For what it is worth (just a changed string in the UI) I don't want to integrate a hack in pvr.hts, sorry.

ksooo:
:-1: from my side.

Glenn-1990:
The only thing that you want to change is that Kodi shall not show recordings with an error (no free tuner, no signal,...) as "Active" in the Timers window, but as "Error" (or similar), right?

Yes that's right, but I'm not parsing the strings, I'm just checking if there is an error or not?

I also have another approach:
https://github.com/Glenn-1990/pvr.hts/commit/4d85d059e2689a740d217f7038f0fb61cd1bea61

But, I'll do a pr to tvh, but then we still have the problem with the states that I tried to fix now..

Jalle19:
Perhaps Kodi needs to be extended to have a state for recordings too, not just timers?

PR #17, opened 2015-03-18
Added support for radio channel groups.

ksooo:
Added support for radio channel groups.

Please note that this feature relies on a Kodi fix. https://github.com/xbmc/xbmc/pull/6755

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -113,22 +113,6 @@ int CTvheadend::GetTagCount ( void )
   return m_tags.size();
 }
 
-bool CTvheadend::TagContainsChannelType ( const STag &tag, bool bRadio ) const
-{
-  vector<uint32_t>::const_iterator it;
-  SChannels::const_iterator cit;
-
-  for (it = tag.channels.begin(); it != tag.channels.end(); ++it)
-  {
-    if ((cit = m_channels.find(*it)) != m_channels.end())
-    {
-      if (bRadio == cit->second.radio)
-        return true;
-    }
-  }
-  return false;
-}
-
 PVR_ERROR CTvheadend::GetTags ( ADDON_HANDLE handle, bool bRadio )
 {
   if (!m_asyncState.WaitForState(ASYNC_DVR))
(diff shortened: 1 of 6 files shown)


ksooo:
BTW: why don't we have a "feature" label?

xhaggi:
because these are the default ;)

ksooo:
hehe.

xhaggi:
i've change the defaults (replace 'enhancement' with 'Improvement') because IMO this is an improvement

src/Tvheadend.cpp
```diff
@@ -113,7 +113,23 @@ int CTvheadend::GetTagCount ( void )
   return m_tags.size();
 }
 
-PVR_ERROR CTvheadend::GetTags ( ADDON_HANDLE handle )
+bool CTvheadend::IsTagKindMatch( const STag &tag, bool bRadio ) const
```
xhaggi: `IsTagContainChannelType(const STag &tag, bool bRadio)`?

xhaggi:
apart from my comment looks good ;)

ksooo:
@xhaggi IsTagKindMatch => IsTagContainsChannelType now. Much better now, imo.

xhaggi:
sounds good now ;)

Jalle19:
I would drop the "Is" from the method name.

ksooo:
@Jalle19 done. Dropped the "Is". Anything else?

Jalle19:
Looks good. To me it would make more sense to change some of our structs into classes and add some methods to them, but that's fof another time.

ksooo:
I agree. Let's do this over time whenever it makes sense. I thought of this yesterday while adding the new method to Tvheadend class. But adding this new method to STag makes not much sense to me, as implementation needs access to a Tvheadend member. But on second thought... Will rethink.

ksooo:
@Jalle19 couldn't resist. STag is a class now. :-) Please review.

ksooo:
class STag? "S" stands for struct, right? wtf! Will change this to namespace htsp { class Tag ... } later today.

Jalle19:
Yeah the S is for structs only.

ksooo:
Corrected the naming for the new class now.
@Jalle19 mind taking a look?

ksooo:
If nobody objects I will merge this later tonight. Version bump will follow later.

xhaggi:
+1

PR #21, opened 2015-03-31
Fix addon requiring restart no matter what settings are changed

Jalle19:
Two issues at play here:

* we use `type="number"` for the port settings, but without `option="int"`, the current value is sent to `ADDON_SetSetting` as a `float*` instead of `int*`, causing the comparison to fail inevitably. Changing the type to `integer` doesn't work either (even though it should), with that type the setting is not displayed in the dialog at all. See https://github.com/xbmc/xbmc/blob/master/xbmc/addons/AddonDll.h#L519

* the macro that compares if a string setting has changed was wrong, it did the exact opposite of what it was supposed to do.

@ksooo ping, and would also be great if you could runtime test this (I haven't tested at all but the changes worked fine for another addon I'm working on).

### DIFF
--- a/pvr.hts/resources/settings.xml
+++ b/pvr.hts/resources/settings.xml
@@ -2,8 +2,8 @@
 <settings>
   <category label="30000">
 	<setting id="host" type="text" label="30001" default="127.0.0.1" />
-	<setting id="http_port" type="number" label="30002" default="9981" />
-	<setting id="htsp_port" type="number" label="30003" default="9982" />
+	<setting id="http_port" type="number" option="int" label="30002" default="9981" />
+	<setting id="htsp_port" type="number" option="int" label="30003" default="9982" />
 	<setting id="user" type="text" label="30004" default="" />
 	<setting id="pass" type="text" label="30005" option="hidden" default="" />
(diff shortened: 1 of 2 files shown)


ksooo:
... and what is fixed by this? Title implies that now changes to http and htsp port do no longer require a restart of the addon, right?

Jalle19:
Yes that's what it fixes, I guess I didn't make it clear.

ksooo:
@Jalle19 will take a look when back from Easter holidays...

Jalle19:
@ksooo did you get a chance to take a look at this?

ksooo:
@Jalle19 Not yet, sorry. I have always a bad feeling about this "it fixes stuff, but don't know why..."

Jalle19:
@ksooo it's the string comparison I'm not sure about, the other commit is very much valid. I suspect the strings may appear non-equal due to missing null terminator or something like that.

Jalle19:
@ksooo @opdenkamp I found the actual error

ksooo:
Cool. What is it? Does it influence this PR?

Jalle19:
@ksooo see the last commit

ksooo:
@jalle19 hehe. Good catch. Will runtime test tonight (hopefully).

ksooo:
@Jalle19 runtime tested. Seems to work nicely. Thus, :+1: from my side.

PR #22, opened 2015-04-13
add support for reapeating timers, aka time schedules

Glenn-1990:
As series recording support probably won't be merged for Kodi 15.0, we could at least support repeating manual timers (aka time schedules in tvheadend).

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -575,6 +575,14 @@ PVR_ERROR CTvheadend::AddTimer ( const PVR_TIMER &timer )
   uint32_t u32;
   dvr_prio_t prio;
 
+  if (timer.bIsRepeating && timer.iWeekdays)
+  {
+    if (m_conn.GetProtocol() >= 18)
+      return AddTimeRecording(timer);
+    else
+      return PVR_ERROR_NOT_IMPLEMENTED;
+  }
+
   /* Build message */
   htsmsg_t *m = htsmsg_create_map();
   if (timer.iEpgUid > 0)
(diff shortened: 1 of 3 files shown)


ksooo:
Isengard beta start is planned for around April 20th and after this we have feature freeze. I'm very busy with the series reocrding stuff and don't have enough free time for other things (even to review PRs like this), sorry. I'm afraid, it's already to late for this feature PR.

And: This (naturally) higly interfears with what I'm currently doing for after Isengard series recording support -makes my life harder wrt merge conflicts etc. => https://github.com/ksooo/pvr.hts/commits/series-recording-support

dreamcat4:
@Glenn-1990, @ksooo, … it is great work both of you are doing in regards to the series recording support. All of us tvheadend / kodi users - we really appreciate it.

Glenn-1990:
@ksooo 
If I can help you with testing, please let me know

If a user selects multiple days in the timer dialog, an error gets returned:
ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - Command addDvrEntry failed: Could not add ERROR: PVR - AddTimer - addon 'HTS Tvheadend:127.0.0.1:9982 [connected]' returned an error: server error

So at east we need to handle this properly?

ksooo:
@Glenn-1990 Regarding my work on after Isengard series recording support: Thanks for your kind offer to help testing. I will let you know when things are ready.

Regarding this PR. The bug you described for me is "just" one of a whole bunch of problems we currently have in this area. If you have ideas for simple fixes (not adding missing features) either in Kodi or the addon feel free to file PRs for this. Such fixes could be Kodi error messages for instance, maybe like you suggested, PVR_ERROR_NOT_IMPLEMENTED in the addon where this is appropriate.

Glenn-1990:
Modified a bit to have minimal impact to the code base, 
new code only gets triggered when user selects multiple days

Glenn-1990:
@ksooo, Are code changes allowed after Lsengard splits from master for this addon?
Then things will not have impact on your work

ksooo:
@Glenn-1990 the problem is not code changes as such, but the semantics of the changes. To my best knowledge, after start of Isengard beta period, we have a strict feature freeze. And I consider this PR a feature.

src/Tvheadend.cpp
```diff
@@ -671,6 +679,64 @@ PVR_ERROR CTvheadend::UpdateTimer ( const PVR_TIMER &timer )
...
+  htsmsg_add_str(m, "title",        timer.strTitle);
+  htsmsg_add_str(m, "name",         timer.strTitle);
+  htsmsg_add_u32(m, "channelId",    timer.iClientChannelUid);
+  htsmsg_add_str(m, "description",  timer.strSummary);
+  htsmsg_add_str(m, "comment",      "Created by Kodi Media Center");
+
+  /* Convert start and stop time to time after midnight */
+  struct tm *tmi;
+  tmi=localtime(&timer.startTime);
+  htsmsg_add_s64(m, "start",        (tmi->tm_hour*60 + tmi->tm_min));
```
ksooo: Have you actually checked that this code works? ;-) I doubt this, as I've found out that tvheadend expects "start" and "stop" to be u32 (although the spec says s64)...

Glenn-1990: Yes, it makes sense to use u32 then.
Conversion from s64 to u32 works, tested with tvheadend master.

Jalle19:
Addons are not concerned by the feature freeze, that's the main reason why they are addons.

src/Tvheadend.h
```diff
@@ -376,6 +376,8 @@ class CTvheadend
   PVR_ERROR DeleteTimer       ( const PVR_TIMER &tmr, bool force );
   PVR_ERROR UpdateTimer       ( const PVR_TIMER &tmr );
 
+  PVR_ERROR AddTimerecording  ( const PVR_TIMER &tmr );
```
Jalle19: Should probably have an upper-case "r", i.e. `AddTimeRecording`.

src/Tvheadend.cpp
```diff
@@ -671,6 +679,64 @@ PVR_ERROR CTvheadend::UpdateTimer ( const PVR_TIMER &timer )
...
+    CLockObject lock(m_conn.Mutex());
+    m = m_conn.SendAndWait("addTimerecEntry", m);
+  }
+
+  if (m == NULL)
+    return PVR_ERROR_SERVER_ERROR;
+
+  /* Check for error */
+  if (htsmsg_get_u32(m, "success", &u32))
+  {
```
Jalle19: Braces should be dropped here

src/Tvheadend.cpp
```diff
@@ -671,6 +679,64 @@ PVR_ERROR CTvheadend::UpdateTimer ( const PVR_TIMER &timer )
...
+  htsmsg_add_u32(m, "daysOfWeek",   timer.iWeekdays);
+  htsmsg_add_str(m, "title",        timer.strTitle);
+  htsmsg_add_str(m, "name",         timer.strTitle);
+  htsmsg_add_u32(m, "channelId",    timer.iClientChannelUid);
+  htsmsg_add_str(m, "description",  timer.strSummary);
+  htsmsg_add_str(m, "comment",      "Created by Kodi Media Center");
+
+  /* Convert start and stop time to time after midnight */
+  struct tm *tmi;
+  tmi=localtime(&timer.startTime);
```
Jalle19: Missing spaces around the equal sign.

src/Tvheadend.cpp
```diff
@@ -575,6 +575,14 @@ PVR_ERROR CTvheadend::AddTimer ( const PVR_TIMER &timer )
   uint32_t u32;
   dvr_prio_t prio;
 
+  if (timer.bIsRepeating && timer.iWeekdays)
+  {
+    if (m_conn.GetProtocol() > 17)
```
Jalle19: I don't know about you but in cases such as this I think `>= 18` is clearer since it more clearly states that version 18 is the minimum required.

Jalle19:
@Glenn-1990 could you fix the (very) minor comments I made so we could get this in, unless @ksooo has any concerns? Since we need to merge @opdenkamp's latest PR anyway we might as well merge the things we have and make a new release.

ksooo:
No concerns from my side. 

But please wait some more days with the release. I have another pr in the queue I would like to see in that release (subtitle support for epg events).

Jalle19:
@ksooo no problem, thanks for the heads up.

Glenn-1990:
@Jalle19 updated

PR #23, opened 2015-04-30
Implement predictive tuning using multiple tuners

martinwilli:
If multiple tuners are available in TVHeadend, and the addon is configured
to use them, this patch tries to anticipate the users zapping behavior, and
have the channel tuned and buffered before the channel switch happens. This
allows almost instant channel switches if the prediction was correct.

When using the default of a single tuner, the tuning behavior does not change.
If two tuners are configured, sequential up/down-zapping pre-tunes the next
channel in order; for direct channel selection the previously tuned channel
is kept to allow fast channel swapping. If more tuners are configured,
previously used channels are kept open to allow up/down tuning in any order.

The plugin uses a configurable timeout to close a predicted or previously used
channel if it does not get used, defaulting to 10s. It can be disabled by
configuring 0.

--

I have plenty of free tuners in my TVHeadend server, so I'd like to make better use of them. Im no C++ expert, nor have I any experience with the Kodi codebase, so let me know if you think something can be done simpler or cleaner. Also let me know if you want me to split up the patch to smaller changes.

The tuning prediction is currently rather simple and certainly can be improved, but it works actually very fine here for a first version.

One of the tricky points is to Trim() the input buffer correctly on anticipated/unused channels, so we don't waste memory and have just enough data so DVDPlayer doesn't need buffering. I currently use a fixed count of packets; maybe we could improve this by buffering more or less data depending on the bitrate of the channel.

We also have some room for improvements regarding subscription priority; not sure how difficult it is, but we could use a lower priority for predicted/unused streams to not interrupt other frontends using TVHeadend.

### DIFF
--- a/pvr.hts/resources/settings.xml
+++ b/pvr.hts/resources/settings.xml
@@ -14,6 +14,11 @@
 	<setting id="epg_async" type="bool" label="30101" default="false"/>
   </category>
 
+  <category label="30400">
+	<setting id="total_tuners" type="number" option="int" label="30401" default="1" />
+	<setting id="pretuner_closedelay" type="number" option="int" label="30402" default="10" />
+  </category>
+
   <category label="30200">
 	<setting id="trace_debug" type="bool" label="30201" default="false"/>
   </category>
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -44,6 +44,7 @@ CHTSPDemuxer::CHTSPDemuxer ( CHTSPConnection &conn )
   : m_conn(conn), m_pktBuffer((size_t)-1),
     m_seekTime(INVALID_SEEKTIME)
 {
+  m_lastUse = 0;
 }
 
 CHTSPDemuxer::~CHTSPDemuxer ( void )
@@ -84,7 +85,7 @@ void CHTSPDemuxer::Abort0 ( void )
 }
 
 
-bool CHTSPDemuxer::Open ( const PVR_CHANNEL &chn )
+bool CHTSPDemuxer::Open ( uint32_t channelId, enum eSubscriptionWeight weight )
 {
   CLockObject lock(m_conn.Mutex());
   tvhdebug("demux open");
@@ -94,7 +95,8 @@ bool CHTSPDemuxer::Open ( const PVR_CHANNEL &chn )
   
   /* Create new subscription */
   m_subscription = SSubscription();
-  m_subscription.channelId = chn.iUniqueId;
+  m_subscription.channelId = channelId;
+  m_subscription.weight = weight;
 
   /* Open */
   SendSubscribe();
@@ -102,6 +104,8 @@ bool CHTSPDemuxer::Open ( const PVR_CHANNEL &chn )
   /* Send unsubscribe if subscribing failed */
   if (!m_subscription.active)
     SendUnsubscribe();
+  else
+    m_lastUse = time(NULL);
   
   return m_subscription.active;
 }
@@ -116,6 +120,7 @@ void CHTSPDemuxer::Close ( void )
 DemuxPacket *CHTSPDemuxer::Read ( void )
 {
   DemuxPacket *pkt = NULL;
+  m_lastUse = time(NULL);
   if (m_pktBuffer.Pop(pkt, 1000)) {
     tvhtrace("demux read idx :%d pts %lf len %lld",
              pkt->iStreamId, pkt->pts, (long long)pkt->iSize);
@@ -131,6 +136,18 @@ void CHTSPDemuxer::Flush ( void )
   DemuxPacket *pkt;
   tvhtrace("demux flush");
   while (m_pktBuffer.Pop(pkt))
+    PVR->FreeDemuxPacket(pkt);
+}
+
+void CHTSPDemuxer::Trim ( void )
+{
+  DemuxPacket *pkt;
+
+  tvhtrace("demux trim");
+  /* reduce used buffer space to what is needed for DVDPlayer to resume
+   * playback without buffering. This depends on the bitrate, so we don't set
+   * this too small. */
+  while (m_pktBuffer.Size() > 512 && m_pktBuffer.Pop(pkt))
     PVR->FreeDemuxPacket(pkt);
 }
 
@@ -191,6 +208,14 @@ void CHTSPDemuxer::Speed ( int speed )
   SendSpeed();
 }
 
+void CHTSPDemuxer::Weight ( enum eSubscriptionWeight weight )
+{
+  if (!m_subscription.active || m_subscription.weight == weight)
+    return;
+  m_subscription.weight = weight;
+  SendWeight();
+}
+
 PVR_ERROR CHTSPDemuxer::CurrentStreams ( PVR_STREAM_PROPERTIES *streams )
 {
   CLockObject lock(m_mutex);
@@ -239,6 +264,7 @@ void CHTSPDemuxer::SendSubscribe ( bool force )
   m = htsmsg_create_map();
   htsmsg_add_s32(m, "channelId",       m_subscription.channelId);
   htsmsg_add_u32(m, "subscriptionId",  m_subscription.subscriptionId);
+  htsmsg_add_u32(m, "weight",          m_subscription.weight);
   htsmsg_add_u32(m, "timeshiftPeriod", (uint32_t)~0);
   htsmsg_add_u32(m, "normts",          1);
   htsmsg_add_u32(m, "queueDepth",      2000000);
@@ -298,26 +324,33 @@ void CHTSPDemuxer::SendSpeed ( bool force )
     htsmsg_destroy(m);
 }
 
+void CHTSPDemuxer::SendWeight ( void )
+{
+  CLockObject lock(m_conn.Mutex());
+  htsmsg_t *m;
+
+  /* Build message */
+  m = htsmsg_create_map();
+  htsmsg_add_u32(m, "subscriptionId", m_subscription.subscriptionId);
+  htsmsg_add_s32(m, "weight",         m_subscription.weight);
+  tvhdebug("demux send weight %u", m_subscription.weight);
+
+  /* Send and Wait */
+  m = m_conn.SendAndWait("subscriptionChangeWeight", m);
+  if (m)
+    htsmsg_destroy(m);
+}
+
 /* **************************************************************************
  * Parse incoming data
  * *************************************************************************/
 
 bool CHTSPDemuxer::ProcessMessage ( const char *method, htsmsg_t *m )
 {
-  uint32_t subId;
-
   CLockObject lock(m_mutex);
 
-  /* No subscriptionId - not for demuxer */
-  if (htsmsg_get_u32(m, "subscriptionId", &subId))
-    return false;
-
-  /* Not current subscription - ignore */
-  else if (subId != m_subscription.subscriptionId)
-    return true;
-
   /* Subscription messages */
-  else if (!strcmp("muxpkt", method))
+  if (!strcmp("muxpkt", method))
     ParseMuxPacket(m);
   else if (!strcmp("subscriptionStatus", method))
     ParseSubscriptionStatus(m);
@@ -588,6 +621,10 @@ void CHTSPDemuxer::ParseSubscriptionStatus ( htsmsg_t *m )
 {
   const char *status;
   status = htsmsg_get_str(m, "status");
+
+  // not for preTuning subscriptions
+  if (m_subscription.weight != SUBSCRIPTION_WEIGHT_DEFAULT)
+    return;
 
   // this field is absent when everything is fine
   if (status != NULL)
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -42,14 +42,23 @@ using namespace ADDON;
 using namespace PLATFORM;
 
 CTvheadend::CTvheadend(tvheadend::Settings settings)
-  : m_settings(settings), m_dmx(m_conn), m_vfs(m_conn), 
+  : m_settings(settings), m_vfs(m_conn),
     m_queue((size_t)-1), m_asyncState(settings.iResponseTimeout),
     m_timeRecordings(m_conn), m_autoRecordings(m_conn)
 {
+  for (int i = 0; i < 1 || i < m_settings.iTotalTuners; i++)
+  {
+    m_dmx.push_back(new CHTSPDemuxer(m_conn));
+  }
+  m_dmx_active = m_dmx[0];
 }
 
 CTvheadend::~CTvheadend()
 {
+  for (auto *dmx : m_dmx)
+  {
+    delete dmx;
+  }
   m_conn.StopThread(-1);
   m_conn.Disconnect();
   StopThread();
@@ -1204,7 +1213,10 @@ bool CTvheadend::Connected ( void )
   SEvents::iterator eit;
 
   /* Rebuild state */
-  m_dmx.Connected();
+  for (auto *dmx : m_dmx)
+  {
+    dmx->Connected();
+  }
   m_vfs.Connected();
 
   /* Flag all async fields in case they've been deleted */
@@ -1246,9 +1258,18 @@ bool CTvheadend::Connected ( void )
 
 bool CTvheadend::ProcessMessage ( const char *method, htsmsg_t *msg )
 {
-  /* Demuxer */
-  if (m_dmx.ProcessMessage(method, msg))
+  uint32_t subId;
+
+  if (!htsmsg_get_u32(msg, "subscriptionId", &subId))
+  {
+    /* subscriptionId found - for a Demuxer */
+    for (auto *dmx : m_dmx)
+    {
+      if (dmx->GetSubscriptionId() == subId)
+        return dmx->ProcessMessage(method, msg);
+    }
     return true;
+  }
 
   /* Store */
   m_queue.Push(CHTSPMessage(method, msg));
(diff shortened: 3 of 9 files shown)


ksooo:
Many thanks. Will take a closer look asap (next days probably).

Would you mind rebasing to latest pvr.tvh master?

ksooo:
Please remove the German strings.po from the commit. Translation to any other language than en_g will be done later using Transifex.

ksooo:
Are you familiar with the htsp spec? https://tvheadend.org/projects/tvheadend/wiki/Htsp Might be helpful when thinking about subscription priorities... 
"subscripe" method has a field "weight". There is also a method "subscriptionChangeWeight"... 
Perhaps you can make something out of this?

ksooo:
What do you think about the possibility to define "some" favourite channels in the addon settings and to try to prefetch as many as possible/configured. Just thinking out loud...

Glenn-1990:
I think this should be something for the tvheadend side, not for the addon, but that's my opinion..

martinwilli:
Thanks for the pointer with the subscription weight. I'll have a look into that, actually shouldn't be too hard to implement.

Having some favourite channels would be really nice, but I think that is something for a next iteration.

> I think this should be something for the tvheadend side, not for the addon, but that's my opinion..

I actually thought about that, but I don't agree. Having the data streamed to Kodi makes it available  immediately on channel switch. This avoids buffering in DVDPlayer, and allows immediate playback. When using fast tuners, the time for buffering is non-negligible compared to the time TVHeadend needs to tune on the channel and deliver that data.

dreamcat4:
What would be even better, is if this feature had some other way to be re-triggered. In a predictive manner. For example: when I pick up my air mouse remote, or mobile phone, well that aught to triggers it's accelerometer. And if Kodi is in the pvr menus at the time. And Kodi is being informed about that event. We can be reasonably confident that the user may be intending to change the pvr state somehow.

So that also would be the ideal time to start buffering alternative channels...

What you may need for that is perhaps specific kind of API interaction between the main Kodi and it's pvr addons. e.g. a some generic kodi-wide notification that can be subscribed to by our pvr.hts addon.

Of course. That ^ above example may not be possible today. However a generic 'recent activity' notification would not be limited purely just to accellerometer remote input. It could also detect regular keypresses. Or potentially come from other kinds of future support too. e.g. Microsoft Kinect etc. Or even broader than that.

For current users, just regular remotes (no special fancy input support). Then the trigger can be just when *any remote key has been pressed while within the pvr addon's menus*. Including back, menu, select, 'info, epg buttons etc.

Glenn-1990:
It would be better to have predictive muxes instead of channels.
i.e. if you have 2 free tuners, each tuned to a mux of 10 channels, you'll have 20 channels that will start almost instant when switching to them.

And with with timeshift? if all streams will use timeshifting, your system gets stressed a lot with no reason.

Jalle19:
For switching to happen quickly it has to be streamed to Kodi continuosly. Many devices are limited to 100 Mbit/s so streaming more than a few channels at a time becomes unpractical.

I like this idea and I think we should try to start out simple, i.e. a setting where you can choose to "buffer" previous/next/both previous and next channels. Quick sequential zapping is the main use case and that's what we should focus on initially.

ksooo:
I fully agree to what @jalle19 said. 

It's a cool feature with lots of potential for future enhancements, imo. Let's start simple, guys, and let's see where the journey will lead us to. ;-) 

The only drawback I see is risk to take your network and tvh itself to the limits, but this is just user fault (wrong configuration) - it' a power user feature, right? And it can be fully failed, that's important. And default should be disabled (which it is with the current implementation).

 I also agree that this is not a back-end topic. Implementing it at Kodi-side is the right approach, I think.

Btw, I'm on the loser side here as I only have two tuners...

ksooo:
a/failed/disabled

ksooo:
For me, the one thing that is missing for now is to do all predictive subscriptions with very low priority to make sure no tuners are blocked for other clients that want to do a real subscription and there must be some resources available for tbheadendv itself to do background mux and epg scanning and such...

ksooo:
Typing correct sentences on the phone is pita, isn't it?

martinwilli:
I've updated the patch to implement subscription weight management. For the subscription currently watched it now explicitly uses a weight of 150, which in the latest TVHeadend is the implicit default. For predicted channels it uses a weight of 50, and for channels zapping away from a weight of 40.

Timeshifting has not been a problem in my tests, even with four and more active subscriptions. I actually think it is a great feature, as you can switch between channels, and do time-shifting on channels that you haven't watched (say you missed switching back on time to your movie after the commercials). Also, newer TVHeadend versions support timeshifting to memory buffers, which avoids heavy load on your TVHeadend server disk.

Further, I agree that a first version of this functionality should be simple, more or less the feature-set we have. While there are many ways we could improve prediction, the current patch covers the main use cases and is rather simple.

src/HTSPDemuxer.cpp
```diff
@@ -191,6 +208,15 @@ void CHTSPDemuxer::Speed ( int speed )
   SendSpeed();
 }
 
+void CHTSPDemuxer::Weight ( uint32_t weight )
+{
+  CLockObject lock(m_conn.Mutex());
```
Jalle19: I'd move this lock to `CHTSPDemuxer::SendWeight` to keep the lock scope as small as possible (that's how we do it in most other places).

src/Tvheadend.h
```diff
@@ -60,6 +60,9 @@ extern "C" {
 #define FAST_RECONNECT_INTERVAL   (500) // ms
 #define UNNUMBERED_CHANNEL      (10000)
 #define INVALID_SEEKTIME           (-1)
+#define DEFAULT_WEIGHT            (150)
```
Jalle19: I'd use something like `enum SubscriptionWeight` in `CHTSPDemuxer` instead of defines.

pvr.hts/resources/settings.xml
```diff
@@ -14,6 +14,11 @@
 	<setting id="epg_async" type="bool" label="30101" default="false"/>
   </category>
 
+  <category label="30300">
+	<setting id="pretuners" type="number" label="30301" default="1" />
+	<setting id="pretuner_closedelay" type="number" label="30302" default="10" />
```
Jalle19: Both of these need `option="int"` as well, see https://github.com/Jalle19/pvr.hts/commit/4aad8b6aa15d9c30d5eced22f5fa43c19b16f7ee

src/Tvheadend.cpp
```diff
@@ -1640,3 +1661,168 @@ uint32_t CTvheadend::GetNextUnnumberedChannelNumber()
...
 }
+
+void CTvheadend::TuneOnOldest( uint32_t channelId )
+{
+  int oldest = -1;
+
+  for (int i = 0; i < m_dmx.size(); i++)
+  {
+    if (m_dmx[i]->GetChannelId() == channelId)
+      return m_dmx[i]->Weight(PRETUNING_WEIGHT);
```
Jalle19: A void method shouldn't return anything

src/Tvheadend.cpp
```diff
@@ -1640,3 +1661,168 @@ uint32_t CTvheadend::GetNextUnnumberedChannelNumber()
...
+{
+  for (int i = 0; i < m_dmx.size(); i++)
+  {
+    m_dmx[i]->Abort();
+  }
+}
+
+bool CTvheadend::DemuxSeek ( int time, bool backward, double *startpts )
+{
+  m_dmx[m_dmx_active]->Seek(time, backward, startpts);
```
Jalle19: Missing return values here and in the other non-void methods below this one.

src/HTSPDemuxer.cpp
martinwilli: Ok, done.

src/Tvheadend.h
martinwilli: I've defined such an enum in HTSPTypes.h, and used this type where appropriate.

pvr.hts/resources/settings.xml
martinwilli: I've added those.

src/Tvheadend.cpp
martinwilli: Actually, this is more a syntactic sugar thing, as Weight() returns void as well. But I've splitted that up to follow with a _return;_ on that statement.

src/Tvheadend.cpp
martinwilli: Very true, fixed.

martinwilli:
Thanks for reviewing the patch. I've pushed a new version, comments above.

Jalle19:
Thanks, I'll do some runtime testing later.

stan86:
I'm testing it and seems to work well!

src/client.h
```diff
@@ -30,11 +30,13 @@ extern ADDON::CHelper_libXBMC_addon*  XBMC;
...
-#define DEFAULT_HTTP_PORT        9981
-#define DEFAULT_HTSP_PORT        9982
-#define DEFAULT_CONNECT_TIMEOUT  10
-#define DEFAULT_RESPONSE_TIMEOUT 5
+#define DEFAULT_HOST                "127.0.0.1"
+#define DEFAULT_HTTP_PORT           9981
+#define DEFAULT_HTSP_PORT           9982
+#define DEFAULT_CONNECT_TIMEOUT     10
+#define DEFAULT_RESPONSE_TIMEOUT    5
+#define DEFAULT_PRETUNERS           1
```
Jalle19: Perhaps we should change this setting to TOTAL_TUNERS instead since that's what it actually represents, right? I have a feeling someone is going to be confused and set this value to zero because they only have one tuner.

src/Tvheadend.h
```diff
@@ -386,7 +407,10 @@ class CTvheadend
   const tvheadend::Settings   m_settings;
   
   CHTSPConnection             m_conn;
-  CHTSPDemuxer                m_dmx;
+
+  std::vector<CHTSPDemuxer*>  m_dmx;
+  int                         m_dmx_active;
```
Jalle19: I'd store a pointer to the actual demuxer instead of the vector index, that way all the loops over `m_dmx` can be changed to `for (auto *dmx : m_dmx)` which is a bit more concise.

pvr.hts/resources/language/English/strings.po
```diff
@@ -69,3 +69,15 @@ msgstr ""
...
+msgctxt "#30300"
+msgid "Predictive tuning"
+msgstr ""
+
+msgctxt "#30301"
+msgid "Number of tuners to use"
+msgstr ""
+
+msgctxt "#30302"
+msgid "Unused channel close delay"
```
Jalle19: Should probably be "Unused subscription close delay"

Jalle19:
@martinwilli if you could take care of my comments and rebase that would be great!

martinwilli:
Thanks for the review. I've rebased the original patch, and addressed the mentioned issues. I've pushed individual commits for review, but of course can squash these into the original commit for a merge.

muzena:
@martinwilli Can you make *.deb packages with this patch?
I'm tried commit patch but I get HUNK errors.

ksooo:
@martinwill could you please rebase against latest master?

ksooo:
@martinwilli and please squash down to a single commit.

martinwilli:
>  could you please rebase against latest master?
>  and please squash down to a single commit.

Done. Looking forward to finally get this merged.

dreamcat4:
@ksoo I think this has broken the `2.2.4`. Plugin won't load channels now. Why is this experimental feature on by default? How to switch it off? (I can't seem to switch it off either)

```sh
12:17:22 T:140614251300608  NOTICE: Thread BackgroundLoader start, auto delete: false
12:17:22 T:140614942508800   ERROR: SQL: Abort due to constraint violation
                                            Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon, bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (301732504, 1, 0, 0, 0, 0, 'http://id:sn00zey0ul00se@192.168.2.3:9981/1_0_C_2922_90B_3B_11A0000_0_0_0.png', 'Freesat Info', 0, 1, 'client', 0, 15641, -1)
12:17:22 T:140614942508800   ERROR: CommitInsertQueries - failed to execute queries
12:17:22 T:140614942508800   ERROR: SQL: Abort due to constraint violation
                                            Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon, bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (301732504, 1, 0, 0, 0, 0, 'http://id:sn00zey0ul00se@192.168.2.3:9981/1_0_C_2922_90B_3B_11A0000_0_0_0.png', 'Freesat Info', 0, 1, 'client', 0, 15641, -1)
12:17:22 T:140614942508800   ERROR: CommitInsertQueries - failed to execute queries
12:17:22 T:140614942508800   ERROR: SQL: Abort due to constraint violation
                                            Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon, bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (301732504, 1, 0, 0, 0, 0, 'http://id:sn00zey0ul00se@192.168.2.3:9981/1_0_C_2922_90B_3B_11A0000_0_0_0.png', 'Freesat Info', 0, 1, 'client', 0, 15641, -1)
12:17:22 T:140614942508800   ERROR: CommitInsertQueries - failed to execute queries
12:17:22 T:140614942508800   ERROR: CPVRChannelGroups - Load - failed to load channel groups
12:17:22 T:140614942508800   ERROR: PVRManager - Process - failed to load PVR data, retrying
12:17:37 T:140614942508800  NOTICE: Thread BackgroundLoader start, auto delete: false
12:18:56 T:140614242907904  NOTICE: Thread JobWorker start, auto delete: true
12:18:56 T:140614259693312  NOTICE: Thread BackgroundLoader start, auto delete: false
12:18:58 T:140614242907904  NOTICE: PVRManager - starting up
12:18:58 T:140614259693312  NOTICE: Thread PVRManager start, auto delete: false
12:18:58 T:140614942508800  NOTICE: Thread PVRClient start, auto delete: false
12:18:58 T:140614586844928  NOTICE: Thread BackgroundLoader start, auto delete: false
12:18:58 T:140614259693312   ERROR: SQL: Abort due to constraint violation
                                            Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon, bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (301732504, 1, 0, 0, 0, 0, 'http://id:sn00zey0ul00se@192.168.2.3:9981/1_0_C_2922_90B_3B_11A0000_0_0_0.png', 'Freesat Info', 0, 1, 'client', 0, 15641, -1)
12:18:58 T:140614259693312   ERROR: CommitInsertQueries - failed to execute queries
12:18:58 T:140614259693312   ERROR: SQL: Abort due to constraint violation
                                            Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon, bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (301732504, 1, 0, 0, 0, 0, 'http://id:sn00zey0ul00se@192.168.2.3:9981/1_0_C_2922_90B_3B_11A0000_0_0_0.png', 'Freesat Info', 0, 1, 'client', 0, 15641, -1)
12:18:58 T:140614259693312   ERROR: CommitInsertQueries - failed to execute queries
12:18:58 T:140614259693312   ERROR: SQL: Abort due to constraint violation
                                            Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon, bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (301732504, 1, 0, 0, 0, 0, 'http://id:sn00zey0ul00se@192.168.2.3:9981/1_0_C_2922_90B_3B_11A0000_0_0_0.png', 'Freesat Info', 0, 1, 'client', 0, 15641, -1)
12:18:58 T:140614259693312   ERROR: CommitInsertQueries - failed to execute queries
12:18:58 T:140614259693312   ERROR: CPVRChannelGroups - Load - failed to load channel groups
12:18:58 T:140614259693312   ERROR: PVRManager - Process - failed to load PVR data, retrying
```

ksooo:
@dreamcat4 I believe we are in alpha phase, so you have to expect difficulties when dealing with bleading edge. However, the log fragment you posted contains no hint that the problem is in any way related to this feature.

 And, no the feature is not turned on by default. => https://github.com/kodi-pvr/pvr.hts/blob/master/pvr.hts/resources/settings.xml#L18 (total_tuners =1)

Jalle19:
I've seen that error before, it's not related to this PR but it's interesting nontheleess. I'm not sure if Kodi screws up (likely) or if tvheadend can somehow use the same ID for multiple channels.

dreamcat4:
Hmm interesting. Sorry i kindda assumed because this was the biggest changset since 2.2.3.

I am just trying to get a build with the series recording support. Tried master nightlies:

Linux x86 - 2.2.2 - api version mismatch
Windows - 2.2.3 - WORKS! :)
Linux - 2.2.4 - broken (this error)

Unfortunately due to poor timing the wsniperx builds of the ubuntu (.deb) verision of this addon seems to have skipped 2.2.3 entirely.

I agree the log message isn't particularly insightful. Enabled 'trace' option in pvr.hts settings but that didn't seem to change anything in the kodi.log (perhaps requires a restart?)

PR #25, opened 2015-05-04
Improve the file management in generated solutions

Jalle19:
There was some regression in the usability of the Visual Studio solution after we switched to cmake. This brings it up to par with how it was before.


ksooo:
I'm not a cmake expert, but this looks sane to me.

Jalle19:
Me neither, but I've used the same approach in https://github.com/kodi-pvr/pvr.vbox/blob/master/CMakeLists.txt and it works fine. The biggest improvement is the ability to bump versions etc. directly in the IDE instead of having to hunt down the file manually.

PR #30, opened 2015-05-16
[debian] fix packaging once more: correct install file permissions

wsnipex:
for some strange reason, packaging breaks if the install file has execute permissions

@hudokkow ping. Same is needed for some other addons...


Jalle19:
Thanks!

hudokkow:
Arghhh, I followed your original PR. :P

PR #33, opened 2015-05-25
Coverity fixes

ksooo:


Jalle19:
Thanks a lot, merge when you feel like it.

Jalle19:
We should probably merge this and make a new release before Isengard final is out.

ksooo:
Yep.

ksooo:
@negge this one should go into the next release. 

What about #23? Have you found time to runtime test it? This feature is imo cool, but as I only have 2 tuners actually not that valuable for me. Thus, my priorities are a bit different...

Jalle19:
No I haven't tested it yet, I'm still using Helix at home but once Isengard is released I'll probably update, should make it easier to test.

PR #34, opened 2015-05-30
CTvheadend::ParseEvent now handles "seasonNumber", "episodeNumber" an…

ksooo:
…d "partNumber" htsp fields. Seems, that this was forgotten as other places in the addon already use this information.

Just saw @ronie's latest cool changes wrt adding support for season, episode, ... to Confluence's PVR guide windows and dialogs and realized this bug in pvr.hts...


ksooo:
Boy! That was fast!

PR #35, opened 2015-06-03
fix includes

uNiversaI:


opdenkamp:
as pointed out in one of the other PRs, this breaks when 2 libs use the same header filename

wsnipex:
could you please outline that scenario?

opdenkamp:
what's to outline about that? 2 libs can use the same filename, compiler will pick whatever it finds first.

wsnipex:
are you saying we have kodi headers in various places? If yes something is seriously broken.

opdenkamp:
in general this is not a good idea, not kodi specific. specify the directory where a header is coming from and you won't run into issues. that some other libs do the same thing as this pr is doing does not mean that it's a good idea to do this.

opdenkamp:
quick example between jtag uploads ;)

in libCEC there's a header named `version.h` (pretty generic name) sitting in the root of the header dir. if you add this include dir directly in the cpp flags, then having a line like `#include "version.h"` in code will depend on things like the order of the cpp flags. which is not a good idea.

wsnipex:
again, we're only talking about kodi headers here.

opdenkamp:
it was an example

Jalle19:
Isn't this basically a revert of an earlier change?

uNiversaI:
@jalle if you mean #24 then no

Jalle19:
So is there any point in merging this? @opdenkamp ?

uNiversaI:
All the other binary addons -> audio encoders / decoders have been done now.

Jalle19:
Okay, but what is the point? To me it's clearer to include the directory name in the includes.

uNiversaI:
See https://github.com/xbmc/audioencoder.vorbis/pull/9 for discussion, The bikeshed already been painted/repainted there.

opdenkamp:
bikeshed my ass and already explained why this is a bad idea

uNiversaI:
> bikeshed my ass and already explained why this is a bad idea

@wsnipex one last comment, I have nothing further to contribute and only done PVR with these changes to help out devs with monkeywork.

@opdenkamp PR's are here tbh I care less about PVR as I dont even use it.

wsnipex:
@opdenkamp no you didn't explain anything. You just gave an invalid example. There will be no duplicate kodi headers.

And I agree with @uNiversaI, take it or leave it, idc

Jalle19:
I don't really care either, I just want to know if this is valid and whether it should be merged or not.

opdenkamp:
I explained why it's a bad idea in general, with an example that applies to libCEC.

The problem with this is not Kodi itself, but that you require _all_ headers that you include to have unique filenames, from everything you include and not just Kodi's own stuff.

I saw that the reason for doing this change is that (in theory) some Linux dist could stick the headers elsewhere (only read things quickly). Why don't you just add `\dir\to\kodi\headers\from\pkgconfig\..` to the -I cppflag?

PR #39, opened 2015-06-11
Check htsp version upon connect. Refuse to use servers with HTSP version

ksooo:
not matching minimal HTSP version required by client.

This one gets important once the series recording changes in pvr.hts get merged as those changes require a recent tvheadend (at least 4.0.0, HTSP v20). Staying compatible with ancient tvheadend 3.4 (HTSP v12) forever imo makes no sense and would clutter code with dozens of protocol checks. It's time to tell users it's time for an update of tvheadend, esp. as tvh4 is (kinda?) official release.

I'd like to get this PR in for Isengard's pvr.hts.

### DIFF
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -2,6 +2,7 @@
 - fixed: packaging issues
 - fixed: minor issues found by static analysis
 - added: support for season/episode/part numbers for EPG tags
+- added: check htsp version upon connect. Refuse to use servers with HTSP version not matching minimal HTSP version required by client.
 
 2.1.8
 - added: support for epg subtitles, aka episode names
--- a/pvr.hts/resources/language/English/strings.po
+++ b/pvr.hts/resources/language/English/strings.po
@@ -69,3 +69,9 @@ msgstr ""
 msgctxt "#30201"
 msgid "Trace (detailed)"
 msgstr ""
+
+#empty strings from id 30202 to 30299
+
+msgctxt "#30300"
+msgid "HTSP version mismatch (supported: v%d, required: v%d). Please update tvheadend."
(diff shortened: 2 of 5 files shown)


Jalle19:
Sounds good to me. Openelec will most likely start tracking the 4.0 branch which raises the minimum HTSP version we "need" to support substantially.

ksooo:
Have you thought about updating tvheadend? Latest stable is 4.0.4 and so many things have greatly been enhanced since 3.4.

Anyway, breakage was not intended. Will have a look. Thought, tvh 3.4 supports version 12, because the add-on uses it. Have no idea who bumped required version to 12 but my feeling is that this is not required.

ksooo:
@jalle19 the 12 first appears in the initial revision of pvr.tvh (yes tvh, not hts ) but does not make sense for me as all add-on functionality that uses backend features introduced with htsp version greater 10 (at least in current master) is protected with getprotocol() calls. Will have a closer look today and most probably just lower the required htsp version to 10.

Jalle19:
Yeah I can't even remember how or why that value was raised. It should be safe to limit it to 10.

ksooo:
Okay, will open a PR for this later today.

Jalle19:
Thanks. It might be a good idea to use separate constants for "minimum required server version" and "client version".

ksooo:
Yeah. That was my idea, too.

da-anda:
this is causing a shitload of error messages (kaitoasts) when the backend is down. The version test should only be performed if a running backend was found.

ksooo:
@da-anda you are damn  right. I overlooked this scenario while testing. Will have to fix this. Thanks for reporting.

PR #43, opened 2015-06-18
Revert version check for Isengard

Jalle19:
@ksooo perhaps this is the best solution for now? You said it yourself, it was only in preparation for J**** features anyway.

### DIFF
--- a/pvr.hts/addon.xml
+++ b/pvr.hts/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="2.1.10"
+  version="2.1.9"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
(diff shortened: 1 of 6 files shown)


ksooo:
I dont think we should revert. It's working fine with 2.1.10 with the only glitch that it nerves with kaitoasts if tvheadend server is down, which is a very special situation IMO. Fix for this is trivial and already in the queue (https://github.com/ksooo/pvr.hts/commits/fix-htsp-version-check). Just need to test this (extensively) this evening when I'm back home before I open the PR.

ksooo:
... and I do really very sorry for the trouble I've caused with this. ;-/

Jalle19:
Fixing it is fine by me, although I'm going away for a few days now so you're on your own.

ksooo:
np. Will take care.

ksooo:
Just one question: How do you manage to get one PR into both master and Isengard branch without doing a second PR for Isengard branch? 'git merge' in a shell or is there some github support for this?

Jalle19:
You can cherry pick the commits locally as long as the diff is compatible.

ksooo:
... you mean it is okay to directly work on the kodi-pvr/pvr.hts repo for this case; to push the cherry pick directly to kodi-pvr/pvr.hts? I know that this will work as I have the needed access rights for this repo, just asking beause we're generally not doing this (in favour to PRs).

Jalle19:
You can cherry pick locally to a branch based of Isengard, then PR that. I don't really have a preference between pushing and PRing.

ksooo:
As discussed, we're going to fix this for Isengard.

PR #44, opened 2015-06-18
Fixed: do not show HTSP version mismatch error if tvheadend server is down

ksooo:


ksooo:
... did extensive runtime testing this time, including test against tvh 3.4 and with non-reachable tvh server.

da-anda:
why does it have to show up at all if the backend is down? you can tell Kodi to not show the "connection lost" warnings, but with this you get the version mismatch error still at least once, right?

ksooo:
@da-anda In case the backend is down it does not show up. In case the backend is up and there is a version mismatch it will show up exactly one time.

da-anda:
@ksooo ah, sorry, I missed line 155 when looking at the code ( https://github.com/kodi-pvr/pvr.hts/pull/44/files#diff-34f48ad91ac6c202ac60b0348ae90e30R155 )

PR #53, opened 2015-06-29
Fix CMakeLists.txt language resource folder

hudokkow:


hudokkow:
Since we have to bump add-ons again before RC (see https://github.com/xbmc/xbmc/pull/7362), I'm merging this.

PR #56, opened 2015-07-02
fix bug causing recordings not to display for older TVH

adamsutton:
This was caused by a mistake in the HTSP spec documentation. Also
correctly some un-init vars that could have resulted in undefined
behaviour.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -282,6 +282,7 @@ PVR_ERROR CTvheadend::SendDvrDelete ( uint32_t id, const char *method )
   if (htsmsg_get_u32(m, "success", &u32))
   {
     tvherror("malformed deleteDvrEntry/cancelDvrEntry response: 'success' missing");
+    u32 = PVR_ERROR_FAILED;
   }
   htsmsg_destroy(m);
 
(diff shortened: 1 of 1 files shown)


Jalle19:
Thanks for the fix, I'll handle the backporting.

ksooo:
Just for your information. The retention handling fix is relevant for Isengard only, as we decided to drop support for ancient tvh versions with J******. The many runtime protocol checks needed for this, esp. for autorec, timerec, ... are cluttering the code to much.

adamsutton:
@ksooo personally I think that's a very bad idea. I'm not entirely happy with how new features, which are clearly "optional" (like retention) having been listed as "required" in the HTSP spec. Something I will have to take up with @perexg should I ever get back onto TVH work.

Calling my version of TVH ancient is very much over stating the fact. I admit that I'm quite a bit behind master these days, but I'm still using a non-stable version from when I was running the project which was considerably a head of the "stable" versions of the time and so at most should really only be considered a year or so old. Maybe by some modern standards that's "ancient", but by most sane standards it most certainly isn't. 

What my version is is STABLE. It's been running now for 5 months, and before that about the same again (only down because of a power failure). It does what it needs to, and simply breaking the integration between that version and Kodi just "because" would be a very bad idea!

And more so most of the "extra" HTSP protocol version checks that have been added since I wrote this code are completely redundant. It should be fairly obvious that by definition something added in a future version of HTSP is pretty much by definition "optional", since the previous versions managed quite happily without them.

The only stuff where the version check is required is where for one reason or another, and I was quite careful to limit this, the API changed in a non-backward compatible manner. I.e. the content of some previously existing field actually changed in meaning (there are about 2-3 instances of this).

If you want to drop support for "ancient" versions of HTSP, then really at most you want to be dropping support for anything prior to about v12, which I think was done a long time ago.

OK, rant over!

PR #60, opened 2015-07-11
[Isengard] menu hook timers

Glenn-1990:
@ksooo First of all, thanks for getting serie recordings is kodi master :-)

I'm maintaining a autorec pvr.tvh branch since gotham now.
Me and some friends are using this since gotham without any trouble.

I know this adds a lot of new code, but keep in mind that this new code only runs when the user selects a menu hook. Everything is coded within the epg and timer menu hook calls. 

There won't be a functional change when a user is using the addon as before.

I don't know if this can make it in Isengard, but I'll personally use it until kodi 16, and keep this branch up to date for other users.

Also it is possible to add a timerec with the current code, but there was no way to delete it afterwards.
This PR adds a timer menu hook to delete the complete timer schedule (timerec or autorec).

ping @Jalle19  @ksooo


ksooo:
To be honest, I do not like the idea of having official tvh addons with that much diverged functionality. To me, this sounds like a support nightmare, because users will not understand that Isengard addon is so different from J*** addon or they don't even know or care and start filing bugs just because of the difference.

Sorry, -1 from me.

Glenn-1990:
Ok, no problem, I'll close this.

PR #61, opened 2015-07-14
[debian packaging]: fix library install path

wsnipex:
fixes ubuntu packaging


dreamcat4:
Great! Does this fix mean that master nightlies will start working again?

Jalle19:
Any other that needs changing?

wsnipex:
yeah, all other addons most probably need the same treatment.

PR #67, opened 2015-07-19
include satpos in mux description

Glenn-1990:
mux frequency is not unique, so it makes sense to add the orbital position like most STB's do


ksooo:
Looking good.

Jalle19:
Thanks. In the future I think we should extend the API so addons can report this and more in a better way (e.g. which tuner is used, delivery system type etc.).

PR #68, opened 2015-07-22
API: GetTimerTypes: Heap-allocate TimerType instances to avoid stack overflow

ksooo:
@Memphiz could you please take a look whether this works for you (MacOS)?


ksooo:
@Jalle19 as this fix is pretty important I'm planning a bump of the addon. What's your take on #23 and #67? Shall we shove those two in (given #23 gets rebased in time)? Again, I have not runtime tested #67! I could not really test whether this feature works fine as I only have two tuners atm...

Jalle19:
Shove 'em in, we're in alpha at the moment after all. I haven't runtime tested either since I hadn't upgraded my HTPC to Isengard until a few days ago, which means I'll most likely get around to it soon.

ksooo:
:+1:

Memphiz:
@ksooo will do so once home - looks good in general - thx for beeing so quick here :)

ksooo:
@Memphiz Any good?

Memphiz:
gimme 10 mins

Memphiz:
beer garden temperatures here - had to set my priorities :D

Memphiz:
@ksooo confirmed that this fixes the crash :) - thx

PR #71, opened 2015-07-23
drop CStdString

Jalle19:
@ksooo can you review? I've only build-tested so far.

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -21,6 +21,7 @@
 
 #include "platform/threads/mutex.h"
 #include "platform/threads/atomics.h"
+#include "platform/util/StringUtils.h"
 #include "platform/util/timeutils.h"
 #include "platform/sockets/tcp.h"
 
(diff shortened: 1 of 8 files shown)


src/HTSPConnection.cpp
```diff
@@ -107,23 +107,24 @@ CHTSPConnection::~CHTSPConnection()
...
 {
   va_list va;
-  CStdString auth, url;
   tvheadend::Settings settings = tvh->GetSettings();
 
-  auth = settings.strUsername;
-  if (auth != "" && settings.strPassword != "")
+  // Generate the authentication string (user:pass@)
+  std::string auth = settings.strUsername;
+  if (!auth.empty() && !settings.strPassword.empty())
```
ksooo: I'd prefer <code>if (!(auth.empty() || settings.strPassword.empty()))</code>. Reads IMO more "natural", but leave it as is if you don't feel the same.

src/HTSPConnection.cpp
```diff
@@ -107,23 +107,24 @@ CHTSPConnection::~CHTSPConnection()
...
     auth += ":" + settings.strPassword;
-  if (auth != "")
+  if (!auth.empty())
     auth += "@";
-  url.Format("http://%s%s:%d", auth.c_str(), settings.strHostname.c_str(), settings.iPortHTTP);
+
+  std::string url = StringUtils::Format("http://%s%s:%d", auth.c_str(), settings.strHostname.c_str(), settings.iPortHTTP);
```
ksooo: platform/util/StringUtils.h is not included by the cpp file...

src/Tvheadend.h
```diff
@@ -26,6 +26,7 @@
 #include "platform/threads/threads.h"
 #include "platform/threads/mutex.h"
 #include "platform/util/buffer.h"
+#include "platform/util/StringUtils.h"
```
ksooo: StringUtils not used in the header file. Better move include(s) into the cpp files actually using it.

src/Tvheadend.h
```diff
@@ -488,15 +488,18 @@ class CTvheadend
   }
   inline const char *GetServerName    ( void )
   {
-    return m_conn.GetServerName();
+    static std::string serverName = m_conn.GetServerName();
```
ksooo: really static?

ksooo:
Only some nitpicking here, looking good otherwise. Not runtime-tested, though.

src/Tvheadend.h
Jalle19: I was supposed to move the conversion from `std::string` to char pointers all the way up to `client.cpp` but I guess I forgot. It has to be static at some point, otherwise we'll return a dangling stack pointer if we return `blabla.c_str()`.

Jalle19:
@ksooo I took care of your comments and moved the string to pointer conversion to client.cpp.

src/Tvheadend.h
ksooo: Okay. Sure thing.

ksooo:
@Jalle19 :+1:

PR #73, opened 2015-07-23
Timeshift API: Add support for GetBufferTimeStart() and GetBufferTimeEnd()

ksooo:


FernetMenta:
I don't know much about tvh but I want to give you some info about something I noticed while I did vnsi. GetBufferTimeStart() and friends are called from a different thread than demuxer methods

src/Tvheadend.cpp
```diff
@@ -2283,3 +2283,13 @@ time_t CTvheadend::DemuxGetTimeshiftTime() const
 {
   return m_dmx_active->GetTimeshiftTime();
 }
+
+time_t CTvheadend::DemuxGetTimeshiftBufferStart() const
+{
+  return m_dmx_active.GetTimeshiftBufferStart();
```
Jalle19: `m_dmx_active` is a pointer, at least judging by the method above this one.

src/Tvheadend.h
```diff
@@ -242,6 +242,16 @@ class CHTSPDemuxer
   {
     return (time_t)m_timeshiftStatus.shift;
   }
+  inline time_t GetTimeshiftBufferStart() const
+  {
+    // Note: start/end mismatch is not a bug. tvh uses inversed naming logic here!
+    return static_cast<time_t>(m_timeshiftStatus.end);
```
Jalle19: I'm not sure casting to `time_t` is correct here. While working on some refactoring locally I changed the method above these to return `int64_t` instead. Considering tvheadend reports these values as microseconds it's quite easy to end up truncating the value if casting to a 32-bit value at this stage.

ksooo: Makes sense. Will change this. Thx.

src/Tvheadend.cpp
ksooo: Damn!

src/Tvheadend.h
Jalle19: TBH I have no idea what I was thinking.

ksooo:
@Jalle19 Fixed. Better now?

Jalle19:
Looks reasonable

ksooo:
Took the chance to fix the two Coverity issues introduced by predictive tuning (as a separate commit)...

Jalle19:
Thanks!

PR #72, opened 2015-07-23
Factor out a lot of stuff to separate files

Jalle19:
This is a step in completing the refactoring roadmap we laid out a while ago. Here's a summary of the changes:

* channels, tags, recordings, timers, events and schedules have been factored out into classes living under the `tvheadend::entity` namespace. The naming here is up for debate. Basically an entity is something that can be marked as dirty, so all of these classes extend the base `Entity` class which provides that functionality.

* the new `Schedule` class has a new concept, `Segment`. A segment is a part of the schedule. The class provides a `GetSegment(time_t, time_t)` method to extract events between a specific time interval.

* a place for utilities have been added (`tvheadend::utilities`) which for now contains a file with just a single helper function. The helper provides `erase_if` support for maps, so all the code that loops through things and checks for dirtiness has been compacted into a simple `Utilities::erase_if(...)` call.

* the first two commits are just stuff I found during the way

@ksooo please review if you can, I've only build tested so far. It might help to use `?w=0` when looking at the diff since due to some loops being removed the indentation has changed.

PR #76, opened 2015-07-25
Optimize inactive demuxer packet queue

Jalle19:
This is based on https://github.com/kodi-pvr/pvr.hts/pull/72 so only the last two commits are related.


ksooo:
Very nice. Much appreciated.

Jalle19:
@ksooo after some more testing it seems this PR doesn't really improve anything. The previous approach of keeping 512 packets in the buffer seems to work better. Closing this for now.

PR #77, opened 2015-07-25
add Provides section to debian/control to fix upgrade path from

Jalle19:
Helix versions due to a change in the package name

@wsnipex


Jalle19:
@wsnipex any news on that meta package?

PR #79, opened 2015-07-26
Implement subscription conflict handling

Glenn-1990:
schermafdruk van 2015-10-27 21 58 41

### DIFF
--- a/pvr.hts/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.hts/resources/language/resource.language.en_gb/strings.po
@@ -241,3 +241,43 @@ msgstr ""
 msgctxt "#30502"
 msgid "Streaming profile %s is not available"
 msgstr ""
+
+#empty strings from id 30503 to 30549
+
+#. Streaming conflict management
+
+msgctxt "#30550"
+msgid "TV conflict"
+msgstr ""
+
+msgctxt "#30551"
+msgid "All adapters are in use."
+msgstr ""
+
+msgctxt "#30552"
+msgid "To keep watching TV, you can interrupt the lowest priority service."
+msgstr ""
+
+msgctxt "#30553"
+msgid "To watch TV, you can interrupt the lowest priority service."
+msgstr ""
+
+msgctxt "#30554"
+msgid "This can be an active recording or an other TV client."
+msgstr ""
+
+msgctxt "#30555"
+msgid "Do nothing"
+msgstr ""
+
+msgctxt "#30556"
+msgid "Interrupt service"
+msgstr ""
+
+msgctxt "#30557"
+msgid "Interrupting service... %i%%"
+msgstr ""
+
+msgctxt "#30558"
+msgid "Streaming conflict management"
+msgstr ""
--- a/src/tvheadend/Subscription.cpp
+++ b/src/tvheadend/Subscription.cpp
@@ -31,6 +31,8 @@ Subscription::Subscription(CHTSPConnection &conn) :
   m_weight(SUBSCRIPTION_WEIGHT_NORMAL),
   m_id(0),
   m_state(SUBSCRIPTION_STOPPED),
+  m_prevState(SUBSCRIPTION_STOPPED),
+  m_startTime(time(NULL)),
   m_conn(conn)
 {
 }
@@ -95,9 +97,19 @@ eSubsriptionState Subscription::GetState() const
   return m_state;
 }
 
+eSubsriptionState Subscription::GetPrevState() const
+{
+  CLockObject lock(m_mutex);
+  return m_prevState;
+}
+
 void Subscription::SetState(eSubsriptionState state)
 {
   CLockObject lock(m_mutex);
+  if (state == m_state)
+    return;
+
+  m_prevState = m_state;
   m_state = state;
 }
 
@@ -113,6 +125,18 @@ void Subscription::SetProfile(const std::string &profile)
   m_profile = profile;
 }
 
+time_t Subscription::GetStartTime() const
+{
+  CLockObject lock(m_mutex);
+  return m_startTime;
+}
+
+void Subscription::SetStartTime(time_t time)
+{
+  CLockObject lock(m_mutex);
+  m_startTime = time;
+}
+
 void Subscription::SendSubscribe(uint32_t channelId, uint32_t weight, bool restart)
 {
   /* We don't want to change anything when restarting a subscription */
@@ -122,6 +146,7 @@ void Subscription::SendSubscribe(uint32_t channelId, uint32_t weight, bool resta
     SetWeight(weight);
     SetId(GetNextId());
     SetSpeed(1000); //set back to normal
+    SetStartTime(time(NULL)); // now
   }
 
   /* Build message */
@@ -150,6 +175,10 @@ void Subscription::SendSubscribe(uint32_t channelId, uint32_t weight, bool resta
   htsmsg_destroy(m);
 
   SetState(SUBSCRIPTION_STARTING);
+
+  /* As this might be a pre- posttuning subscription */
+  UpdateStateFromWeight();
+
   tvhdebug("demux successfully subscribed to channel id %d, subscription id %d", GetChannelId(), GetId());
 }
 
@@ -232,17 +261,31 @@ void Subscription::SendWeight(uint32_t weight)
   }
   if (m)
     htsmsg_destroy(m);
+
+  /* As this might be a pre- posttuning subscription now */
+  UpdateStateFromWeight();
+}
+
+void Subscription::UpdateStateFromWeight()
+{
+  if (GetWeight() == static_cast<uint32_t>(SUBSCRIPTION_WEIGHT_PRETUNING) ||
+      GetWeight() == static_cast<uint32_t>(SUBSCRIPTION_WEIGHT_POSTTUNING))
+  {
+    SetState(SUBSCRIPTION_PREPOSTTUNING);
+  }
+  else if (GetState() == SUBSCRIPTION_PREPOSTTUNING)
+  {
+    /* Switched from pre- posttuning to active, initiate a virtual start */
+    SetState(SUBSCRIPTION_STARTING);
+    SetStartTime(time(NULL));
+  }
 }
 
 void Subscription::ParseSubscriptionStatus ( htsmsg_t *m )
 {
   /* Not for preTuning and postTuning subscriptions */
-  if (GetWeight() == static_cast<uint32_t>(SUBSCRIPTION_WEIGHT_PRETUNING) ||
-      GetWeight() == static_cast<uint32_t>(SUBSCRIPTION_WEIGHT_POSTTUNING))
-  {
-    SetState(SUBSCRIPTION_PREPOSTTUNING);
+  if (GetState() == SUBSCRIPTION_PREPOSTTUNING)
     return;
-  }
 
   const char *status = htsmsg_get_str(m, "status");
 
@@ -264,7 +307,13 @@ void Subscription::ParseSubscriptionStatus ( htsmsg_t *m )
       else if (!strcmp("userLimit", error))
         SetState(SUBSCRIPTION_USERLIMIT);
       else if (!strcmp("noFreeAdapter", error))
-        SetState(SUBSCRIPTION_NOFREEADAPTER);
+      {
+        /* If streaming conflict management enabled */
+        if (Settings::GetInstance().GetStreamingConflict())
+          HandleConflict(); // no free adapter, AKA conflict
+        else
+          SetState(SUBSCRIPTION_NOFREEADAPTER);
+      }
       else if (!strcmp("tuningFailed", error))
         SetState(SUBSCRIPTION_TUNINGFAILED);
       else if (!strcmp("userAccess", error))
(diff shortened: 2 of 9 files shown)


src/HTSPDemuxer.cpp
```diff
@@ -630,21 +654,127 @@ void CHTSPDemuxer::ParseSubscriptionSpeed ( htsmsg_t *m )
...
+  {
+    if (error != NULL)
+    {
+      if (!strcmp("badSignal", error))
+        m_subscription.state = SUBSCRIPTION_NOSIGNAL;
+      else if (!strcmp("scrambled", error))
+        m_subscription.state = SUBSCRIPTION_SCRAMBLED;
+      else if (!strcmp("userLimit", error))
+        m_subscription.state = SUBSCRIPTION_USERLIMIT;
+      else if (!strcmp("noFreeAdapter", error))
```
ksooo: Didn't we have some time ago a discussion regarding parsing error strings? My mind has not changed since then. Is it guaranteed that these strings are part of tvheadend's API and that nobody will at any time change these strings or that at one day localized strings will appear here?

Glenn-1990: Yes, but that was the 'status' message, which was a regular string.
Now we use 'subscriptionerror' which represents SM_CODE from tvh in string format, they are not supposed to change at some time in the future, at least that's not why I implemented that.

https://github.com/tvheadend/tvheadend/blob/master/src/htsp_server.c#L3630

ksooo: Hmm. Another piece of code, same story, same design flaw, if you ask me.

Why do you not just return SM_CODE in the tvh API? This would be a much better approach, IMO. What you do here is convert the stuff you translated  in tvh from numerical values into strings back into numerical values (maybe even the same numerically as used in tvh - not checked).

Anyhow, it is like it is now as the actual flaw is in tvh. Maybe you take into account what I said (if you can follow my arguments at all) when you add stuff to the tvh API next time.

Glenn-1990: Well, my first intention was to just sent SM_CODE as integer, but then it came to my mind that the SM_CODE values are just simple defines in tvheadend.h, so they might actually change somewhere in the future. That why I actually added those strings, these strings are not meant to be readable, but only parseble (It even are no sentences).

Only conversion is from int to string on the tvh side.

Seems that I still need this to the hstp specs wiki

ksooo: I have write access to the tvheadend htsp spec in the wiki. Send me the text and I will update the spec. Then (and only then), the strings are really part of the API and I'm fine with this.

ksooo: > Well, my first intention was to just sent SM_CODE as integer, but then it came to my mind that the SM_CODE values are just simple defines in tvheadend.h, so they might actually change somewhere in the future

This is not a matter of data type, it's a matter of putting stuff into the htsp spec. If it's in the spec, it's abug to change it. If it is not in the spec, clients cannot rely on anything, strictly speaking. Again, no matter of data type. Using a string instead of a numeric value does not make things better.

Glenn-1990: @ksooo 

subscriptionError  str   optional   Subscription error code (Added in version 20).

Note: valid values for subscriptionError

noFreeAdapter		No free adapter for this service
scrambled                       Service is scrambled
badSignal			Bad signal status
tuningFailed                    Tuning of this service failed
subscriptionOverridden  Subscription overridden by another one
muxNotEnabled              No mux enabled for this service
invalidTarget                   Invalid target
userAccess		        User does not have acces rights for this service
userLimit                         User limit reached

ksooo: What does"Invalid target" mean? 
What does "User limit reached" mean? 

Could you please extend the descriptions for these two a little to make things more clear?

Glenn-1990: User limit  = each user has a number of maximum streaming connections set in it's user profile, defaulting to unlimited
invalid target = if tvh can not write the actual recording/livestream to the filesystem or when the recording or streaming config is incorrect

Glenn-1990:
@ksooo reworked, might still need some work

I also think that my solution to the no response is not optimal?

src/HTSPDemuxer.cpp
ksooo: Okay. Thanks. Added to the spec now. https://tvheadend.org/projects/tvheadend/wiki/Htsp

Glenn-1990:
@ksooo I already have a better idea for the sendweight issue, we can do that in the dialog thread, so we won't block anything.

@Jalle19 Currently running out of time, so go ahead with re factoring (including SSubscription), I'll rebase later. Actually, if SSubscription has it's own class, things will get easier (i.e. better looking) :p

src/HTSPDemuxer.cpp
```diff
@@ -40,6 +41,24 @@ using namespace std;
 using namespace ADDON;
 using namespace PLATFORM;
 
+bool bDialogForceStart = false;
+
+void DialogLivestreamAborted( void )
+{
+  bDialogForceStart = GUI->Dialog_YesNo_ShowAndGetInput(
+      XBMC->GetLocalizedString(30461),XBMC->GetLocalizedString(30451),
```
xhaggi: spaces between parameter `xx, xx`

src/HTSPDemuxer.cpp
```diff
@@ -40,6 +41,24 @@ using namespace std;
...
+      XBMC->GetLocalizedString(30457),XBMC->GetLocalizedString(30458),
+      XBMC->GetLocalizedString(30459),XBMC->GetLocalizedString(30460));
+}
+
+void DialogLivestreamNostart( void )
+{
+  bDialogForceStart = GUI->Dialog_YesNo_ShowAndGetInput(
+      XBMC->GetLocalizedString(30461),XBMC->GetLocalizedString(30450),
```
xhaggi: same here

src/HTSPTypes.h
```diff
@@ -62,12 +62,28 @@ enum eHTSPEventType
   HTSP_EVENT_REC_UPDATE = 4,
 };
 
-enum eSubscriptionWeight {
-  SUBSCRIPTION_WEIGHT_DEFAULT = 150,
-  SUBSCRIPTION_WEIGHT_PRETUNING = 50,
-  SUBSCRIPTION_WEIGHT_POSTTUNING = 40,
+enum eSubsriptionState
+{
+  SUBSCRIPTION_STOPPED                        = 0,  /* subscription is stopped or not started yet                                                                      */
```
xhaggi: no need for the spaces after the comment

src/HTSPTypes.h
```diff
@@ -62,12 +62,28 @@ enum eHTSPEventType
...
+  SUBSCRIPTION_NOFREEADAPTER_NOSTART          = 4,  /* subscription could not be started because all our adapters are busy (other subscriptions with higher weight)    */
+  SUBSCRIPTION_NOFREEADAPTER_NOSTART_WAITING  = 5,  /* subscription could not be started because all our adapters are busy, waiting for user input                     */
+  SUBSCRIPTION_SCRAMBLED                      = 6,  /* subscription is not running because the channel is scrambled                                                    */
+  SUBSCRIPTION_NOSIGNAL                       = 7,  /* subscription is not running because of a weak/no input signal                                                   */
+  SUBSCRIPTION_USERLIMIT                      = 8,  /* userlimit is reached, so we could not start this subscription                                                   */
+  SUBSCRIPTION_UNKNOWN                        = 9,  /* subscription state is unknown, also used for pretuning and posttuning subscriptions                                                                                   */
```
xhaggi: end up in formatting issues like this ;)

src/Tvheadend.cpp
```diff
@@ -1284,6 +1284,11 @@ void* CTvheadend::Process ( void )
 
   while (!IsStopped())
   {
+    /* Do it here as we don't want to block the htsp RX call */
+    if (m_dmx_active->m_subscription.state == SUBSCRIPTION_FORCERUNNING
+     && m_dmx_active->m_subscription.sendWeightAsync)
+      m_dmx_active->SendWeight();
```
xhaggi: wrong indentation

src/HTSPDemuxer.cpp
```diff
@@ -630,18 +653,112 @@ void CHTSPDemuxer::ParseSubscriptionSpeed ( htsmsg_t *m )
...
+    /* Subscription was running before, but the adapter got stolen by another subscription      */
+    /* Ask user if he wants to continue watching by hijacking an adapter back (increase weight) */
+    std::thread job(DialogLivestreamAborted);
+    job.detach();
+    m_subscription.state = SUBSCRIPTION_NOFREEADAPTER_ABORT;
+  }
+  else if (m_subscription.state == SUBSCRIPTION_STARTING ||
+           m_subscription.state == SUBSCRIPTION_NOFREEADAPTER_NOSTART_WAITING)
+  {
+    /* Conflict case 2:                                                                  */
```
xhaggi: use block comment pattern please
```
/*
 * xxxx
 * yyyy
 */
```

xhaggi:
@Glenn-1990 only looked at formatting.

Glenn-1990:
@ksooo 
I'm working on this at the moment, you suggests to move the functions with a lot of reads and writes to SSubsription to a member of it? Makes sense to me.

I already converted SSubsription from struct to a class, but it's still in htspTypes.cpp, 
should I move it to a separate file instead?

ksooo:
Leave it in HTSPTypes.cpp for now.

src/HTSPTypes.cpp
```diff
@@ -437,3 +441,337 @@ void AutoRecording::SetFulltext(uint32_t fulltext)
...
+}
+
+/* Threadsafe setter for m_state */
+void Subscription::SetState(eSubsriptionState state)
+{
+  m_mutex.Lock();
+  m_state = state;
+  m_mutex.Unlock();
+}
+
```
Glenn-1990: @ksooo Need some advise here.
I want to make m_state thread safe, but I do not know much about multi threading, the only thing I know is that I need to protect variables accessed within multiple threads with mutex locks..

m_state is also accessed in GetState(), but that is an const function and const functions are thread safe?

Does this all make sense? :p

Jalle19: You could take a look at https://github.com/kodi-pvr/pvr.hts/blob/master/src/AsyncState.cpp and https://github.com/kodi-pvr/pvr.vbox/blob/master/src/vbox/StartupStateHandler.h. Your basic idea is correct, you need to use a mutex lock whenever you access the `m_state` variable. `const' functions are not thread-safe by definition since another thread could come in and overwrite the value of the variable while you're reading it. Since locking a mutex is a non-const operation it is common to mark the mutex variable as mutable so it can be used from `const` methods (see e.g. https://github.com/kodi-pvr/pvr.vbox/blob/master/src/vbox/StartupStateHandler.h#L123).

src/HTSPTypes.cpp
```diff
@@ -437,3 +441,378 @@ void AutoRecording::SetFulltext(uint32_t fulltext)
...
+uint32_t Subscription::GetSpeed() const
+{
+  return m_speed;
+}
+
+uint32_t Subscription::GetWeight() const
+{
+  /* used in multiple threads */
+  CLockObject lock(m_mutex);
+  return m_weight;
```
Glenn-1990: @Jalle19 thanks for your help on this.

m_weght, m_state and m_subscriptionId are made threadsafe in the right way I guess? :p

Glenn-1990:
@Jalle19 @ksooo 
Should be in working state now.

Any feedback would be welcome ;-)

pvr.hts/resources/language/resource.language.en_gb/strings.po
```diff
@@ -163,3 +163,57 @@ msgstr ""
...
+#empty strings from id 30403 to 30449
+
+#. Subscription states representation
+
+msgctxt "#30450"
+msgid "No free adapter available"
+msgstr ""
+
+msgctxt "#30451"
+msgid "Livestream aborted, adapter stolen by other subscription"
```
Jalle19: Can you change "Livestream" to "Streaming" or "Live stream"?

pvr.hts/resources/language/resource.language.en_gb/strings.po
```diff
@@ -163,3 +163,57 @@ msgstr ""
...
+msgstr ""
+
+#. Subscription conflict dialog
+
+msgctxt "#30457"
+msgid "Do you want to increase the priority to hijack an adapter?"
+msgstr ""
+
+msgctxt "#30458"
+msgid "WARNING: this can abort an active recording or livestream"
```
Jalle19: Continuing on the previous comment I think "live stream" is better after all.

src/HTSPTypes.h
```diff
@@ -243,6 +269,53 @@ typedef std::map<std::string, AutoRecording> AutoRecordingsMap;
...
+  uint32_t GetWeight() const;
+  int32_t  GetSpeed() const;
+
+  void SendSubscribe(uint32_t channelId , uint32_t weight , bool force = false);
+  void SendUnSubscribe();
+  bool SendSeek(int time);
+  void SendSpeed(int32_t speed, bool force = false);
+  void SendWeight(uint32_t weight);
+
+  void ParseSubscriptionStatus(htsmsg_t *m);
```
Jalle19: I'm not sure I like the semantics of having the subscription class being responsible for sending HTSP commands, but since you've already changed it's fine (unless someone else agrees with me).

src/HTSPTypes.cpp
```diff
@@ -437,3 +441,394 @@ void AutoRecording::SetFulltext(uint32_t fulltext)
...
+     * This to prevent the dialog from popping up when zapping
+     */
+    sleep(DIALOG_NOSTART_DELAY);
+
+    /* Break if the conflict was solved */
+    if (IsConflictSolved(initialId))
+      return;
+  }
+
+  GUI->Lock();
```
Jalle19: @FernetMenta is it correct to lock like this? We haven't used the GUI helper in this addon before so it's completely new to me.

FernetMenta: why do you want to grab graphicsContext lock? would you do this for showing a dialog inside Kodi?

Jalle19: See the subsequent lines, a dialog is shown asking for user input. The lock is released after that call. Are you saying no locking is required?

FernetMenta: > Are you saying no locking is required

yes.

Jalle19:
@Glenn-1990 any chance you could split the Subscription class refactoring to a separate PR?

Glenn-1990:
@Jalle19 where should it be placed? Root directory or entity?

src/HTSPTypes.h
Glenn-1990: I don't know what fits the best, but autorecording and timerecording classes are also sending HTSP commands?

Glenn-1990: And at least it would be nice if I could change the subscription weight inside this class (i.e. sendweight)
Needed we we want to increase weight async..

Jalle19:
I'd place it either in tvheadend/, it's not really an entity since it doesn't require SetDirty().

Glenn-1990:
I'll rebase this when https://github.com/kodi-pvr/pvr.hts/pull/119 is in

Glenn-1990:
Maybe we should have this as a default disabled setting?

src/tvheadend/Subscription.cpp
```diff
@@ -264,7 +268,30 @@ void Subscription::ParseSubscriptionStatus ( htsmsg_t *m )
       else if (!strcmp("userLimit", error))
         SetState(SUBSCRIPTION_USERLIMIT);
       else if (!strcmp("noFreeAdapter", error))
-        SetState(SUBSCRIPTION_NOFREEADAPTER);
+      {
+        /* If streaming conflict management enabled */
+        if (Settings::GetInstance().GetStreamingConflict())
+        {
```
Jalle19: I would factor out this block to a separate method, e.g. `HandleSubscriptionConflict`.

Glenn-1990: done

Glenn-1990:
@Jalle19 What do you think about the English strings?

ksooo:
If you ask me, this stuff is all far too tekky for a normal end user. As I said several times, why don't you just buy a server that fits your needs instead of adding loads of code just dealing with the situation that your server does not have the balls to serve your needs? Not 100% serious about this, but cannot understand your priorities... ;-) 

I really do not like that you add some real GUI (dialog) to pvr.hts, because I thinks this is no good design.

But again, just my 2 Ct, I will not block this PR.

Glenn-1990:
Well, I'm not facing such a streaming conflict many times as I've a quad tuner, but when I face it, it seems unclear to the user what is happening...
So this dialog makes it clear that the channel he was watching was stopped because of a conflict with a starting recording or other livestream and gives him the ability to keep watching, which is a basic conflict feature most pvr apps have. (pvr.myth also does conflict handling)

I'm not adding a GUI dialog, I just call the kodi one, so no messing with xml files.
This is where the GUI api is for not? For addon specific stuff.

Maybe we can try to make it simpler?
i.e.
dialog title: live stream conflict
All adapters are in use.
To keep watching, interrupt an other client or recording.   
buttons: interrupt, cancel

This is a very good example, page 2: http://www.connectctc.com/files/tv-support/Prime-User-Guide/Conflicts/How-to-Resolve-Conflicts.pdf

ksooo:
Yeah. Simplify it. Especially untekkify the wording. Maybe like in the document you posted. Everything about"stolen adapters" and "priorities" and "hijack" has to go imo, because people will not understand, I'm afraid.

Regarding using a dialog. I still have the feeling that conflict handling belongs into Kodi core even though this might be much more work than just hacking this into one add-on.

Jalle19:
I agree about the techiness - the feature is nice though. I don't think it can be inplemented in core Kodi either since very few addons provide the necessary API to even determine if a conflict has taken place.

ksooo:
I tend to disagree regarding implementing this not in Kodi core. Conflict management is an essential PVR feature and therefore should be handled by Kodi core in a consistent easy for ask add-ons. The fact that not many servers (today!) offer a suitable API for this is imo unrelated. 

But again, I'm not going to block this PR if the UI will be changed to be far less techy.

Glenn-1990:
schermafdruk van 2015-10-27 21 58 41

ksooo:
Is there any chance to present to the user what other conflicting subscriptions are active and let him choose from a list which one to interrupt in order to solve the conflict? This would be cool (and users will understand what's going on).

BTW, if I remember correctly, since Isengard we call it just "TV" everywhere in Kodi, not "Live TV", even in English.

Glenn-1990:
@ksooo Currently that's not possible, or we have to parse the active recordings and check the start time, but that's not a clean solution...

Glenn-1990:
@ksooo what if we just show a dialog like this?:
"All services are in use, please disable active recordings and/or other clients to resume watching"
-> OK button

And nothing else,  so not changing weight etc, just make the user clear what just happened to him.

PR #80, opened 2015-07-27
debian packaging: provide transitional package for deprecated kodi-pvr-tvheadend-hts

wsnipex:
replaces https://github.com/kodi-pvr/pvr.hts/pull/77
needs backport to Isengard

team nightly ppa already has an updated hts package

@Jalle19 ping


Jalle19:
Awesome, thank you very much!

PR #81, opened 2015-07-27
debian packaging: provide transitional package for deprecated kodi-pvr-tvheadend-hts

Jalle19:
Backport to Isengard.

@wsnipex is the package versioning correct?


wsnipex:
seems to need to rebase, version was already bumped. Otherwise its fine

Jalle19:
@wsnipex yeah it seems I based it on an old head. Can you check again, I bumped the version one notch.

wsnipex:
looks good

Jalle19:
So now the stable PPA needs updating and we're good?

PR #83, opened 2015-08-04
Minimal support for PVR API 3.0.0

metaron-uk:
Addon part of https://github.com/xbmc/xbmc/pull/7626
Changes as described in the commit messages.

EDIT: Previous concerns all now addressed thanks to recently merged commits.


ksooo:
Despite from the time conversion change (which I'm uncertain about) looking all good to me. Thanks.

PR #85, opened 2015-08-05
Timerec,Autorec: Fix server local time to UTC conversion.

ksooo:
Runtime-tested by me, @metaron-uk and @PhobosK.

@Jalle19 any objections?

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -93,7 +93,8 @@ void *CHTSPRegister::Process ( void )
 CHTSPConnection::CHTSPConnection ()
   : m_socket(NULL), m_regThread(this), m_ready(false), m_seq(0),
     m_serverName(""), m_serverVersion(""), m_htspVersion(0),
-    m_webRoot(""), m_challenge(NULL), m_challengeLen(0), m_suspended(false)
+    m_webRoot(""), m_challenge(NULL), m_challengeLen(0), m_suspended(false),
+    m_serverTimezone(0)
 {
 }
 
@@ -159,6 +160,12 @@ std::string CHTSPConnection::GetServerString ( void )
   CLockObject lock(m_mutex);
   return StringUtils::Format("%s:%d [%s]", settings.strHostname.c_str(), settings.iPortHTSP,
              m_ready ? "connected" : "disconnected");
+}
+
+int32_t CHTSPConnection::GetServerTimezone( void )
+{
+  CLockObject lock(m_mutex);
+  return m_serverTimezone;
 }
 
 bool CHTSPConnection::HasCapability(const std::string &capability) const
@@ -470,8 +477,32 @@ bool CHTSPConnection::SendAuth
   return (msg != NULL);
 }
 
+bool CHTSPConnection::SendGetSysTime ( void )
+{
+  /* Build message */
+  htsmsg_t *msg = htsmsg_create_map();
+
+  /* Send and Wait */
+  if (!(msg = SendAndWait0("getSysTime", msg)))
+    return false;
+
+  /* Process */
+  int32_t s32;
+  if (htsmsg_get_s32(msg, "timezone", &s32))
+  {
+    tvherror("malformed getSysTime response: 'timezone' missing");
+    htsmsg_destroy(msg);
+    return false;
+  }
+  else
+    m_serverTimezone = s32;
+
+  htsmsg_destroy(msg);
+  return true;
+}
+
 /**
- * Register the connection, hello+auth
+ * Register the connection, hello+auth+systime
  */
 void CHTSPConnection::Register ( void )
 {
@@ -502,6 +533,12 @@ void CHTSPConnection::Register ( void )
     tvhdebug("sending auth");
     
     if (!SendAuth(user, pass))
+      goto fail;
+
+    /* Send GetSysTime */
+    tvhdebug("sending getsystime");
+
+    if (!SendGetSysTime())
       goto fail;
 
     /* Rebuild state */
--- a/src/HTSPTypes.cpp
+++ b/src/HTSPTypes.cpp
@@ -270,17 +270,23 @@ void RecordingBase::SetChannel(uint32_t channel)
 // static
 time_t RecordingBase::LocaltimeToUTC(int32_t lctime)
 {
-  /* Note: lctime contains minutes from midnight (up to 24*60) as local time. */
+  /* Note: lctime contains minutes from midnight server local time. */
 
-  /* complete lctime with current year, month, day, ... */
+  /* Obtain time_t for current day midnight GMT */
   time_t t = time(NULL);
-  struct tm *tm_time = localtime(&t);
+  struct tm *tm_time = gmtime(&t);
+  tm_time->tm_hour = 0;
+  tm_time->tm_min  = 0;
+  tm_time->tm_sec  = 0;
+  t = mktime(tm_time);
 
-  tm_time->tm_hour  = lctime / 60;
-  tm_time->tm_min   = lctime % 60;
-  tm_time->tm_sec   = 0;
+  /* Add server timezone (hours) to t to get a time_t for midnight server local time */
+  t += tvh->GetServerTimezone() * 60 * 60;
 
-  return mktime(tm_time);
+  /* Add minutes past midnight server local time */
+  t += lctime * 60;
+
+  return t;
 }
 
 // static
--- a/src/Tvheadend.h
+++ b/src/Tvheadend.h
@@ -191,7 +191,8 @@ public:
   std::string GetServerName    ( void );
   std::string GetServerVersion ( void );
   std::string GetServerString  ( void );
-  
+  int32_t     GetServerTimezone( void );
+
   bool        HasCapability(const std::string &capability) const;
 
   inline bool IsConnected       ( void ) const { return m_ready; }
@@ -208,6 +209,7 @@ private:
   bool        ReadMessage      ( void );
   bool        SendHello        ( void );
   bool        SendAuth         ( const std::string &u, const std::string &p );
+  bool        SendGetSysTime   ( void );
 
   PLATFORM::CTcpSocket               *m_socket;
   PLATFORM::CMutex                    m_mutex;
@@ -221,6 +223,7 @@ private:
   std::string                         m_webRoot;
   void*                               m_challenge;
   int                                 m_challengeLen;
+  int32_t                             m_serverTimezone; // hours west of GMT.
 
   CHTSPResponseList                   m_messages;
   std::vector<std::string>            m_capabilities;
@@ -520,6 +523,10 @@ public:
   std::string GetServerString  ( void )
   {
     return m_conn.GetServerString();
+  }
+  int32_t GetServerTimezone    ( void )
+  {
+    return m_conn.GetServerTimezone();
   }
   inline int GetProtocol ( void ) const
   {


Jalle19:
I think we should test with the server and client in different timezones. I can send you credentials to a GMT+2 summertime (so GMT+3) server if anyone wants to do a runtime test.

ksooo:
@Jalle19 sounds good. Please send me the credentials.

metaron-uk:
I've been thinking and am now unsure about the server timezone response.

I got a reply of "-1" from my server (HTS Tvheadend 4.1-210~g4d5a4eb-dirty) which I took to mean GMT-1 and adjusted the code to fit but actually my timezone should be GMT+1 at the moment.

The appropriate line in the TVHeadend code is https://github.com/tvheadend/tvheadend/blob/5c3b74df10e86181f8b95702ddc9e68aadca5abf/src/htsp_server.c#L1092 I think.

Also, a test with server and client is likely to fail because 'localtime' is used when converting Kodi time to send to HTSP. (https://github.com/kodi-pvr/pvr.hts/blob/master/src/AutoRecordings.cpp#L182)
```LocaltimeToUTC```  only gets called converting HTSP to Kodi time, so really we need a ```UTCtoServerLocalTimeMinutesPastMidnight``` function for the reverse translation instead of using ```localtime```.

**Comment by Jalle19** [comment]
@ksooo sent

File: src/HTSPConnection.cpp [hunk 1]
```diff
@@ -484,6 +515,12 @@ void CHTSPConnection::Register ( void )
     if (!SendAuth(user, pass))
       goto fail;
 
+    /* Send GetSysTime */
+    tvhdebug("sending getsystime");
```
Jalle19: Should probably be `tvhtrace`.

ksooo:
@metaron-uk yes, I'm also very uncertain (especially because I'm not an expert with time conversions), that's why I asked you yesterday whether you're thinking all is well. ;-)

My local tvh server also gives me a timezone of "-1", but in Hamburg we're currently having CEST, which means GMT+2. But isn't it the case that as long we're doing server-client time conversion right, it shouldn't actually matter what the server sends as long as he internally handles time conversion correct?

> Also, a test with server and client is likely to fail because 'localtime' is used when converting Kodi time to send to HTSP.

Yes, that's definitely an issue. We need a conversion UTC -> server local time here.

ksooo:
@Jalle19 Credentials received. Thanks. Will do some testing later today / tonight (CEST ;-).

src/HTSPConnection.cpp
ksooo: It's consistent with the other "sending ..." logs some lines above.

metaron-uk:
@ksooo - sorry about giving you duff information yesterday. I don't often make assertions I can't back up fully, but every now and again I realize I did and I try and make amends if I can.
If we're both getting -1 in different timezones there is definitely something wrong.

If everything is sent in GMT between client and server you are correct, the conversion doesn't matter as long as the server gets it right. Unfortunately as we're converting GMT to server localtime in the client we need to know the correct offset or we'll send it the wrong time :-(

ksooo:
Some investigations done:

<pre>
#include <sys/time.h>
#include <stdio.h>

int main ()
{

  struct timeval tv;
  struct timezone tz;

  if (gettimeofday(&tv, &tz) == -1)
  {
    printf("gettimeofday failed\n");
    return -1;
  }

  printf("timezone: %d minutes west of GMT\n", tz.tz_minuteswest);
  printf("dst: %d\n", tz.tz_dsttime);

  return 0;
}
</pre>

Output on my local Ubuntu 14.04 LTS (x86_64):
<pre>
kai@kso-desktop:~/src/github$ ./gettimeofday 
timezone: -120 minutes west of GMT
dst: 0
</pre>

Output on my local OpenELEC 3.95.5 (x86_64):
<pre>
OpenELEC:~/downloads # ./gettimeofday 
timezone: -60 minutes west of GMT
dst: 0
</pre>

Hmm, don't know waht to make out of this...

metaron-uk:
On my box:
andrew@blackbox ~ $ ./gettimeofday 
timezone: -60 minutes west of GMT
dst: 0
andrew@blackbox ~ $ date
Thu  6 Aug 09:47:41 BST 2015
andrew@blackbox ~ $ date -u
Thu  6 Aug 08:47:56 UTC 2015

So a return of -1 to the tvheadend function call is as expected. 
What does date and date -u give you?

src/HTSPConnection.cpp
Jalle19: Okay, in that case it's fine.

ksooo:
Ubuntu 14.04:
<pre>
kai@kso-desktop:~$ date
Do 6. Aug 10:52:17 CEST 2015
kai@kso-desktop:~$ date -u
Do 6. Aug 08:52:21 UTC 2015
</pre>

OpenELEC 5.95.3
<pre>
OpenELEC:~ # date
Thu Aug  6 10:53:52 CEST 2015
OpenELEC:~ # date -u
Thu Aug  6 08:53:56 UTC 2015
</pre>

Jalle19:
Shouldn't tvheadend return all times in GMT? Kodi should also only accept GMT times, right?

ksooo:
> Shouldn't tvheadend return all times in GMT?

It doesn't, unfortunately. In the spec, there are several "Minutes from midnight" and these are unfortunately local time. Even if we consider this a bug, fixing on tvheadend side would require a bump of the minimal HTSP version pvr.hts accepts.

> Kodi should also only accept GMT times, right?

 Unfortunately not. ^^^^

metaron-uk:
On my Pi running a custom compiled openelec (5.0.8 Kodi Helix based) I get:
TestPi:~ # date
Thu Aug  6 10:22:56 BST 2015
TestPi:~ # date -u
Thu Aug  6 09:22:58 UTC 2015
TestPi:~ # /storage/gettimeofday 
timezone: 0 minutes west of GMT
dst: 0

Maybe it's an OpenElec 'feature'?

ksooo:
OpenELEC uses Busybox, if this of any interest...

ksooo:
> Maybe it's an OpenElec 'feature'?

@stefansaraev any thoughts on this?

ksooo:
@metaron-uk to sum up: 
* current implementation as present in this PR of <code>RecordingBase::LocaltimeToUTC</code> is considered okay.
* conversion UTC -> server local time is missing in <code>AutoRecordings::SendAutorecAdd</code> and <code>TimeRecordings::SendTimerecAdd</code>
* OpenELEC (at least for me) behaves strange wrt to gettimeofday (or my system is somehow misconfigured).

By any chance, could you take a look at point 2 and ideally come up with an implementation as it seems that I cannot test because of my OE acting up (no other test boxes avail atm).

ksooo:
On second thought, even the implementation of <code>RecordingBase::LocaltimeToUTC</code> is currently, as is in this PR, incorrect, right?

metaron-uk:
OK, I will take a look at a UTC to server local conversion for ```AutoRecordings::SendAutorecAdd``` and ```TimeRecordings::SendTimerecAdd```

The current 4c29351 patch is 'correct' IMO, but my OE installation appears to have the same problem as yours: tz_minuteswest returned by gettimeofday doesn't include the daylight saving time offset, whereas on my gentoo box and your ubuntu one it does.
As getttimeofday appears to be a libc function (not busybox), I looked for obvious differences, but both my main box and the pi are running glibc 2.20 versions. This will need more investigation.

Personally I expect more users of pvr.hts to be running tvheadend on OE in the same timezone than running tvheadend (not on OE) in a different timezone.
It may be better to use my original patch for now (@PhobosK said the metaron-uk/pvr.hts@ea8ba20 patch worked for him) and fix the lack of support for a server in a different timezone issue later once we work out a way around this OE oddity.

What do you think?

ksooo:
@metaron-uk sounds like a good idea for now. Can you please file your original fix as separate PR, including 2.2.5 changelog change?

metaron-uk:
OK will do.
I've just checked rasbian (aka debian). That seems to ignore DST the same way OE does.
Suspect this may be a 'GNU / distribution' issue...

metaron-uk:
OK here's the answer:
http://stackoverflow.com/questions/635780/why-does-glibc-timezone-global-not-agree-with-system-time-on-dst
Looks like tvheadend needs a 'fix' before we can rely on the result of ```getsystime``` :-(.

ksooo:
I was afraid that you will find out exactly this. Thanks for investigating.

How about you filing a PR (inluding HTSP version bump) for tvheadend? Just kindly asking, in case you have time in your hands for this...

metaron-uk:
;-)

metaron-uk:
@ksooo ... and you suggested I was cheeky ;-). Anyhow it's been a pleasure working with you on this.
Now to get some of my changes into the pvr.mythtv client .....
I don't think that will be quite as simple for some reason, but then it *is* still my backend of choice.

ksooo:
@metaron-uk 
> ... and you suggested I was cheeky

No, it was the other way around. You called yourself cheeky. <grin>
I was kindly asking you to do me a favour as it seems to me that you were actually knowing what you were talking about... Great stuff! Thanks man. Was a pleasure working with you. We can do this anytime (anytime? dejavu!) again.

Glenn-1990:
For openelec, try changing the timezone to another one, change it back to yours, and reboot.

ksooo:
@Glenn-1990 
> For openelec, try changing the timezone to another one, change it back to yours, and reboot.

Not related to this problem here.

Jalle19:
@ksooo is this still valid?

ksooo:
@Jalle19 problem still exists. Once tvh contains a proper fix (https://github.com/tvheadend/tvheadend/pull/700) has to be adapted to the related tvh API change.

Jalle19:
The upstream patch was just merged.

ksooo:
Yep, now tvheadend is fixed, but there is another problem. To deny connection to all servers not containing the fix is definitely not a good idea. This would bump the minimum required server version to bleeding edge and I don't know when an official release containing the fix will arise. Even worse, in most (the normal) cases, where server and client are actually in the same time zone this fix is not needed at all.

How to deal with this? A addon setting "Deny connection to servers with unknown time zone" which is unchecked by default and must be checked by the user to activate the server timezone check? If checked,  the client will try to obtain the timezone from server and in case it does not contain the fix, an error message will be displayed and the connection closed after that?

@metaron-uk @Jalle19 others: What do you think

stefansaraev:
@ksooo sorry I somehow missed your ping here.

on openelec. gettimeofday() works fine. busybox has nothing to do with it.

Jalle19:
@ksooo I can't even remember what problem this is supposed to solve so it's hard to comment on how to move forward.

ksooo:
> on openelec. gettimeofday() works fine

@stefansaraev really? => https://github.com/kodi-pvr/pvr.hts/pull/85#issuecomment-128293040

ksooo:
@Jalle19 the problem is that some tvheadend htsp methods require values in "minutes from midnight" where "midnight" is actually server-local midnight! Thus, the client needs to know the exact timezone offset in order to do the required UTC (that Kodi provieds/wants) <-> "minutes from midnight" (that tvh provides/wants at some places) conversions right.

=> https://tvheadend.org/projects/tvheadend/wiki/Htsp => addAutorecEntry, addTimerecEntry

stefansaraev:
> @stefansaraev really? => #85 (comment)

yes, really. no clue what you are doing. works for me (yes, your example code, before you ask, I took the time to test it)

ksooo:
@stefansaraev thanks for testing. Strange... 

Maybe this gives you a hint? http://stackoverflow.com/questions/635780/why-does-glibc-timezone-global-not-agree-with-system-time-on-dst

Maybe @metaron-uk can explain this better than I can?

Jalle19:
@ksooo I see, it's too bad a convention like that has snuck into the API. Sanity would be much better preserved if everything was UTC all the way.

ksooo:
@stefansaraev I'm not an expert wrt time conversion and don't have a clue what exactly is broken on the tvh server side - may it be a bug in tvheadend, in OE, Ubuntu, ... or just some misconfiguration ... don't know ... and actually don't really care. Just had the idea that this by any chance could be a bug in OE and wanted to give you a chance to sneak into it. ;-)

However, I assume https://github.com/tvheadend/tvheadend/pull/700 fixes the problem for tvheadend (have not yet tested it) - that's all I'm interested in atm - to be honest.

@Jalle19 IMO the tvh API is broken and should be fixed to consequently use UTC, but this would still not solve the timezone problem we currently have with pvr.hts  - either to make pvr.hts' minimum required tvheadend version upcoming 4.1 (whenever this will be) or to add some manual workaround (user setting). The former is not a good idea, imo. I'd like to hear opinions about the user setting approach or - even better - other ideas for solving the problem. 

Or we actively decide not to care, because this is really a rather exceptional use case. Consequence would "just" be autorecs and timerecs not working as expected if cleint and server are in different time zones. ;-)

Jalle19:
I opt for the "don't care" approach for now since I really don't understand the code that well anyway. We can consider a fix again once a user actually reports this as a bug.

ksooo:
Okay. I'm perfectly fine with this.

metaron-uk:
Just to be clear, the current code works fine as long as the tvheadend server and kodi client are set to the same timezone. This will be 99.9% of all cases as far as I can see, unless someone is running a public tvheadend server everyone wants to connect to for some reason.
If you ever wanted to support server and client in different timezones, the code would need to change, in which case I would suggest using the new 'gmtoffset' parameter rather than the 'timezone' parameter.

ksooo:
Sure. This and given the high effort for implementing support for server and client in different time zones is exactly why we decided to do nothing about it, atm.

ksooo:
Well, not exactly high effort, but more the fact that it is not trivial to teach the users the actual problem and how they have to deal with it. -> Some user setting they hardly will understand or denying connections to every tvh server that does not send the new gmtoffset field. Grr.

metaron-uk:
What about: Use gmtoffset if present, fallback to kodi client localtime if not with log entry 'not a problem unless your server is in a different timezone'?

ksooo:
This would be exactly what I mean with "does a normal user not understand". ;-)

PR #82, opened 2015-07-29
Predictive tuning fixes

Glenn-1990:
Found some issues with predictive tuning when working on the subscription conflict stuff.

PR #86, opened 2015-08-05
Autorec: Fix handling of start end stop time.

ksooo:
Runtime-tested by me and @PhobosK.

@Jalle19 Any objections?

### DIFF
--- a/src/AutoRecordings.cpp
+++ b/src/AutoRecordings.cpp
@@ -172,22 +172,22 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
   if (strcmp(timer.strDirectory, "/") != 0)
     htsmsg_add_str(m, "directory", timer.strDirectory);
 
-  /* Note: not sending any of the following for values will be interpreted by tvh as "any". */
+  /* Note: not sending any of the following three values will be interpreted by tvh as "any". */
   if (timer.iClientChannelUid >= 0)
     htsmsg_add_u32(m, "channelId", timer.iClientChannelUid);
 
   if (timer.startTime > 0)
   {
-    /* approx start time in minutes from midnight. tvh will match value +- 15 mins. */
+    /* Exact start time (minutes from midnight). */
     struct tm *tm_start = localtime(&timer.startTime);
-    htsmsg_add_s32(m, "approxTime", tm_start->tm_hour * 60 + tm_start->tm_min);
+    htsmsg_add_s32(m, "start", tm_start->tm_hour * 60 + tm_start->tm_min);
   }
 
-  if ((timer.endTime - timer.startTime) > 0)
+  if (timer.endTime > 0)
   {
-    /* min/max duration: endtime -+ 15 min */
-    htsmsg_add_u32(m, "minDuration", timer.endTime - timer.startTime - 15 * 60);
-    htsmsg_add_u32(m, "maxDuration", timer.endTime - timer.startTime + 15 * 60);
+    /* Exact stop time (minutes from midnight). */
+    struct tm *tm_stop = localtime(&timer.endTime);
+    htsmsg_add_s32(m, "startWindow", tm_stop->tm_hour * 60 + tm_stop->tm_min);
   }
 
   /* Send and Wait */
@@ -289,26 +289,6 @@ bool AutoRecordings::ParseAutorecAddOrUpdate(htsmsg_t *msg, bool bAdd)
     return false;
   }
 
-  if (!htsmsg_get_u32(msg, "minDuration", &u32))
-  {
-    rec.SetMinDuration(u32);
-  }
-  else if (bAdd)
-  {
-    tvherror("malformed autorecEntryAdd: 'minDuration' missing");
-    return false;
-  }
-
-  if (!htsmsg_get_u32(msg, "maxDuration", &u32))
-  {
-    rec.SetMaxDuration(u32);
-  }
-  else if (bAdd)
-  {
-    tvherror("malformed autorecEntryAdd: 'maxDuration' missing");
-    return false;
-  }
-
   if (!htsmsg_get_u32(msg, "retention", &u32))
   {
     rec.SetRetention(u32);
@@ -339,13 +319,23 @@ bool AutoRecordings::ParseAutorecAddOrUpdate(htsmsg_t *msg, bool bAdd)
     return false;
   }
 
-  if (!htsmsg_get_s32(msg, "approxTime", &s32))
+  if (!htsmsg_get_s32(msg, "start", &s32))
   {
     rec.SetStart(s32);
   }
   else if (bAdd)
   {
-    tvherror("malformed autorecEntryAdd: 'approxTime' missing");
+    tvherror("malformed autorecEntryAdd: 'start' missing");
+    return false;
+  }
+
+  if (!htsmsg_get_s32(msg, "startWindow", &s32))
+  {
+    rec.SetStop(s32);
+  }
+  else if (bAdd)
+  {
+    tvherror("malformed autorecEntryAdd: 'startWindow' missing");
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


ksooo:
@Jalle19 we have not yet released 2.2.5 although addon.xml in master is 2.2.5 for quite some time now, right? Latest version in Kodi master is 2.2.4

ksooo:
I'd like to officially release 2.2.5 the next days.
Despite of the stuff already in master I want #85 and #86 merged for 2.2.5
What about #82 and #72?

ksooo:
Oh, I've just seen this: https://github.com/xbmc/xbmc/pull/7702 
Cool thing, we can stick with 2.2.5 for awhile I guess. ;-)

Jalle19:
I'd hold off on #72 for now, it seems to introduce a performance regression when pushing EPG data to Kodi. Will have to test some more.

src/AutoRecordings.cpp
```diff
@@ -172,22 +172,22 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
...
 
-  if ((timer.endTime - timer.startTime) > 0)
+  if (timer.endTime > 0)
   {
-    /* min/max duration: endtime -+ 15 min */
-    htsmsg_add_u32(m, "minDuration", timer.endTime - timer.startTime - 15 * 60);
-    htsmsg_add_u32(m, "maxDuration", timer.endTime - timer.startTime + 15 * 60);
+    /* Exact stop time (minutes from midnight). */
+    struct tm *tm_stop = localtime(&timer.endTime);
+    htsmsg_add_s32(m, "startWindow", tm_stop->tm_hour * 60 + tm_stop->tm_min);
```
Glenn-1990: I bit late but this seems messing thing up...

End time should reflect to endWindow instead of startWindow, as we don't have an endWindow we should use maxduration for this..

src/AutoRecordings.cpp
```diff
@@ -172,22 +172,22 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
...
   if (timer.iClientChannelUid >= 0)
     htsmsg_add_u32(m, "channelId", timer.iClientChannelUid);
 
   if (timer.startTime > 0)
   {
-    /* approx start time in minutes from midnight. tvh will match value +- 15 mins. */
+    /* Exact start time (minutes from midnight). */
     struct tm *tm_start = localtime(&timer.startTime);
-    htsmsg_add_s32(m, "approxTime", tm_start->tm_hour * 60 + tm_start->tm_min);
+    htsmsg_add_s32(m, "start", tm_start->tm_hour * 60 + tm_start->tm_min);
```
Glenn-1990: When using approxtime, tvh will records shows starting 15 mins before or after this.

If we set "start" to our actual start time, there will be no margin, if a broadcaster decides to show it 5 min in advance, the show will not be recorded :s

Glenn-1990:
@ksooo What was the actual problem that should be fixed with this?

ksooo:
> What was the actual problem that should be fixed with this?

https://github.com/kodi-pvr/pvr.hts/pull/38#issuecomment-127954190

Glenn-1990:
Ok, I see the problem but it's still not fixed as it should be.

We are assigning end time to 'start before' in tvheadend, which represents the max start time..

the proper way would be:
-for start time: 
hstp "start" = (kodi start time - diff)
hstp "startWindow" = (kodi start time + diff)
If we make diff=15 mins, we get the same behavior as "approxTime"

-for end time:
htsp "maxDuration" = (kodi end time - kodi start time) + diff
and additionally, htsp "minDuration" = (kodi end time - kodi start time) - diff

ksooo:
@Glenn-1990 I agree that we're currently misusing <code>startWindow</code> as it does not denote the end time, but latest start time. 

As I read it, what you're suggesting is functionally exactly what we had before this PR (plus getting rid of deprecated approxTime) and after all I think this is the correct implementation. I was also mislead by the HTSP docs when claiming the original implementation buggy, but have to revise myself.

@PhobosK what do you think?

ksooo:
To sum up: Technically, the original implementation is correct, current implementation is not, because of the misuse of kodi end time for latest start time. The latter is not transparent for the user, because he expects kodi end time to denote the end of the show, not latest start time.

According to use case two and three from PhobosK setting min/max duration seems to cause more trouble than to do any good, but not setting it does not harm. To be able to define the start window very flexible (more the hard coded 0.5 hours around kodi start time like it was with the original approach) seems to be a very good thing, though.

However, none of the two approaches is perfect, each has it's drawbacks. See the problem Glenn-1990 describes above with the current approach.

Interestingly, although not "correct", the current approach seems to cover more use cases than the technically "correct" original implementation.

That's why I'm uncertain which way to go and currently tend to keep things as they are right now.

PhobosK:
Hi again,
Glen-1990, ksooo,
Here is the logic I see:
TVHeadend's autorec is used for EPG events recording isn't it (All other cases go to *Time Schedules* entries don't they)?
- So the event to be recorded depends on the EPG's start and end data for that event. In TVHeadend rules about auto-recording, what is used for time calculations is only *start*/*startWindow* + *minDuration*/*maxDuration* + *startExtra*/*stopExtra* (https://github.com/tvheadend/tvheadend/blob/master/src/htsp_server.c#L1701). The *approxTime* is deprecated already as we see...
- Since Kodi tries to simplify the process of creating rules for the end user, we cannot use all those fields for time calculations. But using TVHeadend as a backend doing all the time calculations we can reliably depend on using only *start*/*startWindow* + *startExtra*/*stopExtra* mapped to the appropriate Kodi's *start time*/*end time*/*start padding time*/*stop padding time*. All these do not have to be "transparent" for the end user... they just need to work for him/her the way he wants/needs... And they do that now with the current implementation (I've already given some examples in that other bug report).


So I still really cannot see your logic in doing the calculations (about autorec) in the pvr instead of using the backend.... 
And I still don't see any problem with the current implementation (using *start* and *startWindow*). What we do right now in the pvr is leaving all those calculations to the TVHeadend backend.
It's true that *startWindow* doesn't exactly map to Kodi's *end time* pvr option, but TVHeadend makes the calculations in such a way that it actually uses it as an *end time* (of course depending on the EPG data). And I disagree that the user is mislead... he/she just doesn't know how this is technically mapped to what he chooses as *end time*....

Referring to https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr_autorec.c#L169:
Yes it uses only the  *'te' (= event START time)*, but *&e* is actually the tested EPG event and *start* and *startWindow* are taken into consideration when deciding if the EPG event matches the user conditions to be auto-recorded - [line 158](https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr_autorec.c#L158) and [line 166](https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr_autorec.c#L166)  using *tad* (the user desired duration calculated, taking into account the *start* and *startWindow*)

As a side note, ever since this last implementation, my family members (who are totally non-tech people) are using this a lot for auto-recording and it never failed them, nor they find it hard to understand what to do... This is not a real statistics, but still is an example of an ordinary user experience...

So IMHO the now existing implementation of Autorec (using *start*/*startWindow*) is OK and should stay.... 

But you are more experienced in Kodi/TVHeadend coding, so it is your choice after all...

ksooo:
@PhobosK thanks. As I said, I'm with you and also think the implementation should stay as it is now.

Glenn-1990:
@ksooo, I still think your first approach was the correct one, but I think it contained a bug that caused kodi tho show another time (start and end) when the timer was transferred back to kodi..
That's what @PhobosK reported initially.

I believe that the correct implementation is this:
https://github.com/Glenn-1990/pvr.hts/commit/d71e0dd66e4d827b5b14846b8c8fa4490c37df66
@PhobosK can you test this and point out your issues?

How could we explain users that "start before" in the webui is the same as end time in kodi?

ksooo:
@PhobosK many good arguments pro @Glenn-1990's approach here: https://github.com/Glenn-1990/pvr.hts/commit/d71e0dd66e4d827b5b14846b8c8fa4490c37df66 I tend to follow these arguments, but want to get your opinion after you've tested yourself, first.

PhobosK:
https://github.com/Glenn-1990/pvr.hts/commit/d71e0dd66e4d827b5b14846b8c8fa4490c37df66#commitcomment-13083078

PR #87, opened 2015-08-06
Fix timer Start/End clocks showing incorrect time after 'save'

metaron-uk:
This fix works only if the tvheadend server and kodi are in the same timezone.
Servers in different timezones must therefore be considered unsupported.

The proper fix for this is https://github.com/kodi-pvr/pvr.hts/pull/85 which uses gmt offset from the server, however currently on several linux platforms which use vanilla glibc (debian/openElec + others) tvheadend fails to include any DST offset in the value returned.


ksooo:
Despite from the bump issues, looking good. :-)

Already runtime-tested by me, @metaron-uk and @PhobosK

ksooo:
Merge conflicts. Could you please rebase?

metaron-uk:
Done.

PR #89, opened 2015-08-10
Fix: Use epg data only for creation of epg-based timers.

ksooo:
This is actually an ancient bug that can easily fixed be now that we have timer types. More details here: https://github.com/xbmc/xbmc/pull/7782

@Jalle19 Any objections?


Jalle19:
Go ahead!

PR #90, opened 2015-08-11
Improve connection handling on suspend/wakeup.

ksooo:
This is an old patch I have on my private build for about a year now.

Today, on system suspend, pvr.hts looses connection to tvh (socket closed) and immediately retries to reconnect, even though suspend is in progress and the attempt cannot succeed. This results in error log messages while suspending and a non-deterministic behaviour after resume. This PR fixes this problem by preventing reconnects as long as pvr.hts is "suspended".

@Jalle19 objections?

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -93,7 +93,7 @@ void *CHTSPRegister::Process ( void )
 CHTSPConnection::CHTSPConnection ()
   : m_socket(NULL), m_regThread(this), m_ready(false), m_seq(0),
     m_serverName(""), m_serverVersion(""), m_htspVersion(0),
-    m_webRoot(""), m_challenge(NULL), m_challengeLen(0)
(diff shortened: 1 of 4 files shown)


src/HTSPConnection.cpp
```diff
@@ -495,10 +515,9 @@ void CHTSPConnection::Register ( void )
     return;
   }
 
-  /* Rate limit retry */
 fail:
-  Sleep(5000);
```
Jalle19: Is this removal intentional?

Jalle19:
Looks good in general, any improvement to wake/suspend handling is much appreciated.

src/HTSPConnection.cpp
ksooo: Yes.

ksooo:
Okay then. This on was (most probably ) my last contribution for the next 2 weeks. Vacation time. ;-)

Jalle19:
Have a nice vacation, you've definitely earned it!

PR #94, opened 2015-09-01
Force cmake to configure addon.xml

oliv3r:
The current source code only includes addon.xml.in whereas build_addon
etc need a valid .xml file. To resolve this, addon.xml.in needs to be
copied using CMake's configure_file.

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ if(WIN32)
   list(APPEND DEPLIBS ws2_32)
 endif()
 
+configure_file(pvr.hts/addon.xml.in ${CMAKE_CURRENT_SOURCE_DIR}/pvr.hts/addon.xml)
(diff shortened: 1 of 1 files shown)


Jalle19:
@Montellese

Montellese:
What version of `build_addon()` are you using? The latest version available in xbmc's master takes care of generating `addon.xml` from `addon.xml.in`. But obviously this will not work with older versions of `build_addon()`.

Jalle19:
Ping.

oliv3r:
I switched to using the Isenguard branch for 15.0 and that works.

I'm installing it on gentoo where there is a generic 15.0 kodi package, but no pvr.hts package, so came here. I'll re-open this issue if it applies with later versions of xbmc.

PR #95, opened 2015-09-11
workaround: playback of active recordings

Glenn-1990:
People are suffering from a bug that causes recordings to stutter:
http://forum.kodi.tv/showthread.php?tid=234640

This bug (in kodi) is triggered by the epg event id, which is only used by tvh and not by other addons at the moment. 
A fix is in the make for v16: https://github.com/xbmc/xbmc/pull/7905

So I suggest to just disable the event id for Isengard..

@Jalle19 @ksooo

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -396,7 +396,8 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
       }
 
       /* EPG event id */
-      rec.iEpgEventId = rit->second.eventId;
+      /* workaround: causes high cpu usage */
+      //rec.iEpgEventId = rit->second.eventId;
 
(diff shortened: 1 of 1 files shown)


ksooo:
As I've written here https://github.com/xbmc/xbmc/pull/7905 I still vote for a fix in master and in Isengard.

Thus, my -1 for this for now.

Jalle19:
I don't really have an opinion here except I'm not against any fix. It seems fixing it properly on the Kodi side will require more work than previously thought.

ksooo:
@Jalle19 This PR can be closed. See https://github.com/xbmc/xbmc/pull/8022.

Glenn-1990:
@ksooo thanks for jumping in the xbmc thread ;-)

PR #100, opened 2015-09-19
Remove circular buffer from the VFS

### LINKED ISSUE #92: Forward/backward skips often go to random places or start of recording

Jalle19:
As mentioned by @adamsutton in #99, this drastically simplifies the code and indeed seems to fix #92. @ksooo @adamsutton for review (especially the first commit which is kinda unrelated), @FernetMenta for sanity check.

@ksooo we'll have to backport this to Isengard even though it's quite a drastic "fix".


ksooo:
I have no objections re Isengard backport. This bug seems PITA for many people.

Despite from the multithreading question a raised, the change looks okay to me.

Jalle19:
I removed the mutex because AFAIK the whole class is only accessed from one thread. `m_fileId` was never mutex protected anyway (except as a side effect where locking was used anyway).

ksooo:
Ah, okay then.

Jalle19:
@ksoop do you mind runtime testing this? @Glenn-1990 you too perhaps?

ksooo:
I never was able to reproduce this bug. Thus I could only test whether things still do work for me.

adamsutton:
@Jalle19 @ksooo could you hold off merging this till I've had a chance to read it? Possibly tomorrow, Monday latest. I've had a very quick scan and I think a few changes are probably in order, but I need to read it through more carefully.

src/HTSPVFS.cpp
```diff
@@ -130,70 +88,40 @@ void CHTSPVFS::Close ( void )
...
-{
-  CLockObject lock(m_mutex);
-  m_buffer.reset();
-  m_bHasData = false;
-  m_bSeekDone = true;
-  m_currentReadLength = INITAL_READ_LENGTH;
-  m_seekCondition.Signal();
 }
 
 int CHTSPVFS::Read ( unsigned char *buf, unsigned int len )
```
adamsutton: Did I really write a function using int's like this! urgh! really this should be something sensible like size_t

I was obviously having a bad day, or just putting up with how Kodi dev's had done things.

src/HTSPVFS.cpp
```diff
@@ -130,70 +88,40 @@ void CHTSPVFS::Close ( void )
...
-  }
-
-  /* Wait for data */
-  m_condition.Wait(m_mutex, m_bHasData, 5000);
-
   /* Read */
-  ret = m_buffer.read(buf, len);
-  m_offset += ret;
-  return (int)ret;
+  int read = SendFileRead(buf, len);
```
adamsutton: Should be checking for return value here. I assume we're relying on caller to do the right thing.

But at the very least we shouldn't be doing m_offset += -1; in error case.

src/HTSPVFS.cpp
```diff
@@ -232,10 +165,15 @@ bool CHTSPVFS::SendFileOpen ( bool force )
...
+  {
+    CLockObject lock(m_conn.Mutex());
+
+    if (force)
+      m = m_conn.SendAndWait0("fileOpen", m);
+    else
+      m = m_conn.SendAndWait("fileOpen", m);
+  }
+
   if (m == NULL)
```
adamsutton: Erm, this looks like a bad idea, -1 not false.

src/HTSPVFS.cpp
```diff
@@ -286,10 +228,15 @@ long long CHTSPVFS::SendFileSeek ( int64_t pos, int whence, bool force )
...
+    CLockObject lock(m_conn.Mutex());
+
+    if (force)
+      m = m_conn.SendAndWait0("fileSeek", m);
+    else
+      m = m_conn.SendAndWait("fileSeek", m);
+  }
+
   if (m == NULL)
     return false;
```
adamsutton: And here. Wrong return type.

src/HTSPVFS.cpp
```diff
@@ -345,30 +279,16 @@ bool CHTSPVFS::SendFileRead()
...
-    return false;
+    return -1;
 
   /* Process */
-  if (htsmsg_get_bin(m, "data", &buf, &len))
+  if (htsmsg_get_bin(m, "data", buffer, &read))
   {
     htsmsg_destroy(m);
     tvherror("malformed fileRead response: 'data' missing");
-    return false;
```
adamsutton: You need an error return here!

src/HTSPVFS.cpp
```diff
@@ -345,30 +279,16 @@ bool CHTSPVFS::SendFileRead()
...
-  }
-
-  /* Gradually increase read length */
-  CLockObject lock(m_mutex);
-
-  if (m_currentReadLength * 2 < MAX_READ_LENGTH)
-    m_currentReadLength *= 2;
-
-  htsmsg_destroy(m);
```
adamsutton: That's gonna leak rather badly!

src/HTSPVFS.cpp
adamsutton: There is a mistake in a comment below this line (nto changed in the PR) it says "offset" is optional, yet requires it (which I think is correct).

adamsutton:
@Jalle19 @ksooo 

Couple of mistakes in my comments, I can't remember the whole locking philosophy, but I think the lock removals are OK, I think these were due to addition of extra threads.

I've made some suggested (but untested) changes, https://github.com/adamsutton/pvr.hts/commit/ae43216a46f7d78038ccfe410ca23f87152f7bb2

beralt:
I would be happy to assist with some testing if needed.

src/HTSPVFS.cpp
Jalle19: @ksooo maybe you can comment on the comment?

Jalle19:
@adamsutton I added your patch and runtime tested it, seems to be working.

src/HTSPVFS.cpp
```diff
@@ -345,30 +287,26 @@ bool CHTSPVFS::SendFileRead()
...
-  CLockObject lock(m_mutex);
-
-  if (m_currentReadLength * 2 < MAX_READ_LENGTH)
-    m_currentReadLength *= 2;
-
+  /* Cleanup */
   htsmsg_destroy(m);
-  return true;
+
+  return (int)read;
```
ksooo: Could you please use a c++ style cast here.

Jalle19: To make @adamsutton happy maybe we should return `size_t` all the way up the chain and perform the cast to `int` in `client.cpp` (since that's where the `int` requirement comes from)?

Jalle19:
@ksooo @adamsutton better?

ksooo:
@Jalle19 :+1:

ksooo:
Runtime-test was not a success. Crashes immediately when trying to play a recording.

Compile time warning:
<pre>
..../pvr.hts-master/src/HTSPVFS.cpp: In member function 'CHTSPVFS::SendFileRead(unsigned char*, unsigned int)':
....pvr.hts-master/src/HTSPVFS.cpp:296:74: warning: 'buffer' may be used uninitialized in this function [-Wmaybe-uninitialized]
   if (htsmsg_get_bin(m, "data", buffer, reinterpret_cast<size_t *>(&read)))
</pre>

=> https://github.com/kodi-pvr/pvr.hts/pull/100/files#diff-8c04d63d169ca5038cb71164664a467bR272

Crash:

<pre>
0  0x00007fe74ff489dd in htsmsg_get_bin (msg=<optimized out>, name=<optimized out>, binp=0x0, lenp=0x7fe73a7f3a50) at /home/kai/src/openelec/OpenELEC.tv/build
.OpenELEC-Generic.x86_64-5.95.5/pvr.hts-master/lib/libhts/htsmsg.c:381
#1  0x00007fe74ff418ae in CHTSPVFS::SendFileRead(unsigned char*, unsigned int) (this=this@entry=0x7fe744002f40, buf=0x7fe704233100 "pnp-org:device-1-0\">\n    
<specVersion>\n        <major>1</major>\n        <minor>0</minor>\330\060#\004\347\177", len=<optimized out>) at /home/kai/src/openelec/OpenELEC.tv/build.OpenE
LEC-Generic.x86_64-5.95.5/pvr.hts-master/src/HTSPVFS.cpp:296
#2  0x00007fe74ff4191f in CHTSPVFS::Read(unsigned char*, unsigned int) (this=0x7fe744002f40, buf=<optimized out>, len=<optimized out>) at /home/kai/src/openele
c/OpenELEC.tv/build.OpenELEC-Generic.x86_64-5.95.5/pvr.hts-master/src/HTSPVFS.cpp:100
#3  0x00007fe74ff40aa0 in CTvheadend::VfsRead(unsigned char*, unsigned int) () at /home/kai/src/openelec/OpenELEC.tv/build.OpenELEC-Generic.x86_64-5.95.5/pvr.h
ts-master/src/Tvheadend.h:581
#4  ReadRecordedStream (pBuffer=<optimized out>, iBufferSize=<optimized out>) at /home/kai/src/openelec/OpenELEC.tv/build.OpenELEC-Generic.x86_64-5.95.5/pvr.ht
s-master/src/client.cpp:572
#5  0x0000000000ca988e in PVR::CPVRClient::ReadStream(void*, long) ()
...
</pre>

bas-t:
I cherry-picked this into Isengard
Runtime test in Isengard: success
(Debian Jessie)

bas-t:
However, when trying to play a recording (directly, not in a nfs share), kodi crashes.
This may or may not be the result of just cherry-picking and not properly backporting.

See: http://pastebin.com/8Muu4j5h

puithove:
Same crash here on Jarvis - LiveTV fine, immediate crash trying to play a recording.  I'm a total git noob though.  cherry-picked so I could build against 4.0.0 API.

Jalle19:
I probably screwed something up with the reinterpret cast, will look into it later.

Jalle19:
@ksooo mind testing again?

bas-t:
@Jalle19 Your latest commit fixes the crash, but the recording will still not play.

ksooo:
@Jalle19 the reinterpret_cast is okay; not related to the crash.

This should fix it:

<pre>
diff --git a/src/HTSPVFS.cpp b/src/HTSPVFS.cpp
index 3067bc1..10961f2 100644
--- a/src/HTSPVFS.cpp
+++ b/src/HTSPVFS.cpp
@@ -269,7 +269,7 @@ long long CHTSPVFS::SendFileSeek ( int64_t pos, int whence, bool force )
 ssize_t CHTSPVFS::SendFileRead(unsigned char *buf, unsigned int len)
 {
   htsmsg_t   *m;
-  const void **buffer;
+  const void *buffer;
   ssize_t read;
 
   /* Build */
@@ -293,7 +293,7 @@ ssize_t CHTSPVFS::SendFileRead(unsigned char *buf, unsigned int len)
   }
 
   /* Get Data */
-  if (htsmsg_get_bin(m, "data", buffer, reinterpret_cast<size_t *>(&read)))
+  if (htsmsg_get_bin(m, "data", &buffer, reinterpret_cast<size_t *>(&read)))
   {
     tvherror("malformed fileRead response: 'data' missing");
     read = -1;
@@ -302,7 +302,7 @@ ssize_t CHTSPVFS::SendFileRead(unsigned char *buf, unsigned int len)
   }
   else
   {
-    memcpy(buf, *buffer, read);
+    memcpy(buf, buffer, read);
   }
 
   /* Cleanup */
</pre>

bas-t:
"The crash" is not right. I was obviously suffering from another crash.
So again: @Jalle19 Your latest commit fixes the crash that I'm experiencing (fullproof reproducable), but the recording will still not play. This is not related to the crash that @ksooo reported.

ksooo:
@Jalle19 For me, things work just fine without your latest commit and my patch for the data type of "buffer".

@bas-t cannot see how latest commit from Jalle19 could fix your crash, which is according to the log you posted, exactly the crash I encountered, btw. Could you maybe test without Jalle19's latest commit and with my patch included instead?

bas-t:
@ksooo will do.

bas-t:
@ksooo it won't compile. Am I missing something ? I get this:

/data/kodi-addons/pvr.hts/src/HTSPVFS.cpp: In member function ‘ssize_t CHTSPVFS::SendFileRead(unsigned char*, unsigned int)’:
/data/kodi-addons/pvr.hts/src/HTSPVFS.cpp:296:58: error: expected ‘<’ before ‘(’ token
   if (htsmsg_get_bin(m, "data", &buffer, reinterpret_cast(&read)))
                                                          ^
/data/kodi-addons/pvr.hts/src/HTSPVFS.cpp:296:58: error: expected type-specifier before ‘(’ token
/data/kodi-addons/pvr.hts/src/HTSPVFS.cpp:296:58: error: expected ‘>’ before ‘(’ token
CMakeFiles/pvr.hts.dir/build.make:169: recept voor doel 'CMakeFiles/pvr.hts.dir/src/HTSPVFS.cpp.o' is mislukt
make[2]: *** [CMakeFiles/pvr.hts.dir/src/HTSPVFS.cpp.o] Fout 1

ksooo:
@bas-t you f****d up the patch file, I guess. Try this one: http://paste.ubuntu.com/12504317/

ksooo:
@bas-t in fact it was me, who killed the patch file while pasting it into gh comment.... Always forget about problems with \> and \< and html. haha.

bas-t:
@ksoo no, you did. look at the difference!
Anyhow, with the corrected patch, it compiles just fine.
Going to test it now

Op 20-09-15 om 19:47 schreef Kai Sommerfeld:

bas-t:
I missed your post, and reacted too soon.

Op 20-09-15 om 19:54 schreef Kai Sommerfeld:

ksooo:
all fine

bas-t:
Runtime test without latest @Jalle19 patch and with @ksooo patch (http://paste.ubuntu.com/12504317/).

SUCCESS! Way to go, both of you (and @adamsutton).

Only remaining thing on my system seems to be that in case I jump to the beginning, or mayby past that, given the size of the jump, I sometimes don't get audio. Skipping a bit forward again restores audio.

Mind you, I'm still testing in Isengard only.
Thank you all so much for this breakthrough, this is exactly what kept me from using Kodi at all.

ksooo:
@bas-t cool. Thanks for your support.

As you've tested on Isengard and I on Jarvis, we already know that the changes of this PR will work on both versions. Cool.

@Jalle19 mind adding the fix to the addon changelog?

@adamsutton thank you for jumping by and giving the advices needed to find the fix. ESpecially like the reduction in complexity (less, sometimes is more).

Jalle19:
Thanks for testing. Should we keep it the way it is now or should I change it back to using reinterpret_cast? IMO a static cast on the return value is "safer" (reinterpret_cast is considered the most dangerous type of cast).

ksooo:
As the reinterpret_cast is semantically correct here and we tested with that cast in place, I'd say let's stick with it.

Jalle19:
@ksooo done

Jalle19:
@ksooo added changelog update too, should be good to go now.

ksooo:
@Jalle mind squasing commit 3 and 4 into commit 2 (or at least 4 into 2)?

Jalle19:
@ksooo done (took the easiest route)

ksooo:
Thank you, Sam.

Jalle19:
As if by magic this also fixes the ancient bug where the picture is basically frozen if you attempt to fast-forward or rewind in recordings. Now it's just as smooth as for local content.

adamsutton:
@Jalle19 that's great news, and shows why I shouldn't be allowed to code this stuff!

PR #103, opened 2015-09-20
Isengard backports

Jalle19:
@ksooo I picked the ones that weren't related to any of the API changes. I'll add #100 to the mix once we've merged it into master.


ksooo:
It does not make much sense to inlcude https://github.com/Jalle19/pvr.hts/commit/02c6644c138f8d171b2f71ad8d7b109c835755be as GetBufferTimeStart and GetBufferTimeEnd never get called in Isengard.

Jalle19:
@ksooo that's too bad, that commit was the most PITA to cherry-pick ;-) So should I drop it?

ksooo:
Yes, makes really no sense for Isengard, sorry. Please drop it.

Jalle19:
@ksooo dropped the timeshift commit and added the VFS fix. Haven't build tested any of this yet so this is not yet for merging.

Jalle19:
@ksooo I added the latest fix, I think we're good to go now. I'll run some test builds on Jenkins before PRing the bump.

ksooo:
Your button.

Jalle19:
Tested in production, everything seems to work.

PR #111, opened 2015-10-09
Ignore the auto-generated addon.xml file

Jalle19:


Jalle19:
@Montellese this is correct right?

Montellese:
Yes if you do an in-tree build.

Jalle19:
@Montellese what other types of builds are there?

Montellese:
AFAIK in-tree builds only work on linux for linux. For anything else we do out-of-tree builds where the generated files are in a different directory than the checked out addon files/code.

Jalle19:
Okay, but either way this file should never be checked in to git right?

Montellese:
Yes.

PR #113, opened 2015-10-14
Refactor Settings into a singleton

Jalle19:
@ksooo


ksooo:
Nice. Nothing to complain about....

PR #116, opened 2015-10-15
Refactor class Settings, step 2

ksooo:
Fixes Coverity complaints after https://github.com/kodi-pvr/pvr.hts/commit/f013946b78e5b4770316883edfdf8028140d7d0a @Jalle19 mind taking a look.

BTW: I overlooked the public members in class Settings you've just introduced. I do not like that, tbh and would have bugged you with that if I had seen this before the merge. ;-)

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -112,13 +112,13 @@ std::string CHTSPConnection::GetWebURL ( const char *fmt, ... )
   const Settings &settings = Settings::GetInstance();
 
   // Generate the authentication string (user:pass@)
-  std::string auth = settings.strUsername;
-  if (!(auth.empty() || settings.strPassword.empty()))
-    auth += ":" + settings.strPassword;
+  std::string auth = settings.GetUsername();
+  if (!(auth.empty() || settings.GetPassword().empty()))
+    auth += ":" + settings.GetPassword();
   if (!auth.empty())
     auth += "@";
 
-  std::string url = StringUtils::Format("http://%s%s:%d", auth.c_str(), settings.strHostname.c_str(), settings.iPortHTTP);
+  std::string url = StringUtils::Format("http://%s%s:%d", auth.c_str(), settings.GetHostname().c_str(), settings.GetPortHTTP());
 
   CLockObject lock(m_mutex);
   va_start(va, fmt);
@@ -133,7 +133,7 @@ bool CHTSPConnection::WaitForConnection ( void )
 {
   if (!m_ready) {
     tvhtrace("waiting for registration...");
-    m_regCond.Wait(m_mutex, m_ready, Settings::GetInstance().iConnectTimeout);
+    m_regCond.Wait(m_mutex, m_ready, Settings::GetInstance().GetConnectTimeout());
   }
   return m_ready;
 }
@@ -155,7 +155,7 @@ std::string CHTSPConnection::GetServerString ( void )
   const Settings &settings = Settings::GetInstance();
 
   CLockObject lock(m_mutex);
-  return StringUtils::Format("%s:%d [%s]", settings.strHostname.c_str(), settings.iPortHTSP,
+  return StringUtils::Format("%s:%d [%s]", settings.GetHostname().c_str(), settings.GetPortHTSP(),
              m_ready ? "connected" : "disconnected");
 }
 
@@ -228,7 +228,7 @@ bool CHTSPConnection::ReadMessage ( void )
   cnt = 0;
   while (cnt < len)
   {
-    r = m_socket->Read((char*)buf + cnt, len - cnt, Settings::GetInstance().iResponseTimeout);
+    r = m_socket->Read((char*)buf + cnt, len - cnt, Settings::GetInstance().GetResponseTimeout());
     if (r < 0)
     {
       tvherror("failed to read packet (%s)",
@@ -325,7 +325,7 @@ bool CHTSPConnection::SendMessage0 ( const char *method, htsmsg_t *msg )
 htsmsg_t *CHTSPConnection::SendAndWait0 ( const char *method, htsmsg_t *msg, int iResponseTimeout )
 {
   if (iResponseTimeout == -1)
-    iResponseTimeout = Settings::GetInstance().iResponseTimeout;
+    iResponseTimeout = Settings::GetInstance().GetResponseTimeout();
   
   uint32_t seq;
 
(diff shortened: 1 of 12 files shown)


Jalle19:
Yeah leaving them public was a quick fix, we should change them to use getters/setters like the other new classes.

src/tvheadend/Settings.h
```diff
@@ -55,7 +55,10 @@ namespace tvheadend {
     int         iAutorecMaxDiff;
 
   private:
-    Settings() { }
+    Settings()
+    : iPortHTSP(9981), iPortHTTP(9982), iConnectTimeout(10), iResponseTimeout(5),
```
Jalle19: There are #defines for these somewhere I think?

ksooo: Good point.

ksooo:
@Jalle19 now with proper defaults, encapsulated members... Could you please review?

src/HTSPConnection.cpp
```diff
@@ -115,13 +115,13 @@ std::string CHTSPConnection::GetWebURL ( const char *fmt, ... )
   const Settings &settings = Settings::GetInstance();
 
   // Generate the authentication string (user:pass@)
-  std::string auth = settings.strUsername;
-  if (!(auth.empty() || settings.strPassword.empty()))
-    auth += ":" + settings.strPassword;
+  std::string auth = settings.GetUsername();
+  if (!(auth.empty() || settings.GetPassword().empty()))
+    auth += ":" + settings.GetPassword();
```
Jalle19: Now that `Settings` is an actual class with actual methods I was thinking we could factor away some things like this to separate methods, e.g. `GetAuthenticationString()` in this case. This may be an edge case since it's technically not that related to "settings" so it's up to you if you want to do it.

src/client.cpp
```diff
@@ -142,33 +141,26 @@ ADDON_STATUS ADDON_Create(void* hdl, void* _unused(props))
...
-    settings.iPreTuneCloseDelay = g_iPreTunerCloseDelay;
-  }
-  else
-  {
-    /* When we don't want to use predictive tuning */
-    settings.iTotalTuners = 1;
-    settings.iPreTuneCloseDelay = 0;
-  }
+  settings.SetPreTunerEnabled(g_bPreTunerEnabled);
+  settings.SetTotalTuners(g_bPreTunerEnabled ? g_iTotalTuners : 1);
```
Jalle19: Similar thing here, IMO `SetTotalTuners` could handle this check internally. It's one of the benefits of setters compared to plain assignment.

src/tvheadend/Settings.h
```diff
@@ -31,6 +31,22 @@ namespace tvheadend {
   class Settings {
   public:
 
+    // Default values.
+    static const std::string DEFAULT_HOST;
+    static const int         DEFAULT_HTTP_PORT           = 9981;
```
Jalle19: IMO it would make sense to define all these values in the newly introduced `.cpp` file, currently if you're looking at just the `.cpp` file you may come under the impression that those three constants are the only ones available in the class. Up to you though.

src/tvheadend/entity/Entity.h
```diff
@@ -34,7 +34,7 @@ namespace tvheadend
     class Entity
     {
     public:
-      Entity() : m_dirty(false), m_id(0) {};
+      Entity() : m_id(0), m_dirty(false) {};
```
Jalle19: If you find yourself wondering what to do on a Friday night you could perhaps move this change to a separate commit ;-)

Jalle19:
Apart from the comments, very nice!

src/tvheadend/entity/Entity.h
ksooo: This is already a separate commit. ;-)

Jalle19: WTF you're totally right

src/tvheadend/Settings.h
ksooo: Makes sense.

ksooo:
@Jalle19 for what do we need these g_blah settings vars? Could we not just get rid of those?

Jalle19:
@ksooo nothing really, we really should get rid of them and the `UPDATE_STR`/`UPDATE_INT` macros at the same time and just handle "did a setting just change" in the new `Settings` class instead.

ksooo:
yep. will do that.

Jalle19:
Thanks, just tack it onto this PR.

ksooo:
@Jalle19 consider this ready now. Mind doing final review?

src/tvheadend/Settings.cpp
```diff
@@ -0,0 +1,187 @@
...
+  /* Data Transfer */
+  else if (key == "epg_async")
+    return SetBoolSetting(GetTraceDebug(), value);
+  /* Predictive Tuning */
+  else if (key == "pretuner_enabled")
+  {
+    if (GetTotalTuners() > 1 && *(reinterpret_cast<const bool*>(value)))
+      return ADDON_STATUS_OK;
+    else
+     return ADDON_STATUS_NEED_RESTART;
```
Jalle19: Missing one space

src/tvheadend/Settings.cpp
```diff
@@ -0,0 +1,187 @@
...
+  /* Auto recordings */
+  SetAutorecApproxTime(ReadIntSetting("autorec_approxtime", DEFAULT_APPROX_TIME));
+  SetAutorecMaxDiff(ReadIntSetting("autorec_maxdiff", DEFAULT_AUTOREC_MAXDIFF));
+}
+
+ADDON_STATUS Settings::SetStringSetting(const std::string &oldValue, const void *newValue)
+{
+  if (oldValue == std::string(reinterpret_cast<const char *>(newValue)))
+  {
+    // value not changed.
```
Jalle19: Minor thing, but if you'd add these comments in the method documentation in the header file then this method and the two below it could be shortened to 4 lines each instead of the current 10.

Jalle19:
Very nice, it's good to finally get rid of the globals and those pesky macros (even though it all worked just fine).

Jalle19:
@ksooo push the button when you're ready, the last comments aren't that important since they're just cosmetics.

ksooo:
Next step: Rip logging stuff out of Tvheadend.h. Currently, Settings.cpp needs to include Tvheadend.h only because of using tvherror.

Jalle19:
@ksooo I actually started work on that last night. I would like to have e.g. `Logger::LogInfo("blabla %s", "foo")`. Ideally I'd like to inject the `XBMC->Log(...)` part through a callback so the logger doesn't explicitly have to include `client.h` to get access to that global (because you can't have multiple `libXBMC_addonHelper` objects in your code.

ksooo:
This is actually literally what I had in mind. So, your turn?

Jalle19:
I'll try ;-)

ksooo:
I trust in you. ;-)

Issue #115, opened 2015-10-14
Predictive tuning broken when channel numbers are not consecutive


Jalle19:
Just realized this while I was looking at making the code more understandable and possibly even improving it. Currently the predictive tuning is triggered if `toNum` is the same as `fromNum + 1`, which may not be true for two logically consecutive channels.

ksooo:
sigh

Issue #91, opened 2015-08-18
Don't show "no free adapter" when pre-tuning fails


Jalle19:
Consider this scenario:

1. User sets "use pretuners" to 2.
2. User starts recording something (basically hogging one tuner for some time).
3. User starts zapping through channels

What happens:

The "no free adapter" message is shown all the time. This shouldn't happen.

Glenn-1990:
This should be fixed with: 
https://github.com/kodi-pvr/pvr.hts/pull/82

Jalle19:
Yes.

Jalle19:
I have to check if this message came from the pre-tuning subscription or from a recording that was about to start. The particular show has its priority set to "Unimportant" in tvheadend so that could be the issue. Pre/post-tuning priority should be lower than the lowest DVR priority, is that the case now?

Glenn-1990:
Yes, it should be, pretuning uses weight 40 and that should be lower than unimportant.

PR #117, opened 2015-10-17
Predictive tuning improvements

Jalle19:
@ksooo this is still WIP (the current commit works though) but you might wanna take a look.

### DIFF
--- a/src/tvheadend/ChannelTuningPredictor.h
+++ b/src/tvheadend/ChannelTuningPredictor.h
@@ -1,0 +1,116 @@
+#pragma once
+
+/*
+ *      Copyright (C) 2005-2015 Team Kodi
+ *      http://www.xbmc.org
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, write to
+ *  the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ *  http://www.gnu.org/copyleft/gpl.html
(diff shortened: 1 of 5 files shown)


Jalle19:
Fixes https://github.com/kodi-pvr/pvr.hts/issues/115 by the way, remind me to add it to the commit message too.

Glenn-1990:
I also spotted another bug with predictive tuning...
https://github.com/kodi-pvr/pvr.hts/blob/master/src/Tvheadend.cpp#L2026
The weight of the new active subscription is set to normal before! the old subscription gets decreased in weight. So the 2 subscriptions will be set to normal at the same time for a very short period, this could lead to hiccups in other streams if no other tuners are available.

I can also make a PR for this?

Jalle19:
Nice catch, fixing that would probably fix #91. If you fix I can pick your commit into this PR, no need to make a separate pull.

Jalle19:
@Glenn-1990 don't bother with a fix, I'm working on this right now anyway so I'll add a commit for it. Thanks for bringing it to attention though!

Jalle19:
@ksooo no longer WIP now.

ksooo:
Stumbled only over the magic number used by the new channel predictor. Everything else looks good to me.

Jalle19:
@ksooo added `CHANNEL_ID_NONE` to replace the magic number.

Jalle19:
@ksooo good to go?

src/tvheadend/ChannelTuningPredictor.h
```diff
@@ -0,0 +1,116 @@
...
+{
+
+  namespace predictivetune
+  {
+
+    /**
+     * Used to indicate that predictive tuning failed to determine which
+     * channel to be tuned next
+     */
+    const unsigned int CHANNEL_ID_NONE = -1;
```
ksooo: unsigned ... = -1 ? Really?

Jalle19: Yes, it's the same as UINT32_MAX

ksooo: Ofc, just wanted to be sure that this was intended.

One last thing: why unsigned int and not uint32_t like used elsewhere with channel ids? https://github.com/kodi-pvr/pvr.hts/pull/117/files#diff-ce6ea06b04dac3a0c1f703fe9ed2cc37R65

Jalle19: Just a habit, I can change it before merging.

ksooo:
good to go.

Issue #40, opened 2015-06-12
Feature request: add support for profile selection


TheTroll:
Hi,

it would be very nice to allow to select the streaming profile.
TVheadend exports the list of profiles that can be used...

Thanks!

Jalle19:
I'm not really sure how to do it since the addon settings have to be static, aka. there's no communication with the backend before the addon is started. This means there's no way to retrieve the list of profiles at this point. It may be solvable using "client-specific settings", but I've never used that and I'm not even sure if it's possible to use those in a skin-independent matter.

Jalle19:
Furthermore, I don't really think the client should be the one deciding which streaming profile should be used. Is there really a use case where you'd want to change the profile more than once? If not then I don't see why you couldn't specify the profile to use from the web interface.

TheTroll:
Thanks for your answer!

My goal is to reduce the number of usernames for a single user.
I use kodi at home on my TV but also on my ipad.

On my TV I want the PASS profile and on my ipad a transcoding one...

It would be really great if I could avoid to duplicate usernames (i mean a user called "blah" and user called "transcoding-blah")

Thanks!

Jalle19:
You can use the same username but different passwords, or limit the TV's user to the TV's IP address. The access entries are matched from top to bottom.

TheTroll:
Indeed this is a possibility...
By the way I notice something very strange about that.
If I have a "user" with 2 lines.

- first line has  all access enable + the htsp profile selected
- second line has nothing enable + a transcoding profile selected

It is impossible to do a http streaming...
But if I invert the lines, it is ok and the PVR HTS client takes the HTSP profile

Jalle19:
This is starting to be off-topic (join #hts on irc.freenode.net if you need general support) but that's probably because the first entry is only valid for HTSP, so when you try streaming over HTTP is falls back to the second entry which has no streaming permissions. You need to check "Streaming" in order to stream.

TheTroll:
Yes sorry .. I ll check on irc
Thanks

Jalle19:
@ksooo as discussed on IRC, what's your opinion on this: add a string setting to the addon where the user can enter the name of the streaming profile to use. If the profile exists, use it, otherwise don't. That way we don't need any user interface hacks.

TheTroll:
I think it's a very good alternative!

ksooo:
[...]

PR #119, opened 2015-10-18
Add support for streaming profiles

Jalle19:
See #40 

@ksooo for review. I've split the feature into three commits for easier review, we can squash it before merging if you want to.

I'm thinking whether we should add another boolean option to disable predictive tuning if a streaming profile is set. Having multiple transcoded subscriptions can easily overwhelm a low-powered server.

Since this touches settings I'd like to get #116 merged before this one. #79 would be nice to get in before as well since it refactors `Subscription` into a class (if the feature is not ready we could opt for just merging that particular part).

@TheTroll can you test this?

### DIFF
--- a/pvr.hts/resources/settings.xml
+++ b/pvr.hts/resources/settings.xml
@@ -15,6 +15,9 @@
     <setting id="autorec_approxtime" type="enum" label="30051" lvalues="30052|30053" default="0"/>
     <setting id="autorec_maxdiff" enable="eq(-1,1)"  type="slider" option="int" range="0,5,120" label="30054" default="15" />
     
+    <setting label="30500" type="lsep"/>
+    <setting id="streaming_profile" type="text" label="30501" default="" />
+
     <setting label="30400" type="lsep"/>
     <setting id="pretuner_enabled" type="bool" label="30403" default="false" />
     <setting id="total_tuners" enable="eq(-1,true)" type="slider" option="int" range="2,1,10" label="30401"  default="2" />
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -165,6 +165,16 @@ bool CTvheadend::HasStreamingProfile(const std::string &streamingProfile) const
   ) != m_profiles.cend();
 }
 
+std::string CTvheadend::GetStreamingProfile() const
+{
+  std::string streamingProfile;
+
+  if (HasStreamingProfile(Settings::GetInstance().GetStreamingProfile()))
+    streamingProfile = Settings::GetInstance().GetStreamingProfile();
+
+  return streamingProfile;
+}
+
 /* **************************************************************************
  * Tags
  * *************************************************************************/
@@ -1416,6 +1426,12 @@ void CTvheadend::SyncCompleted ( void )
     XBMC->QueueNotification(
         QUEUE_ERROR,
         XBMC->GetLocalizedString(30502), streamingProfile.c_str());
+  }
+  else
+  {
+    /* Tell each demuxer to use this profile from now on */
+    for (auto *dmx : m_dmx)
+      dmx->SetStreamingProfile(streamingProfile);
   }
 }
 
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -153,6 +153,18 @@ void CTvheadend::QueryAvailableProfiles()
   htsmsg_destroy(m);
 }
 
+bool CTvheadend::HasStreamingProfile(const std::string &streamingProfile) const
+{
+  return std::find_if(
+      m_profiles.cbegin(),
+      m_profiles.cend(),
+      [&streamingProfile](const Profile &profile)
+      {
+        return profile.GetName() == streamingProfile;
+      }
+  ) != m_profiles.cend();
+}
+
 /* **************************************************************************
  * Tags
  * *************************************************************************/
@@ -1392,6 +1404,19 @@ void CTvheadend::SyncCompleted ( void )
   SyncDvrCompleted();
   SyncEpgCompleted();
   m_asyncState.SetState(ASYNC_DONE);
+
+  /* Query the server for available streaming profiles */
+  QueryAvailableProfiles();
+
+  /* Show a notification if the profile is not available */
+  std::string streamingProfile = Settings::GetInstance().GetStreamingProfile();
+
+  if (!streamingProfile.empty() && !HasStreamingProfile(streamingProfile))
+  {
+    XBMC->QueueNotification(
+        QUEUE_ERROR,
+        XBMC->GetLocalizedString(30502), streamingProfile.c_str());
+  }
 }
 
 void CTvheadend::SyncChannelsCompleted ( void )
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -106,6 +106,53 @@ std::string CTvheadend::GetImageURL ( const char *str )
   }
 }
 
+void CTvheadend::QueryAvailableProfiles()
+{
+  /* Build message */
+  htsmsg_t *m = htsmsg_create_map();
+
+  /* Send */
+  {
+    CLockObject lock(m_conn.Mutex());
+    m = m_conn.SendAndWait("getProfiles", m);
+  }
+
+  /* Validate */
+  if (m == nullptr)
+    return;
+
+  htsmsg_t *l;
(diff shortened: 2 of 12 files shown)


ksooo:
Yeah, looks good. Obviously the new setting heavily interferes with #116...

TheTroll:
Hi,

I ll take a look thx!

Jalle19:
@ksooo yeah I'll rebase once it's in.
@TheTroll thanks

Jalle19:
@TheTroll can you test?

TheTroll:
Hi, I will before the end of the week.
Sorry for the delay!

Jalle19:
No problem, I'll test again when I get home (just checking that the rebase went well).

@ksooo @Glenn-1990 we should refactor `SendSubscribe` so that there are two separate methods for "restart" and "normal subscribe". That way we don't have to make all the parameters optional and it improves readability a bit. Currently you have to use e.g. `GetProfile()` inside the method instead of just using the parameter directly which may be a bit confusing and prone to errors.

src/tvheadend/Subscription.h
```diff
@@ -71,14 +72,19 @@ namespace tvheadend
...
      * @param channelId the channel to subscribe to
      * @param weight the desired subscription weight
+     * @param profile the profile to use for the subscription
      * @param restart restart the current subscription (i.e. after lost connection), other parameters will be ignored
      */
-    void SendSubscribe(uint32_t channelId = 0, uint32_t weight = SUBSCRIPTION_WEIGHT_NORMAL, bool restart = false);
+    void SendSubscribe(uint32_t channelId = 0,
+                       uint32_t weight = SUBSCRIPTION_WEIGHT_NORMAL,
+                       const std::string &profile = "",
+                       bool restart = false);
```
Glenn-1990: Is it possible to change the profile on run-time?
Otherwise we could just pass the profile with the constructor, that will make all this profile passing unnecessary?

Jalle19:
@ksooo I had to move the logic that queries for available profiles to `CTvheadend::SyncCompleted()`. The initial sync is done on a separate thread that `client.cpp` which meant that when connecting to the server over WAN (which is slower) the `getProfiles` call timed out because tvheadend was busy pushing all the initial sync data over the wire. Now that I moved the call the issue disappears, plus we can keep the streaming profile related methods private.

Will do some final runtime testing when I get home, then I'll merge and update the changelog etc.

src/tvheadend/Subscription.h
Jalle19: Technically it should be but currently no since an addon restart is triggered if you change the setting. I'm perfectly fine with requiring a restart and feeding the desired profile through the constructor instead, anything to keep the parameter count on this method down is good in by book. @ksooo thoughts?

src/Tvheadend.cpp
```diff
@@ -1345,6 +1410,19 @@ void CTvheadend::SyncCompleted ( void )
...
+  QueryAvailableProfiles();
+
+  /* Show a notification if the profile is not available */
+  std::string streamingProfile = Settings::GetInstance().GetStreamingProfile();
+
+  if (!HasStreamingProfile(streamingProfile))
+  {
+    XBMC->QueueNotification(
+        QUEUE_ERROR,
+        XBMC->GetLocalizedString(30502), streamingProfile.c_str());
```
Glenn-1990: This error is shown when I didn't set any streaming profile?

Jalle19: Haha could be that I didn't test that scenario yet :-D

Glenn-1990:
It's probably a silly question, but why do wee need this, we can set the desired streaming profile in the tvheadend config right?

sorano:
I'm thinking that we can provide different profiles for end-users, and they can choose the profile that fits their current scenario?

For example, if the user is in a place with low bandwidth, he can select a profile with heavy transcoding to save bandwidth.

Jalle19:
@Glenn-1990 see https://github.com/kodi-pvr/pvr.hts/issues/40, there are a few reasonable use cases for this (especially the one about recordings)

Jalle19:
@Glenn-1990 you mind giving this a final spin (mainly to check that untouched configurations still work)? @ksooo I squashed the latest fixups, mind doing a final review before merging?

pvr.hts/resources/settings.xml
```diff
@@ -15,6 +15,9 @@
     <setting id="autorec_approxtime" type="enum" label="30051" lvalues="30052|30053" default="0"/>
     <setting id="autorec_maxdiff" enable="eq(-1,1)"  type="slider" option="int" range="0,5,120" label="30054" default="15" />
     
+    <setting label="30500" type="lsep"/>
+    <setting id="streaming_profile" type="text" label="30501" default="" />
+
```
ksooo: I have the feeling that our "Advanced settings" page becomes to big. Is it just me or should we do a settings pages redesign soon?

src/Tvheadend.cpp
```diff
@@ -106,6 +106,71 @@ std::string CTvheadend::GetImageURL ( const char *str )
...
+  /* Validate */
+  if (m == nullptr)
+    return;
+
+  htsmsg_t *l;
+  htsmsg_field_t *f;
+
+  if ((l = htsmsg_get_list(m, "profiles")) == NULL)
+  {
+    tvherror("malformed getProfiles: 'profiles' missing");
```
ksooo: <code>htsmsg_destroy(m);</code> is missing here.

src/Tvheadend.cpp
```diff
@@ -106,6 +106,71 @@ std::string CTvheadend::GetImageURL ( const char *str )
...
+      profile.SetUuid(str);
+    if ((str = htsmsg_get_str(&f->hmf_msg, "name")) != NULL)
+      profile.SetName(str);
+    if ((str = htsmsg_get_str(&f->hmf_msg, "comment")) != NULL)
+      profile.SetComment(str);
+
+    tvhdebug("profile name: %s, comment: %s added",
+             profile.GetName().c_str(), profile.GetComment().c_str());
+
+    m_profiles.push_back(profile);
```
ksooo: <code>htsmsg_destroy(m);</code> is missing here.

src/Tvheadend.cpp
```diff
@@ -106,6 +106,71 @@ std::string CTvheadend::GetImageURL ( const char *str )
...
+}
+
+bool CTvheadend::HasStreamingProfile(const std::string &streamingProfile) const
+{
+  return std::find_if(
+      m_profiles.cbegin(),
+      m_profiles.cend(),
+      [&streamingProfile](const Profile &profile)
+      {
+        return profile.GetName() == streamingProfile;
```
ksooo: Are streaming profile names case sensitive?

Jalle19: No idea but I guess so.

pvr.hts/resources/settings.xml
Jalle19: @Glenn-1990 already refactored it, but yeah, we may need an additional tab. I'd leave that to a separate PR though.

ksooo: Sure, nothing for this PR. Maybe you can file an issue for that so we don't forget.

Glenn-1990:
@Jalle19 working now with default config ;-)

Jalle19:
Good, I'll just have to plug the leaks then and we should be good to go.

TheTroll:
Hi,

I have moved to 15.2 and pvr.hts master, is it the right combination to test the new feature ?
Where can I select the profile ?

Thanks

Jalle19:
You can't use pvr.hts master with 15.2, if you want to test this you should grab a nightly build.

TheTroll:
So master is meant to be for kodi master ?

Jalle19:
Yes, exactly.

TheTroll:
pvr.hts master doen't compile anymore with kodi master ...
But it compiles and run with 15.2 branch :)

TheTroll:
Ok my bad, I was using the the isengard branch!
Let me test with the right branch

TheTroll:
Now it works thanks :)
Why not adding this settings in "client specific" instead of the addon's settings ?

Jalle19:
Because none of us have any idea how to use that and users are less likely to find it.

TheTroll:
Ok :)
Last question, I see that the profile is entered manually. Any way to retrieve it and let the user choose from a list ?

Jalle19:
Nope.

PR #121, opened 2015-10-18
Subscription rework

Glenn-1990:
@Jalle19

### DIFF
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -31,7 +31,8 @@ using namespace tvheadend;
 
 CHTSPDemuxer::CHTSPDemuxer ( CHTSPConnection &conn )
   : m_conn(conn), m_pktBuffer((size_t)-1),
-    m_seekTime(INVALID_SEEKTIME)
+    m_seekTime(INVALID_SEEKTIME),
+    m_subscription(conn)
 {
   m_lastUse = 0;
 }
@@ -43,11 +44,15 @@ CHTSPDemuxer::~CHTSPDemuxer ()
 void CHTSPDemuxer::Connected ( void )
 {
   /* Re-subscribe */
-  if (m_subscription.active)
+  if (m_subscription.IsActive())
   {
     tvhdebug("demux re-starting stream");
-    SendSubscribe(true);
-    SendSpeed(true);
+    m_subscription.SendSubscribe(true);
+    m_subscription.SendSpeed(true);
+
+    /* Reset status */
+    m_signalInfo.Clear();
+    m_sourceInfo.Clear();
   }
 }
 
@@ -58,8 +63,8 @@ void CHTSPDemuxer::Connected ( void )
 void CHTSPDemuxer::Close0 ( void )
 {
   /* Send unsubscribe */
-  if (m_subscription.active)
-    SendUnsubscribe();
+  if (m_subscription.IsActive())
+    m_subscription.SendUnsubscribe();
 
   /* Clear */
   Flush();
@@ -81,22 +86,21 @@ bool CHTSPDemuxer::Open ( uint32_t channelId, enum eSubscriptionWeight weight )
 
   /* Close current stream */
   Close0();
-  
-  /* Create new subscription */
-  m_subscription = SSubscription();
-  m_subscription.channelId = channelId;
-  m_subscription.weight = weight;
 
-  /* Open */
-  SendSubscribe();
+  /* Open new subscription */
+  m_subscription.SendSubscribe(channelId, weight);
   
+  /* Reset status */
+  m_signalInfo.Clear();
+  m_sourceInfo.Clear();
+
   /* Send unsubscribe if subscribing failed */
-  if (!m_subscription.active)
-    SendUnsubscribe();
+  if (!m_subscription.IsActive())
+    m_subscription.SendUnsubscribe();
   else
     m_lastUse = time(NULL);
   
-  return m_subscription.active;
+  return m_subscription.IsActive();
 }
 
 void CHTSPDemuxer::Close ( void )
@@ -150,26 +154,11 @@ void CHTSPDemuxer::Abort ( void )
 bool CHTSPDemuxer::Seek 
   ( int time, bool _unused(backwards), double *startpts )
 {
-  htsmsg_t *m;
-
-  CLockObject lock(m_conn.Mutex());
-  if (!m_subscription.active)
+  if (!m_subscription.IsActive())
     return false;
 
-  tvhdebug("demux seek %d", time);
-
-  /* Build message */
-  m = htsmsg_create_map();  
-  htsmsg_add_u32(m, "subscriptionId", m_subscription.subscriptionId);
-  htsmsg_add_s64(m, "time",           (int64_t)time * 1000);
-  htsmsg_add_u32(m, "absolute",       1);
-
-  /* Send and Wait */
-  m = m_conn.SendAndWait("subscriptionSeek", m);
-  if (!m)
+  if (!m_subscription.SendSeek(time))
     return false;
-  
-  htsmsg_destroy(m);
 
   /* Wait for time */
   if (!m_seekCond.Wait(m_conn.Mutex(), m_seekTime, Settings::GetInstance().GetResponseTimeout()))
@@ -191,18 +180,16 @@ bool CHTSPDemuxer::Seek
 void CHTSPDemuxer::Speed ( int speed )
 {
   CLockObject lock(m_conn.Mutex());
-  if (!m_subscription.active)
+  if (!m_subscription.IsActive())
     return;
-  m_subscription.speed = speed;
-  SendSpeed();
+  m_subscription.SendSpeed(speed);
 }
 
 void CHTSPDemuxer::Weight ( enum eSubscriptionWeight weight )
 {
-  if (!m_subscription.active || m_subscription.weight == weight)
+  if (!m_subscription.IsActive() || m_subscription.GetWeight() == static_cast<uint32_t>(weight))
     return;
-  m_subscription.weight = weight;
-  SendWeight();
+  m_subscription.SendWeight(static_cast<uint32_t>(weight));
 }
 
 PVR_ERROR CHTSPDemuxer::CurrentStreams ( PVR_STREAM_PROPERTIES *streams )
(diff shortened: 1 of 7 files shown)


src/tvheadend/Subscription.cpp
```diff
@@ -0,0 +1,305 @@
...
+  htsmsg_add_s32(m, "channelId",       GetChannelId());
+  htsmsg_add_u32(m, "subscriptionId",  GetId());
+  htsmsg_add_u32(m, "weight",          GetWeight());
+  htsmsg_add_u32(m, "timeshiftPeriod", (uint32_t)~0);
+  htsmsg_add_u32(m, "normts",          1);
+  htsmsg_add_u32(m, "queueDepth",      2000000);
+  tvhdebug("demux subscribe to %d",    GetChannelId());
+
+  /* Send and Wait for response */
+  if (restart)
```
Glenn-1990: For some reason there was no locking (for m_conn) in SendSubscribe and Sendunsubscribe, although there is locking (sometimes) in the caller (demux), I leaved it this way as I was not sure about it..

src/HTSPTypes.h
```diff
@@ -73,6 +73,21 @@ enum eSubscriptionWeight {
   SUBSCRIPTION_WEIGHT_SERVERCONF = 0,
 };
 
+enum eSubsriptionState
```
Jalle19: Can you move this to `Subscription.h`? I want to get rid of `HTSPTypes.h` in the long run.

src/tvheadend/Subscription.cpp
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  /* Build message */
+  htsmsg_t *m = htsmsg_create_map();
+  htsmsg_add_s32(m, "channelId",       GetChannelId());
+  htsmsg_add_u32(m, "subscriptionId",  GetId());
+  htsmsg_add_u32(m, "weight",          GetWeight());
+  htsmsg_add_u32(m, "timeshiftPeriod", (uint32_t)~0);
+  htsmsg_add_u32(m, "normts",          1);
+  htsmsg_add_u32(m, "queueDepth",      2000000);
```
Jalle19: Now that we have a class, would you mind adding a constant for this magic number?

src/tvheadend/Subscription.h
```diff
@@ -0,0 +1,114 @@
...
+#include "../HTSPTypes.h"
+
+extern "C"
+{
+#include "libhts/htsmsg.h"
+}
+
+class CHTSPConnection;
+
+class Subscription
```
Jalle19: Please wrap this inside the `tvheadend` namespace so it matches the directory layout.

src/tvheadend/Subscription.h
```diff
@@ -0,0 +1,114 @@
...
+  int32_t           GetSpeed() const;
+  eSubsriptionState GetState() const;
+
+  /**
+   * Subscribe to a channel on the backend
+   * @param channelId the channel to subscribe to
+   * @param weight the desired subscription weight
+   * @param restart restart the current subscription (i.e. after lost connection), other parameters will be ignored
+   */
+  void SendSubscribe(uint32_t channelId = 0, uint32_t weight = SUBSCRIPTION_WEIGHT_NORMAL, bool restart = false);
```
Jalle19: Why does `channelId` have a default value?

src/tvheadend/Subscription.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, write to
+ *  the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ *  http://www.gnu.org/copyleft/gpl.html
+ *
+ */
+
+#include <string>
+#include <mutex>
```
Jalle19: This seems out of place, `<mutex>` is for the C++11 mutex class which AFAICT is not used here.

src/tvheadend/Subscription.h
```diff
@@ -0,0 +1,114 @@
...
+#include <condition_variable>
```
Jalle19: Same here (probably a leftover from the subscription handling PR)?

src/tvheadend/Subscription.cpp
Jalle19: You have to follow the chain all the way up, usually when a lock is seemingly missing it's because the same lock is used further up. Or it could be a bug of course (wouldn't be the first time we screwed up the connection locking).

src/tvheadend/Subscription.h
Glenn-1990: Indeed :p

src/tvheadend/Subscription.cpp
Jalle19: Yup, in the previous code `SendSubscribe` was called from a) `Connected` (which is called by `CHTSPConnection` itself, the lock is taken there) and b) from `Open` which already locks the mutex. This should be looked over at some point, ideally you want the lock scope to be as small as possible but in this case it gets complicated since `Connected` is involved which is always called with the lock already taken.

Leave this be for now, since you didn't change any of the calling code this should still be correct.

Jalle19:
Thanks @Glenn-1990, apart from my comments this looks good, once this is in I can rebase the profile stuff so we can eventually get that merged as well.

src/tvheadend/Subscription.h
Glenn-1990: Because we call "m_subscription.SendSubscribe(true)" when the connection to the backend is restored, we do not care what the channelId is in this case as we want just to restart the current stream (with the curret channel id)

Glenn-1990:
@Jalle19 reworked

src/tvheadend/Subscription.h
```diff
@@ -0,0 +1,141 @@
...
+extern "C"
+{
+#include "libhts/htsmsg.h"
+}
+
+/* streaming uses a weight of 100 by default on the tvh side  */
+/* lowest configurable streaming weight in tvh is 50          */
+/* predictive tuning should be lower to avoid conflicts       */
+/* weight 0 means that tvh will use the weight of it's config */
+enum eSubscriptionWeight {
```
Jalle19: Could you move this, `eSubscriptionState` and `PACKET_QUEUE_DEPT` inside the `tvheadend` namespace?

src/tvheadend/Subscription.h
```diff
@@ -0,0 +1,141 @@
...
+  SUBSCRIPTION_SCRAMBLED                      = 4,  /* subscription is not running because the channel is scrambled */
+  SUBSCRIPTION_NOSIGNAL                       = 5,  /* subscription is not running because of a weak/no input signal */
+  SUBSCRIPTION_TUNINGFAILED                   = 6,  /* subscription could not be started because of a tuning error */
+  SUBSCRIPTION_USERLIMIT                      = 7,  /* userlimit is reached, so we could not start this subscription */
+  SUBSCRIPTION_NOACCESS                       = 8,  /* we have no rights to watch this channel */
+  SUBSCRIPTION_UNKNOWN                        = 9,  /* subscription state is unknown, also used for pretuning and posttuning subscriptions */
+  SUBSCRIPTION_PREPOSTTUNING                  = 10, /* used for pre and posttuning subscriptions (we do not care what the actual state is) */
+};
+
+static const int PACKET_QUEUE_DEPT = 2000000;
```
Jalle19: Typo, should be `DEPTH`.

Jalle19:
@ksooo any objections once the final comments are taken care of?

Glenn-1990:
@Jalle19 comments addressed ;-)

src/HTSPDemuxer.cpp
```diff
@@ -191,18 +180,16 @@ bool CHTSPDemuxer::Seek
...
     return;
-  m_subscription.speed = speed;
-  SendSpeed();
+  m_subscription.SendSpeed(speed);
 }
 
 void CHTSPDemuxer::Weight ( enum eSubscriptionWeight weight )
 {
-  if (!m_subscription.active || m_subscription.weight == weight)
+  if (!m_subscription.IsActive() || m_subscription.GetWeight() == (uint32_t)weight)
```
ksooo: Could you please use a C++ style cast here?

src/tvheadend/Subscription.cpp
```diff
@@ -0,0 +1,301 @@
...
+    SetId(GetNextId());
+    SetSpeed(1000); //set back to normal
+  }
+
+  /* Build message */
+  htsmsg_t *m = htsmsg_create_map();
+  htsmsg_add_s32(m, "channelId",       GetChannelId());
+  htsmsg_add_u32(m, "subscriptionId",  GetId());
+  htsmsg_add_u32(m, "weight",          GetWeight());
+  htsmsg_add_u32(m, "timeshiftPeriod", (uint32_t)~0);
```
ksooo: I know this is bsically copied code, but could you please change it to use a C++ style cast?

src/tvheadend/Subscription.cpp
```diff
@@ -0,0 +1,301 @@
...
+  htsmsg_destroy(m);
+  tvhdebug("demux successfully unsubscribed from channel id %d, subscription id %d", GetChannelId(), GetId());
+}
+
+bool Subscription::SendSeek(int time)
+{
+  /* Build message */
+  htsmsg_t *m = htsmsg_create_map();
+  htsmsg_add_u32(m, "subscriptionId", GetId());
+  htsmsg_add_s64(m, "time",           (int64_t)time * 1000LL);
```
ksooo: same here.

src/tvheadend/Subscription.cpp
```diff
@@ -0,0 +1,301 @@
...
+  }
+  if (m)
+    htsmsg_destroy(m);
+}
+
+void Subscription::ParseSubscriptionStatus ( htsmsg_t *m )
+{
+  /* Not for preTuning and postTuning subscriptions */
+  if (GetWeight() == (uint32_t)SUBSCRIPTION_WEIGHT_PRETUNING ||
+      GetWeight() == (uint32_t)SUBSCRIPTION_WEIGHT_POSTTUNING)
```
ksooo: ...and here ... and please elsewhere in the diff. ;-)

src/tvheadend/Subscription.h
ksooo: You forgot to put eSubscriptionWeight inside the tvheadend namespace.

Glenn-1990:
@Jalle19 @ksooo addressed final comments

ksooo:
Thanks. All fine by me.
@Jalle19 your button.

Jalle19:
@Glenn-1990 sorry to continue bothering you but can you please rebase?

Glenn-1990:
@Jalle19 rebased ;-)

PR #122, opened 2015-10-19
Parse subscriptionError also for dvr entries

Glenn-1990:
Currently when an active recording isn't running because there is no free adapter available,
kodi will still show the timer as 'recording' (status field in timers window).

Now, for active recordings we can parse 'subscriptionError' as each active recording will have an active subscription. If 'subscriptionError' returns 'noFreeAdapter' we flag the timer state to 'PVR_TIMER_STATE_CONFLICT_NOK' and kodi will show 'conflict' as status in the timers window now.

No free adapter is practically the same as a conflict in my eyes..

Additionally we could also parse the other errors like scrambled, no signal etc and show them as error in the kodi GUI, what do you think?
https://github.com/xbmc/xbmc/blob/master/xbmc/addons/include/xbmc_pvr_types.h#L192

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -1750,6 +1750,19 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
       rec.SetError(str);
   }
   
+  /* A running recording will have an active subscription assigned to it */
+  if (rec.GetState() == PVR_TIMER_STATE_RECORDING)
+  {
+    /* Parse subscription error */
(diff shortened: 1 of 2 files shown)


Jalle19:
Sounds good.

> Additionally we could also parse the other errors like scrambled, no signal etc and show them as error in the kodi GUI, what do you think?

I don't think it makes sense to attempt to map any other errors to a `PVR_TIMER_STATE`.

Jalle19:
@ksooo good to go?

ksooo:
@Jalle19 shove it in.

PR #125, opened 2015-10-21
Remove actual file when lifetime expires

Glenn-1990:
From htsp v24 and on, it's possible to use lifetime where it's designed for :p

### DIFF
--- a/src/AutoRecordings.cpp
+++ b/src/AutoRecordings.cpp
@@ -102,7 +102,7 @@ void AutoRecordings::GetAutorecTimers(std::vector<PVR_TIMER> &timers)
                               : PVR_TIMER_STATE_DISABLED;
     tmr.iTimerType         = TIMER_REPEATING_EPG;
     tmr.iPriority          = tit->second.GetPriority();
-    tmr.iLifetime          = tit->second.GetRetention();
+    tmr.iLifetime          = tit->second.GetLifetime();
     tmr.iMaxRecordings     = 0;                    // not supported by tvh
     tmr.iRecordingGroup    = 0;                    // not supported by tvh
 
@@ -156,7 +156,11 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
 
   htsmsg_add_s64(m, "startExtra", timer.iMarginStart);
   htsmsg_add_s64(m, "stopExtra",  timer.iMarginEnd);
-  htsmsg_add_u32(m, "retention",  timer.iLifetime);
+  htsmsg_add_u32(m, "retention",  timer.iLifetime); // remove from tvh database
+
+  if (m_conn.GetProtocol() >= 24)
+    htsmsg_add_u32(m, "removal",  timer.iLifetime); // remove from disk
+
   htsmsg_add_u32(m, "daysOfWeek", timer.iWeekdays);
 
   if (m_conn.GetProtocol() >= 20)
@@ -328,6 +332,16 @@ bool AutoRecordings::ParseAutorecAddOrUpdate(htsmsg_t *msg, bool bAdd)
   else if (bAdd)
   {
     tvherror("malformed autorecEntryAdd: 'retention' missing");
+    return false;
+  }
+
+  if (!htsmsg_get_u32(msg, "removal", &u32))
+  {
+    rec.SetRemoval(u32);
+  }
+  else if (bAdd && (m_conn.GetProtocol() >= 24))
+  {
+    tvherror("malformed autorecEntryAdd: 'removal' missing");
     return false;
   }
 
--- a/src/TimeRecordings.cpp
+++ b/src/TimeRecordings.cpp
@@ -82,7 +82,7 @@ void TimeRecordings::GetTimerecTimers(std::vector<PVR_TIMER> &timers)
                               : PVR_TIMER_STATE_DISABLED;
     tmr.iTimerType         = TIMER_REPEATING_MANUAL;
     tmr.iPriority          = tit->second.GetPriority();
-    tmr.iLifetime          = tit->second.GetRetention();
+    tmr.iLifetime          = tit->second.GetLifetime();
     tmr.iMaxRecordings     = 0;                    // not supported by tvh
     tmr.iRecordingGroup    = 0;                    // not supported by tvh
     tmr.iPreventDuplicateEpisodes = 0;             // n/a for manual timers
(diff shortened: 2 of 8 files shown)


Jalle19:
Do we have to check the protocol version before adding these parameters?

Glenn-1990:
An older htsp version should just ignore this field I think? @ksooo

ksooo:
> An older htsp server would just ignore this field I think?

Not sure what the htsp spec says on this, but tvheadend ignores unknown request fields. But to be on the save side I think we should do a protocol check here (we did so with other fields that where added after minimum htsp version for pvr.hts 2.2.x

Jalle19:
Adding checks also makes it clearer to a reader which functionality can be expected.

Glenn-1990:
I'm thinking if we need to change the htsp getter to removal instead of retention..

Jalle19:
> I'm thinking if we need to change the htsp getter to removal instead of retention..

Do you mean we should set the value of Kodi's "life time" to the value that tvheadend calls "removal"? Makes sense to me, but that would also mean the value is different depending on what version of the server you use. Someone could do something stupid like have "lifetime" set to 7 days but "removal" set to 365 days.

Glenn-1990:
https://tvheadend.org/issues/3197

Jalle19:
That is a very good idea, I don't know why it wasn't implemented like that from the beginning.

Jalle19:
I don't think it's worth complicating the logic, we should probably just set lifetime and removal to the same value.

src/AutoRecordings.cpp
```diff
@@ -102,7 +102,8 @@ void AutoRecordings::GetAutorecTimers(std::vector<PVR_TIMER> &timers)
                               : PVR_TIMER_STATE_DISABLED;
     tmr.iTimerType         = TIMER_REPEATING_EPG;
     tmr.iPriority          = tit->second.GetPriority();
-    tmr.iLifetime          = tit->second.GetRetention();
+    tmr.iLifetime          = tit->second.GetRetention() < tit->second.GetRemoval() ?
```
Jalle19: `std::min()`

src/TimeRecordings.cpp
```diff
@@ -82,7 +82,8 @@ void TimeRecordings::GetTimerecTimers(std::vector<PVR_TIMER> &timers)
                               : PVR_TIMER_STATE_DISABLED;
     tmr.iTimerType         = TIMER_REPEATING_MANUAL;
     tmr.iPriority          = tit->second.GetPriority();
-    tmr.iLifetime          = tit->second.GetRetention();
+    tmr.iLifetime          = tit->second.GetRetention() < tit->second.GetRemoval() ?
```
Jalle19: Same here

src/Tvheadend.cpp
```diff
@@ -385,8 +385,9 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
       /* Priority */
       rec.iPriority = recording.GetPriority();
 
-      /* Retention */
-      rec.iLifetime = recording.GetRetention();
+      /* Lifetime = the smallest value, retention or removal */
+      rec.iLifetime = recording.GetRetention() < recording.GetRemoval() ?
```
Jalle19: Same here and in other places

src/tvheadend/entity/Recording.h
```diff
@@ -60,7 +60,8 @@ namespace tvheadend
         m_stopExtra(0),
         m_state(PVR_TIMER_STATE_ERROR),
         m_retention(99), // Kodi default - "99 days"
-        m_priority(50) // Kodi default - "normal"
+        m_removal(-1),   // max value to stay compatible with older backends
+        m_priority(50)   // Kodi default - "normal"
```
Jalle19: While you're at it, perhaps you could add some constants for this and the default retention value?

Glenn-1990: Actually I don't think that these values are needed, they are being overwritten by the actual (htsp) values? @ksooo

Jalle19: Every member variable has to have a default value, even if it's being overwritten at runtime.

Glenn-1990: Offcource, but '0' should be enough

Jalle19: @ksooo up to you, IMO we can leave this as is for now.

Glenn-1990:
Addressed comments, removal seems still a bit buggy in tvh..
It might be '0' instead of default config when adding an autorec through the webui.

Glenn-1990:
https://github.com/tvheadend/tvheadend/pull/741

da-anda:
if we delete files after retention time is running out, I'd like to suggest to add a possibility to keep files forever. I personally don't want any software to delete any recording automatically and thus would prefer to have a way to prevent deletion completely.
The ArgusTV server had retention options "until space is needed" and "forever" in addition to the day values. No idea if something like this would be possible for TVH as well.

src/Tvheadend.cpp
```diff
@@ -778,7 +778,7 @@ bool CTvheadend::CreateTimer ( const Recording &tvhTmr, PVR_TIMER &tmr )
           tvhTmr.GetDescription().c_str(), sizeof(tmr.strSummary) - 1);
   tmr.state              = tvhTmr.GetState();
   tmr.iPriority          = tvhTmr.GetPriority();
-  tmr.iLifetime          = tvhTmr.GetRetention();
+  tmr.iLifetime          = std::min(tvhTmr.GetRetention(), tvhTmr.GetRemoval()); // Lifetime = the smallest value, retention or removal
```
Jalle19: Since this `std::min` logic is now used in four places I would suggest we move it to a separate method, e.g. `Recording::GetLifetime()`.

Jalle19:
@da-anda you should file a feature request on tvheadend.org for that.
@Glenn-1990 apart from my comment I think this looks good
@ksooo what about you?

ksooo:
@Glenn-1990 please update changelog and HTSP_CLIENT_VERSION (https://github.com/kodi-pvr/pvr.hts/blob/master/src/Tvheadend.h#L71). I'm fine with the rest.

Glenn-1990:
@ksooo updated ;-)

Jalle19:
@Glenn-1990 did you see my comment? Somehow your latest push marked it as outdated but it stills seems to be valid.

Glenn-1990:
@Jalle19 done ;-)

Jalle19:
+1, @ksooo your button

da-anda:
could we add a add-on setting to turn this feature on/off until tvh supports endless retention?

Glenn-1990:
@ksooo @da-anda 
If I remember correctly, one of the tvh devs said some time ago that there was no need for a keep forever option in tvh as the user can set the retention/removal value very high..

So the best thing we can do is assigning "forever" to INT_MAX I guess:
https://github.com/Glenn-1990/pvr.hts/commit/6d5611599adc7f4f8c9a0458418bfcd59e1eecd3

ksooo:
> If I remember correctly, one of the tvh devs said some time ago that there was no need for a keep forever option in tvh as the user can set the retention/removal value very high.

lol! I think this should be an official feature of tvh, not just a hack with a "very high" value. Until it is an official tvh feature we should not hack around in the addon, imo.

@Glenn-1990 You know the tvh code very well, I guess. Maybe you can PR this tvh feature? Just asking in case you've got some time and you're interested.

Glenn-1990:
@ksooo @da-anda 
https://github.com/tvheadend/tvheadend/pull/760

da-anda:
thanks @Glenn-1990 :)
Do you think something like "until space is needed" might also make sense in TVH? I know it from the ArgusTV server and it sounded nice.

Glenn-1990:
@da-anda That's a bit harder as we need to monitor the free space etc, and check for it on regular intervals when there are active recordings. Also currently lifetime is an integer defined in days, so we can't use that for 'until space is needed' so we need lots of additional logic, but you can always create a feature request for that.

Personally I don't need that as I run a postprocessor script on recordings. That script will delete the oldest recordings when the free disk size exceeds some threshold.
https://tvheadend.org/boards/5/topics/4300?page=1

PR #127, opened 2015-10-26
Replace the tvhxxx() macros with a separate logger utility

Jalle19:
@ksooo here it is. I'm not particularly proud of using a private singleton for storing the implementation method in a static context but for some reason I couldn't make `m_implementation` static. If you want to feel free to experiment with it.

Only managed to get rid of one `Tvheadend.h` include though, `Settings` still requires the `XBMC` pointer so we can't remove it yet (it could include `client.h` instead though).

### DIFF
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -19,6 +19,7 @@
  *
  */
 
+#include "tvheadend/utilities/Logger.h"
 #include "Tvheadend.h"
 #include "xbmc_codec_descriptor.hpp"
 
@@ -28,6 +29,7 @@ using namespace std;
 using namespace ADDON;
 using namespace PLATFORM;
 using namespace tvheadend;
+using namespace tvheadend::utilities;
 
 CHTSPDemuxer::CHTSPDemuxer ( CHTSPConnection &conn )
   : m_conn(conn), m_pktBuffer((size_t)-1),
@@ -46,7 +48,7 @@ void CHTSPDemuxer::Connected ( void )
   /* Re-subscribe */
   if (m_subscription.IsActive())
   {
-    tvhdebug("demux re-starting stream");
+    Logger::Log(LogLevel::DEBUG, "demux re-starting stream");
     m_subscription.SendSubscribe(0, 0, true);
     m_subscription.SendSpeed(0, true);
 
@@ -82,7 +84,7 @@ void CHTSPDemuxer::Abort0 ( void )
 bool CHTSPDemuxer::Open ( uint32_t channelId, enum eSubscriptionWeight weight )
 {
   CLockObject lock(m_conn.Mutex());
-  tvhdebug("demux open");
+  Logger::Log(LogLevel::DEBUG, "demux open");
 
   /* Close current stream */
   Close0();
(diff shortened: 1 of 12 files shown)


src/tvheadend/utilities/Logger.h
```diff
@@ -0,0 +1,106 @@
...
+      ERROR,
+      INFO,
+      DEBUG,
+      TRACE
+    };
+
+    /**
+     * Short-hand for a function that acts as the logger implementation
+     */
+    typedef std::function<void(LogLevel level, const char *message)> LogImplementation;
```
Jalle19: Should probably be renamed to `LoggerImplementation`.

src/tvheadend/utilities/Logger.h
```diff
@@ -0,0 +1,106 @@
...
+      /**
+       * Logs the specified message using the specified log level
+       * @param level the log level
+       * @param message the log message
+       * @param ... parameters for the log message
+       */
+      static void Log(LogLevel level, const std::string &message, ...)
+      {
+        char buffer[MESSAGE_BUFFER_SIZE];
+        std::string logMessage = "pvr.hts - " + message;
```
Jalle19: In order to use this class in other PVR addons (I'm thinking about `pvr.vbox `mostly) this prefix should probably be configurable.

ksooo:
Nice.

Regardig your comments to yourself - yep, makes sense ;-)

> ...but for some reason I couldn't make m_implementation static. If you want to feel free to experiment with it.

No time for experiments here atm. :-/

Jalle19:
@ksooo I changed some semantics and made the log prefix configurable, plus rebased. In my opinion this is ready to go in. The risk for breakage is very low since there's no logic changes.

src/HTSPDemuxer.cpp
```diff
@@ -417,7 +419,7 @@ void CHTSPDemuxer::ParseSubscriptionStart ( htsmsg_t *m )
            some versions of tvheadend and is here for backward compatibility. */
         if (stream.iWidth == 0 || stream.iHeight == 0)
         {
-          tvhinfo("Ignoring subscriptionStart, stream details missing");
+          Logger::Log(LogLevel::INFO, "Ignoring subscriptionStart, stream details missing");
```
ksooo: Not related to this PR, but why is this INFO, not DEBUG?

src/tvheadend/Subscription.cpp
```diff
@@ -247,7 +249,7 @@ void Subscription::ParseSubscriptionStatus ( htsmsg_t *m )
   const char *status = htsmsg_get_str(m, "status");
 
   if (status != NULL)
-    tvhinfo("Bad subscription status: %s", status);
+    Logger::Log(LogLevel::INFO, "Bad subscription status: %s", status);
```
ksooo: Same here. Why INFO, not DEBUG or ERROR?

ksooo:
Like always, I'd love a changelog entry. ;-)

Jalle19:
@ksooo added changelog entry. I can change the log level of those lines you mentioned in a separate PR once this is merged. No idea why the levels were set like this to begin with, the first one I suspect is a leftover from when I wanted to test it without having debugging enabled.

ksooo:
> I can change the log level of those lines you mentioned in a separate PR once this is merged.

 That would be great.

ksooo:
Thanks for another round of cleaning up the code.

PR #133, opened 2015-11-08
Added support for oneshot timer enable/disable (HTSP v23 feature), closes #131

### LINKED ISSUE #131: Kodi shows disabled recordings as upcoming, when they're never going to happen
As discussed with @Jalle19 on #hts...

Since https://github.com/tvheadend/tvheadend/commit/fa5795379980546c76c5686633215a56d72365b4, tvh gives the user the ability to disable an unwanted recording event rather than deleting it. This was to prevent the repeated dance of unwanted-delete-EPG match-reschedule-back in the list-unwanted-delete...

However, Kodi still shows the event as upcoming (e.g. "next recording" on the main TV screen) even if it's disabled, as there's no way for the addon to filter out these disabled events.

It appears that @ksooo has already updated the HTSP spec and bumped to v23 to accommodate the enabled flag.

ksooo:


src/tvheadend/entity/Recording.h
```diff
@@ -172,6 +177,7 @@ namespace tvheadend
       void SetPriority(uint32_t priority) { m_priority = priority; }
 
     private:
+      uint32_t         m_enabled;
```
Jalle19: IMO if this is used as a boolean we should store it as that instead of letting the inner workings of HTSP creep up to this level.

ksooo: This is how all the "HTSP" types in pvr.hts currently work. They store the htsp types internally, offer setters for htsp types and getters for Kodi types. So, all good imo, because in line with the rest. Okay?

ksooo:
Added changelog.txt entry (no new version number, though).

Good to go?

src/tvheadend/entity/Recording.h
Jalle19: If that's the case then it's fine, I somehow thought we converted to booleans in other places.

PR #136, opened 2015-11-15
Fix GetStatus 'lost connection' handling.

ksooo:
This fixes addon function <code>ADDON_GetStatus</code> 

* to only return "connection lost" if not already suspended (would cause "connection lost" error toast while suspending and "connection restored" on wakeup, once https://github.com/ksooo/xbmc/tree/pvr-addon-connection-lost-handling is merged)
* to not overwrite m_CurStatus (same logic now as in function <code>ADDON_Create</code>)
* to return other status then "connection lost" after this state got set once, esp. after tvh actually reconnected. ;-)

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -120,8 +120,8 @@ ADDON_STATUS ADDON_GetStatus()
   CLockObject lock(g_mutex);
 
   // Check that we're still connected
-  if (m_CurStatus == ADDON_STATUS_OK && !tvh->IsConnected())
-    m_CurStatus = ADDON_STATUS_LOST_CONNECTION;
+  if (m_CurStatus == ADDON_STATUS_OK && !tvh->IsSuspended() && !tvh->IsConnected())
+    return ADDON_STATUS_LOST_CONNECTION;
 
   return m_CurStatus;
 }
--- a/src/Tvheadend.h
+++ b/src/Tvheadend.h
@@ -208,6 +208,8 @@ public:
   inline bool IsConnected       ( void ) const { return m_ready; }
   bool        WaitForConnection ( void );
 
+  inline bool IsSuspended       ( void ) const { return m_suspended; }
+
   inline PLATFORM::CMutex& Mutex ( void ) { return m_mutex; }
 
   void        OnSleep ( void );
@@ -561,6 +563,10 @@ public:
   inline bool IsConnected ( void ) const
   {
     return m_conn.IsConnected();
+  }
+  inline bool IsSuspended ( void ) const
+  {
+    return m_conn.IsSuspended();
   }
   inline void OnSleep ( void )
   {


src/client.cpp
```diff
@@ -120,8 +120,8 @@ ADDON_STATUS ADDON_GetStatus()
   CLockObject lock(g_mutex);
 
   // Check that we're still connected
-  if (m_CurStatus == ADDON_STATUS_OK && !tvh->IsConnected())
-    m_CurStatus = ADDON_STATUS_LOST_CONNECTION;
+  if (m_CurStatus == ADDON_STATUS_OK && !tvh->IsSuspended() && !tvh->IsConnected())
+    return ADDON_STATUS_LOST_CONNECTION;
```
Jalle19: I'm not really sure what the implications of this are. If we just return here, do we use `m_CurStatus` anywhere else? Do you think it's even needed?

FernetMenta: ADDON_STATUS_LOST_CONNECTION is PVR only and should be handled by the addon itself. Finally ADDON_STATUS_LOST_CONNECTION should go away.

ksooo: @Jalle19 I think we need to fix some more addon status handling stuff in pvr.hts. Currently, there is a mixture of directly returning an ADDON_STATUS without setting m_CurStatus and setting m_CurStatus (before returning it). IMO this should be changed to consequently setting m_CurStatus on status change. IMO, we cannot get rid of m_CurrStatus, for instance because we need to implement ADDON_GetStatus properly. What do you think?

@FernetMenta pvr.hts already handles ADDON_STATUS_LOST_CONNECTION by itself (detect loss, automatically reconnect, ...). Only "problem" is to report loss/restore of connection back to the Kodi user. How should that work without this addon status? I'm not a friend of addon calling XBMC->QueueNotification as this is somehow "hardwired" and for instance gives no control whether (also) to write an activity log entry, for instance.

FernetMenta: If you want an addon to write something into activity log, you need to create an API method for this. Something like XBMC->ActivityLog

ksooo: I don't want to let the add-on directly write  to activity log nor directly trigger a UI notification. I want it to report status changes and Kodi to react consistently across add-on boundaries on well defined status changes.

FernetMenta: promoting this function upper layers (generalization) is wrong here. even simple client has no server connection. from kodi POV "connection lost" is not a well defined status. this was a quick hack that never should have been added as a states.
you could add some method to PVR API that allows to signal state and have pvr component handle it. kodi's general addon loader should not be aware of addon specific states.

ksooo: Makes sense, will put this in "my" backlog for later. Maybe I'll find some time in v17 timeframe for this.

But want to keep this PR to fix current status handling issues, though.

FernetMenta: fine. would be great if you found some time for K****

Jalle19: @FernetMenta returning ADDON_STATUS_LOST_CONNECTION triggers the "Connection lost" notification in Kodi, which can be disabled by checking the "Do not show connection warnings" setting in Kodi. I don't see why we'd have to remove the status, many clients have a concept of connecting to a backend which implies there is a possibility that the connection is lost. pvr.iptvsimple obviously doesn't have this concept, but it's not a true PVR addon either.

FernetMenta: > I don't see why we'd have to remove the status

It is PVR only and does spoil the general addon handling. I do not say this should be completely dropped but implemented properly.

ksooo:
>  I do not say this should be completely dropped but implemented properly.

 So, you mean to drop ADDON_STATE_LOST_CONNECTION (incompatible Addon API change!), add a new PVR API function like <code>bool IsConnected()</code> and let PVR pull (?!) for connection state changes using this new function, doing GUI notifications and the like in case?

FernetMenta:
> So, you mean to drop ADDON_STATE_LOST_CONNECTION

yes, correct the mistake of adding this state to the generic addon handler

I would avoid polling and add a method to https://github.com/xbmc/xbmc/blob/master/addons/library.xbmc.pvr/libXBMC_pvr.h
Then addons can signal their state and pvr and take appropriate action on event, without polling

ksooo:
@FernetMenta yeah sounds good.

Jalle19:
@FernetMenta good point

ksooo:
Closing this PR for now. Will come up with something better in K***** timeframe.

ksooo:
@FernetMenta my first take of the new connection state change callback is ready:

kodi core: https://github.com/ksooo/xbmc/commit/faa86b5c3563a2ae9d06c9ef0ec78d3c04fd96da
pvr.hts: https://github.com/ksooo/pvr.hts/commit/7f70ecefff184c7c6e41e5d393ad0267688835c1

It's straight forward and cleans up pvr.hts code. pvr.hts is no longer using ADDON_STATUS_LOST_CONNECTION. API version bump is (ofc) still missing, also removal of ADDON_STATUS_LOST_CONNECTION.

Several functional advantages:
* speedup of pvr.hts startup as it no longer waits for backend connection in ADDON_Create.
* no more announcing "connection lost" in a loop (due to kodi-controlled recreation of the addon in a loop on "connection lost" return code)
* notifies only on real state changes.
* new notify on "connection restore"
* new notifies for e.g. login error
* no more special handling for htsp protocol version check and mismatch notify

What do you think?

Jalle19:
I like it!

FernetMenta:
me too!

PR #137, opened 2015-11-15
Fix WIN32 build breaker (added missing include for std::min)

ksooo:
Regression introduced by #125 :-(

### DIFF
--- a/src/tvheadend/entity/Recording.h
+++ b/src/tvheadend/entity/Recording.h
@@ -21,6 +21,7 @@
  *
  */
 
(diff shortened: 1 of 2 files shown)


ksooo:
Will merge immediately, because fix is a nobrainer and pvr.hts currently does not build build on Kodi WIN32 Jenkins. @MartijnKaijser fyi

Jalle19:
I was going to comment on that but I figured it was included somewhere else. Next time I'll complain about it :-)

Issue #126, opened 2015-10-24
Settings rework v2


Glenn-1990:
As mentioned before, the advanced tab is getting to big.
So I have some suggestion which I'm happy to implement.

3 settings tabs:
tab 1) connection settings: leave it as it is now
tab 2) streaming settings: 
          ->a] predictive tuning
          ->b] streaming profile from https://github.com/kodi-pvr/pvr.hts/pull/119
          ->c] timeshift config (to be created) with the following options, 
                 ---->1) server config (default), 
                 ---->2) force off, 
                 ---->3) select  minutes. 
          ->d] enable subscription conflict managment https://github.com/kodi-pvr/pvr.hts/pull/79
tab 3) dvr:
         ->a] autorecording start window behavior
         ->b] default value for duplicate detection
         ->c] default value for priority https://github.com/kodi-pvr/pvr.hts/issues/124 
tab 4) advanced settings:
         ->a] epg async -> maybe it's time to set this on as default?
         ->b] debug trace

This should be the final settings refactor :p
So, what do you think?

Jalle19:
Sounds good in general, feel free to work on it. Can you elaborate on the timeshift settings?

Glenn-1990:
@Jalle19 
Well now, timeshift is enabled when it's enabled on the server, but a user might want to use timeshift on his main box, but not on the bedroom box, as multiple timeshift instances are stressing for the server..

If the user uses predictive tuning (till 10 subscriptions) with timeshift enabled, that will also stress the server quite a bit.. Allow him to disable timeshift

Additionally we can add also a period, so the master client can be configured with a big timeshift period and the bedroom client can be configured with a small timeshift period.

Jalle19:
Is this controlable via HTSP?

Glenn-1990:
Yes we can set the timeshift period on subsribe, 0 = force off, infinity = server config

Jalle19:
The part about setting it to infinity is not mentioned in the HTSP spec, are you sure that's the case? If so someone should probably update the wiki.

Jalle19:
I would keep it simple for now and only have a checkbox like "Disable timeshifting" which defaults to false. To me it sounds like being able to control the timeshift buffer size from the client is a very edge case.

@ksooo what do you think?

Glenn-1990:
It's currently set to infinity https://github.com/kodi-pvr/pvr.hts/blob/master/src/tvheadend/Subscription.cpp#L120

[...]

PR #138, opened 2015-11-20
Settings rework V2

Glenn-1990:
https://github.com/kodi-pvr/pvr.hts/issues/126

@ksooo @Jalle19 

All timertype specific settings don't take effect until a kodi restart.
I assume that this is a bug in kodi? @ksooo 
Maybe the timertype list needs to be cleared on addon restart (in kodi)?

schermafdruk van 2015-11-20 18 51 45
schermafdruk van 2015-11-20 18 51 43


pvr.hts/resources/language/resource.language.en_gb/strings.po
```diff
@@ -176,10 +200,48 @@ msgctxt "#30361"
...
+#: src/Tvheadend.cpp
+msgctxt "#30370"
+msgid "%d months"
+msgstr ""
+
+#empty strings from id 30371 to 30399
 
 msgctxt "#30400"
-msgid "Predictive tuning"
+msgid "Predictive tuning (experimental)"
```
Jalle19: I wouldn't call this feature experimental anymore, it seems to work pretty well.

ksooo: Yep. +1 for not adding "experimental".

ksooo:
> All timertype specific settings don't take effect until a kodi restart.
I assume that this is a bug in kodi? @ksooo
Maybe the timertype list needs to be cleared on addon restart (in kodi)?

 Yep, runtime changes of timer types are currently not supported by kodi core. Thus, not a bug, just a missing feature. ;-)

ksooo:
Problem is that the different CTimerType instances currently have no chance to observe runtime timertype changes. They are just dumb const containers.

Glenn-1990:
@ksooo @Jalle19 

Using the same defines for days as tvh now:
https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr.h#L125

Keep forever and keep until space needed where implemented in tvh, with these defines, we are instantly supporting them. Using "keep forever" with tvh 4.0 will set the field to int_max, but as soon you update tvh, they will be updated to "keep forever" on the tvh side.

Please can this be merged for jarvis, it's quite critical for users on tvh4.1+
(otherwise lifetime values are messed up)

pvr.hts/resources/language/resource.language.en_gb/strings.po
```diff
@@ -176,7 +204,69 @@ msgctxt "#30361"
...
+#. Recording lifetime representation
+#: src/Tvheadend.cpp
+msgctxt "#30372"
+msgid "%d years"
+msgstr ""
+
+#. Recording lifetime representation
+#: src/Tvheadend.cpp
+msgctxt "#30373"
+msgid "Until space needed"
```
ksooo: What does this actually mean? As I do not even know how tvh implements this I'm asking as a normal user seeing this feature for the first time.

src/HTSPTypes.h
```diff
@@ -54,6 +54,26 @@ typedef enum {
...
+  DVR_RET_1MONTH    = 31,
+  DVR_RET_2MONTH    = 62,
+  DVR_RET_3MONTH    = 92,
+  DVR_RET_6MONTH    = 183,
+  DVR_RET_1YEAR     = 366,
+  DVR_RET_2YEARS    = 731,
+  DVR_RET_3YEARS    = 1096,
+  DVR_RET_ONREMOVE  = INT32_MAX-1, // for retention only
+  DVR_RET_SPACE     = INT32_MAX-1, // for removal only
+  DVR_RET_FOREVER   = INT32_MAX
```
ksooo: I think you should put meaningful comments behind the first and also the last three values so that it becomes clear what these values do actually mean.

pvr.hts/resources/language/resource.language.en_gb/strings.po
Jalle19: At least in my experience tvheadend will just continue recording stuff until your hard drive is full and it will never delete anything unless you do it explicitly.

ksooo: Nonono... IIRC @Glenn-1990 recently implemented "something" for tvh 4.1 that does real diskspace housekeeping.

Jalle19: Perhaps `Let the server decide` would be a better name? I think we use that in another place too so the string could even be reused.

Glenn-1990: This means that the backend will delete the recording when it needs space for a new recording and we are running out of disk space. The oldest recording tagged as "until space needed" will be deleted first. The naming is quite common for advanced STB's and PVR software.

Tivo uses it, mediaportal, argus tv, and many others.
Here a reference of a normal STB:
https://community.shaw.ca/docs/DOC-1536

- Keep Until: (default - Space Needed) This setting allows you to define if the program should be automatically deleted if the cable box requires space for new recordings. You can leave the setting on Space Needed and the oldest recordings will be removed to free up space for new recordings. Alternatively, you can set this option to I Delete which will require you to delete recordings to free up space or risk interruption of scheduled recordings in the event that the PVR is full.

Jalle19: Since it seems to be a common phrase I guess it's fine.

Glenn-1990:
updated ;-)

Glenn-1990:
@ksooo Would be nice to get this in jarvis 
1) avoid weird lifetime behaviour when using tvh 4.1
2) defaults for priority, lifetime and dubdetection is a must, isn't it?

Jalle19:
I don't mind. @ksooo?

ksooo:
@Glenn-1990 thx. Please update v2.2.10 changlog (no version bump) and we're ready for merging this.

Glenn-1990:
@ksooo changelog updated ;-)

ksooo:
Cool. Thanks a lot.

PR #145, opened 2015-12-18
Support for autorec/timerec update (new htsp v25 methods)

Glenn-1990:
Since htsp v25 it's possible to do proper autorec/timerc update, so let's use it.
Unfortunately, the retention field was also changed.
https://github.com/tvheadend/tvheadend/blob/master/src/htsp_server.c#L981

So I think the best thing we can do is using "retention" for htsp < 25 en "removal" for newer versions. 
Removal is the main lifetime setting in recent tvh master builds, "retention" is set on "remove when file removed"

### DIFF
--- a/src/AutoRecordings.cpp
+++ b/src/AutoRecordings.cpp
@@ -158,10 +158,15 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
 
   htsmsg_add_s64(m, "startExtra", timer.iMarginStart);
   htsmsg_add_s64(m, "stopExtra",  timer.iMarginEnd);
-  htsmsg_add_u32(m, "retention",  timer.iLifetime); // remove from tvh database
 
-  if (m_conn.GetProtocol() >= 24)
-    htsmsg_add_u32(m, "removal",  timer.iLifetime); // remove from disk
+  if (m_conn.GetProtocol() >= 25)
+  {
+    htsmsg_add_u32(m, "removal",   timer.iLifetime);  // remove from disk
+    htsmsg_add_u32(m, "retention", DVR_RET_ONREMOVE); // remove from tvh database
+  }
+  else
+    htsmsg_add_u32(m, "retention", timer.iLifetime);  // remove from tvh database
+
 
   htsmsg_add_u32(m, "daysOfWeek", timer.iWeekdays);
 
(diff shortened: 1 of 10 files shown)


src/AutoRecordings.cpp
```diff
@@ -248,8 +253,104 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
 
 PVR_ERROR AutoRecordings::SendAutorecUpdate(const PVR_TIMER &timer)
 {
-  /* Note: there is no "updateAutorec" htsp method, thus delete + add. */
+  if (m_conn.GetProtocol() >= 25)
```
Jalle19: Since the implementation is so differing between these different HTSP versions, maybe it would be clearer to delegate the actual logic to separate functions, i.e. something like this (pseudo):

```
AutoRecordings::SendAutoRecUpdate(timer) {
   if (htsp < 25) {
      SendAutoRecUpdateLegacy();
   } else {
      SendAutorecUpdateActual();
   }
}
```

Naming scheme is obviously up for debate.

src/AutoRecordings.cpp
```diff
@@ -248,8 +253,104 @@ PVR_ERROR AutoRecordings::SendAutorecAdd(const PVR_TIMER &timer)
...
+    std::string strId;
+
+    for (auto tit = m_autoRecordings.begin(); tit != m_autoRecordings.end(); ++tit)
+    {
+      if (tit->second.GetId() == timer.iClientIndex)
+      {
+        strId = tit->second.GetStringId();
+        break;
+      }
+    }
```
Jalle19: Shouldn't we exit early here if the autorec rule cannot be found?

src/TimeRecordings.cpp
```diff
@@ -165,8 +169,63 @@ PVR_ERROR TimeRecordings::SendTimerecAdd(const PVR_TIMER &timer)
...
+    std::string strId;
+
+    for (auto tit = m_timeRecordings.begin(); tit != m_timeRecordings.end(); ++tit)
+    {
+      if (tit->second.GetId() == timer.iClientIndex)
+      {
+        strId = tit->second.GetStringId();
+        break;
+      }
+    }
```
Jalle19: Same here as before.

Glenn-1990:
updated

Glenn-1990:
diff looks quite big but that's because of the retention -> lifetime renaming

Glenn-1990:
@ksooo @Jalle19 updated, tvh master will now use s64 channel id's for timerec/autorec

ksooo:
@Glenn-1990 Please add a changelog entry and thos one is ready for merging. Thanks for this contribution.

Glenn-1990:
@ksooo done ;-)

Jalle19:
@ksooo your button

PR #146, opened 2015-12-18
Isengard: muxer - flush the queued packets with wrong pts on seek

perexg:
(only for Isengard branch)

### DIFF
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -42,7 +42,8 @@ using namespace PLATFORM;
 
 CHTSPDemuxer::CHTSPDemuxer ( CHTSPConnection &conn )
   : m_conn(conn), m_pktBuffer((size_t)-1),
-    m_seekTime(INVALID_SEEKTIME)
+    m_seekTime(INVALID_SEEKTIME),
+    m_seeking(false), m_speedChange(false)
 {
 }
 
@@ -81,6 +82,8 @@ void CHTSPDemuxer::Abort0 ( void )
   CLockObject lock(m_mutex);
   m_streams.Clear();
   m_streamStat.clear();
+  m_seeking = false;
+  m_speedChange = false;
 }
 
 
@@ -151,6 +154,7 @@ bool CHTSPDemuxer::Seek
     return false;
 
   tvhdebug("demux seek %d", time);
+  m_seeking = true;
 
   /* Build message */
   m = htsmsg_create_map();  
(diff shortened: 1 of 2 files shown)


perexg:
Found another issue - see commit ccf7569c5dbc4115fcbd6aff565a844e884c9596 .

perexg:
Found another fatal error in Seek() - see ec37c7b9e26b5f0e2c063537f7215b37f9b65b84 .

Jalle19:
Can you change the PR to point against master directly from Github, or does it not apply cleanly?

ksooo:
Do we still care about Isengard? Jarvis is to be released soon.

perexg:
The commit 4793806196aa0e8118851e1f510d6e62227731ec shows the purpose. Also, Flush() should be called on rewind or FF 4x + .

Glenn-1990:
Why is this only for Isengard? 
Isengard will be obsolete soon as jarvis is almost in RC.

perexg:
Sorry, I work only with precompiled binaries (using kodi 15.2 package in Fedora). The packaing is always delayed a bit.

Jalle19:
Are all these changes part of #155?

perexg:
Yes, they're.

Jalle19:
Okay. @ksooo do we close this or do you want to make a new release for Isengard with applicable backported patches? I believe OE users would get the update since they use a repository for PVR addons.

ksooo:
What is Isengard? ;-) I do not see a need for an Isengard update- If you ask me, I would close this PR.

opdenkamp:
That you don't see the need doesn't mean that it's not useful for others. Not everyone is using the latest greatest version number :)

ksooo:
That's why I said "if you ask me". ;-) If others don't share my point of view, I'm fine with this'll I will not block an Isengard update.

opdenkamp:
I've merged in PRs for older releases quite often when people sent updates. Only thing that needs to be taken care of is that you bump the add-on version numbers properly if the APIs or ABIs are incompatible, so you don't get clashes.

Jalle19:
@seo will you accept addon bumps for OE 6 even though Isengard development has stoppd? If OE won't use a newer version no one else will since other platforms get the addons shipped with Kodi itself.

opdenkamp:
There are a lot more distributions out there that we don't control. There's more out there than just Ubuntu, Windows, etc. One of the big advantages of separating core from stuff like this is that add-ons can still be fixed, even though development of core has moved to a new version.

That being said, I don't care much if you merge this or not, just commenting on the reasons that you're giving for not merging ;)

Jalle19:
@opdenkamp which platforms would that be? As far as I know, OE is the only one with a repository for binary addons.

opdenkamp:
e.g. any Linux distribution that's not called Ubuntu?

Jalle19:
Okay. Let's wait a bit longer before we start backporting stuff in case any fixes show up at the last minute.

CvH:
@Jalle19 
latest/rescent "official" Isengard version is shipped at OE, if there is an update here it would be updated at OE. At least this is the plan ;)

CvH:
@Jalle19 
any news on this for Isengard ?

Jalle19:
Closing this since master has the fixes applied. I hope to be able to do some backports during the weekend.

Issue #135, opened 2015-11-09
Reconnect after tvh failure/restart


dkonerma:
Hi,,

I'm quite sure that pvr.hts has beed quite reliable in re-connection after a tvh failure or restart in the past. Currently (pvr.hts 2.2.9 with tvh 4.1-909~g0a73d7f) this no loinger works.

How to reproduce:
- play a channel with kodi
- stop tvh
- playback stops (paused)
- start tvh
- keeps being paused, tvh UI shows the connection, but no subscription)

On the next channel switch all is working fine again. I would expect that playback resumes automatically.

Thanks in advance
Dietmar.

Jalle19:
Do you remember when this last used to work?

dkonerma:
Unfortunately not... I'm close to bleeding edge for some months now and noticed yesterday by accident. ;-) Do you see the same issue.. wondering if it's just me?

Jalle19:
I'll have to test, I don't normally restart tvheadend or have connection issues so I wouldn't know.

stan86:
Same problem here

ksooo:
Sould be fixed in latest pvr.hts master

PR #147, opened 2015-12-18
Fix for  #135

Glenn-1990:
https://github.com/kodi-pvr/pvr.hts/issues/135

### DIFF
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -5,6 +5,7 @@
 - improved: limit the amount of timer lifetime values
 - improved: separate timer and streaming settings
 - added: default setting for timer lifetime, priority and duplicate detection
+- fixed: restart subscription after a connection restore
 
 2.2.9
 - refactored the code by factoring out lots of things into separate files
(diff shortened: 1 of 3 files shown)


ksooo:
Cool, but can you please elaborate **why** this fixes things?

Glenn-1990:
void SendSubscribe(uint32_t channelId = 0, uint32_t weight = SUBSCRIPTION_WEIGHT_NORMAL, bool restart = false);

Looking to the function declaration, "restart" is the 3rd parameter. 
So we also have to pass the 2 first parameters (which will be ignored anyway), otherwise our boolean will be assigned to the first parameter "channelId"...

Jalle19:
Good catch! This is why I hate methods with many edge-case parameters, makes it easy to do mistakes like this. Too bad the compiler didn't catch it.

ksooo:
What do you guys think about removing the default parameters from declaration of SendSubscribe?

Glenn-1990:
@ksooo yea, that will prevent similar mistakes in the future
PR updated ;-)

Jalle19:
Sounds good to me

Jalle19:
By the way @Glenn-1990 if you want to you can add your name to https://github.com/kodi-pvr/pvr.hts/blob/master/pvr.hts/addon.xml.in#L6, I think you've done enough to warrant having your name there ;-)

ksooo:
> By the way @Glenn-1990 if you want to you can add your name to ...

Definitely a +1 from me for this. :-)

PR #148, opened 2015-12-18
Don't allow a pre or post recording time from 0 minutes.

Glenn-1990:
This is not supported by tvheadend.
When setting this time to "0", the default pre/post recording time from the dvr config will be used.

Test scenario:
1) ensure that the default pre and post recording values for your dvr config are > 0 in tvheadend
2) add an autorec or epg based timer with pre and post record set to "0" in kodi
3) Open the timer settings in kodi after creation, you will see the values from step 1 now...

The best thing we can do is making this error as small as possible, so a pre/post time of "0" will be set to "1".


Jalle19:
To me it sounds like the current behavior is the expected one. A zero could logically be assumed to be "no preference, let the server decide".

ksooo:
I agree to what @Jalle19 said. -1 from me to this PR.

Glenn-1990:
Ok, no problem, closing this now.

Issue #150, opened 2015-12-21
Read only timers


Glenn-1990:
Is it needed to use read only timers? 
User example 1) I added a repeating rule for a show running every evening. 
I'm able to watch this show today, so I want to cancel the timer for this evening, but... this is not possible, I can only delete the complete schedule?
I have to use the tvh webui to handle this case.

User example 2) I have one tuner and am watching live tv, a repeating timer starts recording and interrupts the live stream, so I want to cancel that active recording to allow me watching live tv, this is not possible... Even more, the stop recording button is listed, but doesn't work.

I think canceling a read only timer is possible for tvh 4.1, but maybe we have to handle tvh 4.0 not as read only?

stop recording listed, but not working
![schermafdruk van 2015-12-21 15 14 29](https://cloud.githubusercontent.com/assets/2036512/11932592/e8fe19ec-a7f6-11e5-9640-3fbd1e518ece.png)

unable to cancel a timer which I'll watch in prime time
![schermafdruk van 2015-12-21 15 15 26](https://cloud.githubusercontent.com/assets/2036512/11932737/d3c298d6-a7f7-11e5-8f6e-7563d734eddc.png)

I didn't investigate yet where the problem lies (kodi or pvr.hts), but we should fix this for Jarvis is out.
@ksooo @Jalle19

ksooo:
Case 1 is tricky. IIRC tvh 4.0 does not handle deleting of a timer scheduled by an autorec correctly, although it is possible using the webui. Former or later (latest on tvh restart), the formerly deleted child of the autorec gets automatically recreated by tvh. Not sure whether this was fixed for 4.1.

Fix for case 2 (stop recording not working) is trivial. Will come up with a PR for kodi core. Not a pvr.hts problem.

ksooo:
Just discovered that for case 2 a pvr.hts fix is needed, in addition to a kodi core fix. Will PR both fixes soon. :-(

Glenn-1990:
![schermafdruk van 2015-12-22 12 20 22](https://cloud.githubusercontent.com/assets/2036512/11954153/84b3b144-a8a6-11e5-9e66-fa6b4ebe1146.png)

Updating the active flag is not possible for tvh 4.1, we need the same fix here?

Glenn-1990:
It's not possible to disable a read only timer in the guide window, not button is showed, maybe we can add a "disable timer" button or just the "delete timer"?
[...]

PR #151, opened 2015-12-21
Fix cancel recording for timers created by autorec or timerec.

ksooo:
pvr.hts part of the fix for case 2 of issue #150

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -1060,6 +1060,22 @@ PVR_ERROR CTvheadend::DeleteTimer
     /* EPG-query-based repeating timer */
     return m_autoRecordings.SendAutorecDelete(timer);
   }
+  else if ((timer.iTimerType == TIMER_ONCE_CREATED_BY_TIMEREC) ||
+           (timer.iTimerType == TIMER_ONCE_CREATED_BY_AUTOREC))
+  {
+    /* Read-only timer created by autorec or timerec */
+    const auto &it = m_recordings.find(timer.iClientIndex);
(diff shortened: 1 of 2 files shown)


ksooo:
https://github.com/xbmc/xbmc/pull/8641: kodi core part of the fix for case 2 of issue #150

src/Tvheadend.cpp
```diff
@@ -1060,6 +1060,22 @@ PVR_ERROR CTvheadend::DeleteTimer
     /* EPG-query-based repeating timer */
     return m_autoRecordings.SendAutorecDelete(timer);
   }
+  else if ((timer.iTimerType == TIMER_ONCE_CREATED_BY_TIMEREC) ||
+           (timer.iTimerType == TIMER_ONCE_CREATED_BY_AUTOREC))
+  {
```
Glenn-1990: I wonder if " UpdateTimer" works correct as we don't use "TIMER_ONCE_CREATED_BY_XY" there anywhere. And still have to update the enabled state for tvh 4.1 (also for read only)?

ksooo: As long as timers created by autorecs and timerecs are read-only, UpdateTimer works correct, yes. Abort recording is "different" and therefore okay even if the timer is read-only.

And I made timers created by autorecs and timerecs read-only for a reason. At least with tvh 4.0, although tvh webui allows editing these timers, former or later all changes will be lost, latest after a restart of tvh. At least this was the behavior when I last checked. Don't know whether this works correctly with tvh 4.1 or was fix for recent 4.0.x

Issue #141, opened 2015-12-08
Gui bump missing 


MartijnKaijser:

ksooo:
Not hundred percent sure, but I think a simple recompile of pvr.hts (and all other addons not having explicit \<import addon="kodi.guilib" version="x.y.z"/> tag in addon.xm.in) will do the trick. AT least that fixed the incompatbilty for me. And as there was a transifex update including addon version bump yesterday this will even be fixed without any further action.

MilhouseVH:
> And as there was a transifex update including addon version bump yesterday this will even be fixed without any further action.

Yep, this now seems to be fixed ([forum post](http://forum.kodi.tv/showthread.php?tid=238393&pid=2182251#pid2182251)). Kinda feels a bit random, though - shouldn't the add-on be explicit about the version of `kodi.guilib` it requires?

ksooo:
TBH, I don't get how this is supposed to work.

On the one hand, the version can be defined in addon.xml.in, on the other hand it must be (hard)coded into addon functions like GetGUIAPIVersion.

pvr.hts does not even depend on kodi.guilib, therefore not putting a dependency into addon.xml.in seems okay, but according to https://github.com/kodi-pvr/pvr.hts/commit/254b4e29b246ee9896d3673e7dd350f7a5e411f4, one must supply an implementation for GetGUIAPIVersion and the docs do not specify a "no guilib needed" return value...

MartijnKaijser:
confusing......

PR #153, opened 2015-12-28
kodi.guilib is mandatory, enforce it

hudokkow:


ksooo:
pvr.hts does not need guilib. Why should it be mandatory? I don't think that this is the right fix for solving incompatible add-ons after guilib bumps.

The actual problem is that there is a pvr add-on API function that must return the required guilib API version and thus function does not allow to return "no guilib needed". This case was imo just forgotten when designing the function. Thus, the function must return some real version number, regardless the add-on actually needs guilib or not. And as the version gets compiled into the add-on, it must be recompiled after each guilib bump.

So, two glitches here: first, the redundancy to be able to define the guilib version in addon.xml and using the pvr add-on API function, combined with the inability of the API function tob express "no guilib needed."

I proposed to remove the pvr add-on API function in favor to the xml approach.

@FernetMenta @opdenkamp what do you think? 

@huddokow could you please close the PRs you opened to enforce guilib or make at least a comment that they should not be merged until the right solution had been found.

ksooo:
Just did a quick check against master of all pvr addons:

guilib is actually only used by 
* pvr.mediaportal.tvserver (for recording settings dialog)
* pvr.mythtv
* pvr.vdr.vnsi
* pvr.dvblink
==> guilib dependency and version needs to be defined in addon.xml

there are some senseless calls to instanciate the guilb library and GUI->RegisterMe(hdl), but nothing else in
* pvr.wmc
* pvr.nextpvr
==> Refs to guilib can be removed as guilib is actually not used

all other addons dont even instanciate guiilib library and need kodi/libKODI_guilib.h only to get KODI_GUILIB_API_VERSION & KODI_GUILIB_MIN_API_VERSION to be able to implement GetGUIAPIVersion & GetMininumGUIAPIVersion 
==> As I wrote in my last comment, I opt for removing GetGUIAPIVersion & GetMininumGUIAPIVersion for PVR addon API

opdenkamp:
Core should check which extension points are defined in addon.xml before calling `GetGUIAPIVersion()`. The reason for having the methods is that having an invalid addon.xml can crash Kodi completely.

ksooo:
> The reason for having the methods is that having an invalid addon.xml can crash Kodi completely.

 Sorry, this does not make much sense to me. ;-) It should be not to difficult to provide correct addon.xml's, imo. And if they are broken one would notice very quick due to the crashes you mentioned.

opdenkamp:
It's not difficult to provide correct ones indeed, but it's quite also easy to break things. e.g. Average Joe compiles from source, takes 'master' from both an add-on and Kodi and one hasn't been updated correctly (yet). Kodi goes pew and Joe creates a ticket on Trac because Kodi crashes. These methods prevent things like that (and yes, these things happened or I wouldn't have added those methods ;-))

Just check the extension point in core before calling the methods, remove the methods from the add-ons that don't need it and everything should be fine.

ksooo:
> Just check the extension point in core before calling the methods

Which "extension point" exactly? (Sorry, don't know what an extension point is)

> remove the methods from the add-ons that don't need it

 You mean, remove the implementation of GetGUIAPIVersion & GetMininumGUIAPIVersion from that addons completely? Will this work, can the lib be loaded, then, if the symbols are missing?

opdenkamp:
> Which "extension point" exactly? (Sorry, don't know what an extension point is)

https://github.com/kodi-pvr/pvr.hts/blob/master/pvr.hts/addon.xml.in#L13

> You mean, remove the implementation of GetGUIAPIVersion & GetMininumGUIAPIVersion from that addons completely? Will this work, can the lib be loaded, then, if the symbols are missing?

Forgot about a long term issue with binary add-ons that was still on my TODO list, which is that add-ons only support a single extension point and that you indeed can't remove those methods at this time :) To do that, you'd have to do some refactoring of AddonMgr to support multiple extension points in a single add-on and replace the universal `get_addon()` method in add-ons, splitting it up in methods for each extension point (`get_addon_pvr()`, `get_addon_gui()`, etc.). This was on my TODO list for a different reason that PVR, but the underlying problem is the same.

ksooo:
Sorry Lars, I don't get you. How does checking for "xbmc.pvrclient" help to solve the problem that even the pvr client addons that do not actually use guilib must be recompiled on every guilib version bump? That's the problem I'm trying to solve here.

If removing the functions in question is no option (I still don't buy this, btw, but you are far more experienced than I am.), the only solution that comes to my mind is to introduce a return code for "no guilib needed" for GetGUIAPIVersion & GetMininumGUIAPIVersion.

hudokkow:
@ksooo, right now it is mandatory. If we bump GUI API, addons won't work anymore without bump and recompile. I don't care what solution it is as long as https://github.com/kodi-pvr/pvr.hts/issues/141 doesn't happen again.

If the current state of the code confuses us, see https://github.com/xbmc/xbmc/pull/8494 to understand what I mean, the proverbial average Joe will go nuts because his setup suddenly stopped working.
I'll keep the PRs open until we find a solution.

ksooo:
@hudokkow 

> right now it is mandatory.

 Sorry, it is not. To prevent #141 simply change changelog and do a addon version bump. That's enough. No need to put a guilib dependency into addon.xml for this. This is just wrong.

hudokkow:
Yes, you're absolutely right about that. We are both right and wrong.

Here's what happened the last time we bumped GUI API: grep for kodi.guilib, get x results (less than one third of the addons actually), bump addons that explicitly ask for kodi.guilib. Done. We're off the hook, all is fine, flowers are beautiful, birds chirping everywhere... :wink: 
What actually happened: remaining addons stopped working for a day or two and confusion wasn't greater because translations were updated and all addons bumped.

Easily avoidable if we explicitly ask for kodi.guilib in addon.xml.in. If someone in the future greps addons for the dependency they will know all addons need bumping.

Like I wrote above, I don't care what the solution is as long as someone in the future doesn't get mislead like I did. That's it, that's all.

BTW, I do appreciate you looking into this and properly fixing it in core if you so decide. Addons shouldn't need those 2 methods if they don't use GUI.

ksooo:
@hudokkow as "next" guilib bump will not be for K**** (I guess) there is no need to hurry to merge this PRs. We should take our time to find a better solution. 

@opdenkamp would it be okay for you to extend PVR addon API for K**** by the abovementioned "no guillib needed" return value for GetGUIAPIVersion & GetMininumGUIAPIVersion? This would be just a small change and solve the problem to only need to recompile the adons actually using guilib.

opdenkamp:
> Sorry Lars, I don't get you. How does checking for "xbmc.pvrclient" help to solve the problem that even the pvr client addons that do not actually use guilib must be recompiled on every guilib version bump? That's the problem I'm trying to solve here.

The problem is that binary add-ons don't support multiple extension points, like I tried to explain in https://github.com/kodi-pvr/pvr.hts/pull/153#issuecomment-167545084. That's why right now, there are two extension points combined (pvrclient and guilib) if there's a pvrclient add-on, and that's why there's 2 similar methods now for checking API compatibility.

Bumping a single one of those to something that breaks compatiblity requires a rebuild of everything using it, and because pvr add-ons now use both the pvr and guilib APIs, they all need a bump and rebuild.

The correct way to solve this is to split up the extension points for binary add-ons properly, which requires splitting up the `get_addon()` method. Then you can get rid of those compatibility check methods for those that don't need it

opdenkamp:
oh nice, cross post :)

hudokkow:
@ksooo, absolutely. I just want to prevent the mess in the future.

ksooo:
> The correct way to solve this is to split up the extension points for binary add-ons properly, which requires splitting up the get_addon() method. Then you can get rid of those compatibility check methods for those that don't need it

 Sorry, this goes far beyond my current capabilities...

> and because pvr add-ons now use both the pvr and guilib APIs, they all need a bump and rebuild.

 which easily could be solved by extending PVR addon API for K**** by the above mentioned "no guillib needed" return value for GetGUIAPIVersion & GetMininumGUIAPIVersion. Then, only those addons actually **using** guilib would need a rebuild if guilib API gets bumped.  What do you think about this?

MartijnKaijser:
Why don't we see what @AchimTuran is doing with all the binary addon API stuff

opdenkamp:
>  What do you think about this

that hack would solve the immediate issue ;-)

ksooo:
> that hack would solve the immediate issue ;-)

 Yep. Would you block such a PR (PVR addon API change for K*****)? Just asking before hand to prevent wasting time. ;-)

opdenkamp:
nah, I haven't blocked anything for a long time, you guys are way more active than I have been for a while anyway ;-)

ksooo:
Cool, thx.

Jalle19:
Should we merge this for now until the new API is added?

ksooo:
No, I see absolutely no benefit, because next guillib bump imo will be after merging API 5.0.0 and there will be no such bump for Jarvis. OTOH we would have to remove all the lines added by this PR (and the PRs for the other addons) once API 5.0.0 is merged.

ksooo:
@hudokkow now, that a proper solution is in the works https://github.com/xbmc/xbmc/issues/8736, I'd like you to ask to close all this PRs.

hudokkow:
@ksooo, closed. Thanks for looking into this.

ksooo:
@hudokkow thx

PR #157, opened 2016-01-03
Change some log levels

Jalle19:
@Glenn-1990 can you comment on the second commit? I don't see why we should log something if it's not fatal (we don't return).


Glenn-1990:
@Jalle19 This was copied from the old subscription code.
I guess It's useful when a user says that live tv doesn't work, we can then easily spot if there is nothing wrong with his configuration on the tvheadend side (no service assigned, no free tuner, etc.).

But feel free to delete this ;-)

ksooo:
@Jalle19 Good to go. Your button.

PR #158, opened 2016-01-03
Move AsyncState to tvheadend/utilities

Jalle19:
@ksooo should be a no-brainer


ksooo:
Indeed a no-brainer. Go ahead.

PR #161, opened 2016-01-05
Factor out non-connection related stuff from CHTSPConnection

Jalle19:
Just another step in cleaning up our massive classes. I've moved out everything that's not directly related to performing "connection stuff" from `CHTSPConnection`. This includes:
- Move `CHTSPResponse` to separate files
- Holding server information in a separate instance of `ServerInformation`
- Create a simple URL helper for generating image URLs (`UrlHelper`)

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -74,32 +74,6 @@ CHTSPConnection::~CHTSPConnection()
   StopThread(0);
 }
 
-/*
- * Info
- */
-
-std::string CHTSPConnection::GetWebURL ( const char *fmt, ... )
-{
-  va_list va;
-  const Settings &settings = Settings::GetInstance();
-
-  // Generate the authentication string (user:pass@)
-  std::string auth = settings.GetUsername();
-  if (!(auth.empty() || settings.GetPassword().empty()))
-    auth += ":" + settings.GetPassword();
-  if (!auth.empty())
-    auth += "@";
-
-  std::string url = StringUtils::Format("http://%s%s:%d", auth.c_str(), settings.GetHostname().c_str(), settings.GetPortHTTP());
-
-  va_start(va, fmt);
-  url += m_serverInformation.GetWebRoot();
-  url += StringUtils::FormatV(fmt, va);
-  va_end(va);
-
-  return url;
-}
-
 bool CHTSPConnection::WaitForConnection ( void )
 {
   if (!m_ready) {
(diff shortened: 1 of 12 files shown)


ksooo:
Nice.

src/HTSPConnection.cpp
```diff
@@ -407,39 +408,51 @@ bool CHTSPConnection::SendHello ( void )
...
   /* Basic Info */
   webroot = htsmsg_get_str(msg, "webroot");
-  m_serverName    = htsmsg_get_str(msg, "servername");
-  m_serverVersion = htsmsg_get_str(msg, "serverversion");
-  m_htspVersion   = htsmsg_get_u32_or_default(msg, "htspversion", 0);
-  m_webRoot       = webroot ? webroot : "";
+  serverName = htsmsg_get_str(msg, "servername");
+  serverVersion = htsmsg_get_str(msg, "serverversion");
+  htspVersion = htsmsg_get_u32_or_default(msg, "htspversion", 0);
+  webroot = webroot ? webroot : "";
```
Jalle19: Note to self that the ternary here should be removed since we check for NULL before storing, and `ServerInformation::m_webRoot` defaults to an empty string anyway.

src/HTSPConnection.cpp
```diff
@@ -407,39 +408,51 @@ bool CHTSPConnection::SendHello ( void )
...
+  if (serverVersion)
+    m_serverInformation.SetServerVersion(serverVersion);
+  if (webroot)
+    m_serverInformation.SetWebRoot(webroot);
+
+  m_serverInformation.SetHtspVersion(htspVersion);
+
   Logger::Log(LogLevel::DEBUG, "connected to %s / %s (HTSPv%d)",
-            m_serverName.c_str(), m_serverVersion.c_str(), m_htspVersion);
+            m_serverInformation.GetServerName().c_str(),
```
Jalle19, line 352: Another note to self to fix this whitespace alignment

Jalle19:
@ksooo are you okay with the latest commit? I would have liked to use a "read-write lock" ala `std::shared_mutex` but it seems it's only available on C++14 (some sources say C++17).

Jalle19:
@ksooo okay to merge?

Jalle19:
I've updated the PR title and description.

@ksooo using the image cache doesn't work on my server for some reason (tvheadend itself fails to fill the cache for some reason), are you able to test that? I've verified that the URLs look correct though so it shouldn't be a problem.

Jalle19:
@ksooo I've decided to tack on additional small changes here since you haven't responded ;-)

Jalle19:
The image cache problem I mentioned has been reported upstream so no need to worry about that for now: https://tvheadend.org/issues/3548

ksooo:
Please reopen for Matrix if you feel this is still of interest.

PR #163, opened 2016-01-07
Fix crash when trace logging events 

Jalle19:
Since we moved to the new logger utility everything will crash if a logged string contains a `%` character (because the list of expected variables increases and `vsprintf` tries to read too far).


Jalle19:
Found this by chance since I had an event named `99% norsk`. @ksooo are you okay with this approach? I would have liked to do the escaping transparently in `Logger::Log` but I don't think it's possible to manipulate variable arguments (you can loop through them but you can't modify them).

Jalle19:
Applied the same logic to logged recordings/timers.

Jalle19:
Thanks to @adamsutton and @perexg for the proper solution!

PR #165, opened 2016-01-09
Run automatic build tests on Travis CI

Jalle19:
Let's see if this works...


Jalle19:
@ksooo since I recently managed to break the build I thought this can't hurt. It's only for Linux but it's better than nothing IMO.

Inspired by @adamsutton who added the same thing for tvheadend.

ksooo:
Yeah. Thumbs up.

PR #166, opened 2016-01-11
fixed crash in logger on windows.

afedchin:
as discussed in slack


Jalle19:
@ksooo for reference, apparently you can't pass references to `va_start` like we did (we passed a `const std::string &message`) since it's a macro and operates on the memory address of the parameter (or something). The previous code magically worked on Linux but it failed during runtime on Windows. Changing the parameter to a pointer fixes it.

PR #167, opened 2016-01-11
Factor out CHTSPResponse to a new tvheadend::htsp namespace

Jalle19:
@ksooo fairly straight-forward this one aswell.

### DIFF
--- a/src/tvheadend/htsp/Response.cpp
+++ b/src/tvheadend/htsp/Response.cpp
@@ -1,0 +1,53 @@
+/*
+ *      Copyright (C) 2005-2014 Team XBMC
+ *      http://www.xbmc.org
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


src/tvheadend/htsp/Response.cpp
```diff
@@ -0,0 +1,53 @@
...
+  m_msg = msg;
+  m_flag = true;
+  m_cond.Broadcast();
+}
+
+htsmsg_t* Response::Get(PLATFORM::CMutex &mutex, uint32_t timeout)
+{
+  m_cond.Wait(mutex, m_flag, timeout);
+  htsmsg_t *r = m_msg;
+  m_msg = NULL;
```
ksooo: Not really important, but what do you think about using nullptr instead of NULL? Shouldn't we go for nullptr nowadays?

Jalle19: Sure

Jalle19:
Closing in favor of #161

PR #168, opened 2016-01-11
Validate event before adding timer

Glenn-1990:
This one is bothering me since ages.

Let me explain the problem first:
Sometimes when I add a epg based timer, pvr.hts return a server error.
The log files shows that tvheadend returned "Event does not exist"
I can resolve this problem by clearing the tv guide data and reload it.

Now the problem is that the epg handling in pvr.hts is synchronous by default.
So when an event gets updated/deleted/added on the backend, kodi will not be aware of this.
Kodi will refetch its epg data on intervals of minimum 15 minutes (user sets this in tv settings), for low power systems, kodi will not fetch any epg data when video is playing (by default)...

These problems will increase the risk that the kodi epg data isn't up to date anymore with the backend and this may result in  "Event does not exist".

The problem seems more common on now/next epg events, this is probably due to the fact that some providers do real time adjustments for them.

Possible solutions I found are:
1) switch epg to async by default (some major disadvantages)
2) fallback to a manual timer in case an event doesn't exists anymore

I implemented solution 2 in this PR, I know it's not perfect, but I can't come up with a better solutions.
If someone has a better idea, please let me know ;-)

@Jalle19 @ksooo

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -180,6 +180,44 @@ std::string CTvheadend::GetStreamingProfile() const
   return streamingProfile;
 }
 
+bool CTvheadend::IsEventValid(uint32_t eventId, uint32_t channelId)
+{
+  uint32_t id, channel;
+  htsmsg_t *m;
+
+  if (eventId <= PVR_TIMER_NO_EPG_UID)
+    return false;
+
+  /* Mismatch should not happen with async epg */
+  if (Settings::GetInstance().GetAsyncEpg())
+    return true;
+
+  /* Request event from server */
+  m = htsmsg_create_map();
+  htsmsg_add_u32(m, "eventId", eventId);
+
+  /* Send and Wait */
+  {
+    CLockObject lock(m_conn.Mutex());
+    m = m_conn.SendAndWait("getEvent", m);
+  }
+
+  /* Event not found on server? */
+  if (!m)
+    return false;
+
+  /* Check response */
+  htsmsg_get_u32(m, "eventId",   &id);
+  htsmsg_get_u32(m, "channelId", &channel);
+  htsmsg_destroy(m);
+
+  /* Id or channel mismatch */
+  if (id != eventId || channel != channelId)
+    return false;
+
+  return true;
+}
+
 /* **************************************************************************
  * Tags
  * *************************************************************************/
@@ -968,10 +1006,21 @@ PVR_ERROR CTvheadend::AddTimer ( const PVR_TIMER &timer )
     /* one shot timer */
 
     uint32_t u32;
+    bool bEventValid;
+
+    if (timer.iTimerType == TIMER_ONCE_EPG)
+    {
+      /* Kodi refreshes it's guide data with an interval of minimum 15 minutes */
+      /* So there is a chance that the iEpgUid isn't valid anymore on the backend */
+      /* If so, fallback to an manual timer */
+      bEventValid = IsEventValid(timer.iEpgUid, timer.iClientChannelUid);
+      if (!bEventValid)
+        Logger::Log(LogLevel::LEVEL_INFO, "Event %i isn't valid anymore, fallback to a manual timer", timer.iEpgUid);
+    }
 
     /* Build message */
     htsmsg_t *m = htsmsg_create_map();
-    if (timer.iEpgUid > PVR_TIMER_NO_EPG_UID && timer.iTimerType == TIMER_ONCE_EPG)
+    if (timer.iTimerType == TIMER_ONCE_EPG && bEventValid)
     {
       /* EPG-based timer */
       htsmsg_add_u32(m, "eventId",      timer.iEpgUid);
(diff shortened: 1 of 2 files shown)


ksooo:
If this only happens with synchronous "epg data transfer", why don't we just remove that option and related code? I was asking myself for a longer time why we would need an optional behaviour for this.

ksooo:
And, btw, what are the major disadvantages of async epg data transfer?

ksooo:
Last, I really do not like solution 2. I consider this a hack which a normal user will not understand, I'm afraid.

Glenn-1990:
@ksooo 
Disadvantages where I'm aware of:
- we have to chache lots of data, 31 days of epg events for all channels, that's quite an amount.
This will slow down the epg import as we do not take the "days to display" setting in kodi into account.
Or is there a way to read a kodi setting from pvr.hts?
-  There will be a continuous stream of data updates from the backend to pvr.hts (more etwork traffic) and this will continue trigger PVR->TriggerEpgUpdate(). And with that call we bypass the "prevent updates during playback" (needed for some low power systems).
Also PVR->TriggerEpgUpdate() will update all data for the specified channel, so a lot more than needed.

Why does the user need to understand? :p He will probably not notice that a manual timer was added :-)

Jalle19:
The API would need some changes to really make the async EPG feature work as intended.

ksooo:
Thanks for explaining the disadvantages.

> Why does the user need to understand? :p He will probably not notice that a manual timer was added :-)

-1

ksooo:
> The API would need some changes to really make the async EPG feature work as intended.

 @Jalle19 what exactly is missing? As you know I'm currently collecting stuff for PVR API 5.0.0

Jalle19:
We'd need some method to push both deleted and updated/new events to Kodi. Something like `PVR->PushEpgEvent(uint32_t eventId, PVR_EPG_EVENT_STATE state)` (where `PVR_EPG_EVENT_STATE` is an enum with values `CREATED`, `UPDATED`, `DELETED`).

When this method is called, Kodi would check the state and do the appropriate action on the event.

This is just a suggestion, I haven't had time to think that much about it.

ksooo:
Yeah, imo that's the right way to go. If I find the time I will come up with a PR for this.

Jalle19:
That method signature probably needs a `uint32_t channelId` too.

Glenn-1990:
Also tvh seems to need some patches as it doesn't take "epgMaxTime" into count.
@ksooo would it be possible to read the "epg days setting" from kodi? otherwise we need to load much more data than needed.

Jalle19:
@Glenn-1990 the loading of data from tvheadend to the addon is not really an issue, it's much faster than sending the data from the addon to Kodi anyway. What you see from a user perspective is that the first channel's EPG data takes a while to load (since it blocks until we have received all the events from the server), after that the rest load very quickly since no round-trip to the server is required. The total time spent on fetching guide data is shorter with asynchronous EPG enabled than disabled (at least in my experience, I haven't tried on slow hardware).

ksooo:
> would it be possible to read the "epg days setting" from kodi?

 Sorry, don't know.

Glenn-1990:
@Jalle19 
Yes, I'm aware that the epg receival is blocked until all async transfers are done and that this should not introduce speed issues when retrieval the same amount of data as with synchronous.

I'll explain my problem a bit more:
I'm not at home for a few weeks so I'm testing on my laptop but my server is at home, so the bandwidth is limited in  my case (by upload speed of ISP). That's why I set the epg days option in kodi to "1 day" With this setting my epg import time is about 40 seconds. Now when I switch to async mode, the epg import time will be almost 3 minutes. This because all events are grabbed instead of that "1 day option". When I'm home I'll test the performance in the local network. Ideally we should take that "1 day" into account for async mode too. But tvh doesn't handle that well.

40s

Glenn-1990:
@ksooo, @Jalle19 

Attempt to make async epg more attractive for low end systems with pvr.hts and tvheadend changes.

tvheadend:
- "epgMaxTime" is now correctly taken into account and is updatable.
- This also means that "epgMaxTime" also moves with the current time (instead of being static).
https://github.com/Glenn-1990/tvheadend/commit/fbcba783627979e33002ef87b212319a0ed920c3

pvr.hts:
- Use "epgMaxTime" now, so that we don't load more events than necessary (faster initial load). 
- Don't trigger kodi updates for passed and updated events to save cpu cycles. Updates are not critical as the id will stay the same and the update will be picked up by kodi's GetEpg() call eventually. 
- Once kodi support epg triggers for single events, these checks are not necessary anymore, but they are ment for my jarvis branch.
https://github.com/Glenn-1990/pvr.hts/commit/4fea355d76ea60708f550810ea5367c379a05bab

Please let me know what you guys think of it :-)

ksooo:
Regarding epg triggers for single events you may want to take a look at my api 5.0.0 wip branch: https://github.com/ksooo/pvr.hts/commits/pvr-api-5-0-0

ksooo:
@Glenn-1990 I guess this PR can be closed now that we have true async epg data transfer (with PVR API  5.1.0)?

Glenn-1990:
indeed

Issue #106, opened 2015-10-07
Adapt to latest descrambler info API changes


Jalle19:
See https://tvheadend.org/issues/2616 for details

ksooo:
Just a side note: THis was added in HTSP v24

Jalle19:
It's an old Enigms thing so old it's from the time when cardsharing was so awesome you had to see it in action on every zap.

ksooo:
hehe

Jalle19:
Even thought all the fancy stuff about ECM times are unnecessary before Kodi has a way to display them in the UI we still need to implement this new method to be able to properly show which CAID is being used (since HTSP v24 the CAID is always set to 0x0000 in `subscriptionStart`, clients are supposed to listen for this new message instead).

I'll try to come up with something next week. Will have to set up a complete test environment because I'm not touching my main tvheadend build.

belliash:
Any progress here?

Jalle19:
Sorry not yet, haven't had time to code for a while.

Jalle19:
@perexg I tried to start implementing this today but I can't get tvheadend to actually send any `descramblerInfo` HTSP messages. I'm trying the latest master in a virtual machine with a mux URL from my main tvheadend instance. Oscam is connected using DVBAPI and descrambling works. Please advice.

perexg:
Do you see ECM info in the tvh subscriptions tab ?

Jalle19:
I have oscam r10500 something so it should be new enough. I didn't check the subscription tab, will do later and report back.

perexg:
I think that the network protocol v2 was added around 10700 or so.. You should upgrade oscam.

Jalle19:
It said protocol version 2 when I connected, but I guess I'll upgrade then.

Jalle19:
I finished the part of the code where the status is parsed and stored in the demuxer, but we have now hit a road block. Kodi needs a new API method for retrieving all this data during streaming instead of during channel import since the CAID is not available when fetching the list of channels. I'll have to think about the best way to add this to the API.

belliash:
When you press "O" in Kodi, it displays information about channel priovider, frequency, etc... There is also information about CAID (actually it is FFFF). This information if fetched from TVH on channel playback, not during importing channel list. After changes made by perexg, I believe TVH should pass information about current CAID to clients.

Jalle19:
[...]

PR #169, opened 2016-01-11
Parse, store and log descrambler info

Jalle19:
Partially closes https://github.com/kodi-pvr/pvr.hts/issues/106. The API will need some changes in order to make this information useful (for now it's just stored, not passed along to Kodi in any way).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ source_group("Resource Files" FILES ${HTS_RESOURCES})
 list(APPEND HTS_SOURCES
             ${HTS_SOURCES_TVHEADEND}
             ${HTS_SOURCES_TVHEADEND_ENTITY}
+            ${HTS_SOURCES_TVHEADEND_STATUS}
             ${HTS_SOURCES_TVHEADEND_UTILITIES}
             ${HTS_RESOURCES})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,8 @@ set(HTS_SOURCES_TVHEADEND_ENTITY
                 src/tvheadend/entity/TimeRecording.cpp)
 
 set(HTS_SOURCES_TVHEADEND_STATUS
+                src/tvheadend/status/DescrambleInfo.h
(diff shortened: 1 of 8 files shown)


ksooo:
> The API will need some changes in order to make this information useful

 I will be happy to include this change in our PVR Addon API 5.0.0 PR if you point me to a proper commit that I can cherry-pick. ;-)

Jalle19:
Scribbling some stuff here for future reference:

Addons will need a way to publish descrambler information to Kodi. Kodi should then store this information in `CPVRGUInfo` where the skin can grab it. Kodi should clear the stored information when changing channels or when stopping playback to prevent stale information from being displayed.

Since most existing API methods are "pull" methods (i.e. Kodi pulls the information from the addon) this one should probably follow that convention. This way Kodi can also control whether to fetch the information at all.

The method signature could be something like this: `PVR_ERROR GetDescramblerInfo(PVR_DESCRAMBLER_INFO *info)` where `PVR_DESCRAMBLER_INFO` is a struct. This struct could probably be quite close to the `descrambleInfo` message tvheadend sends to clients since that contains pretty much every piece of information a descrambler could possibly provide.

While we're at it I'd like to also implement an API method for retrieving e.g. how many tuners the server has, which tuners are used at the moment, what is the delivery system used for the current channel (DVB-T/C/S etc.) and so on. There are some skins (notable Aeon Nox master) that technically support all this data, there's just no way to get them at the moment.

ksooo:
> While we're at it I'd like to also implement an API method for retrieving e.g. how many tuners the server has, which tuners are used at the moment, what is the delivery system used for the current channel (DVB-T/C/S etc.) and so on. 

 Nice

Jalle19:
I actually started work on the latter suggestion in a branch, I'll try to revive it at some point. How much in a hurry are you to get the 5.0.0 API bump in?

ksooo:
No hurry at all. Just collecting stuff atm. Want to avoid multiple API bumps for Krypton, if possible.

ksooo:
This one can be closed, I cherry-picked the main commit for #315

PR #174, opened 2016-01-20
Minor tag tweaks

Glenn-1990:
I discovered some strange things when investigation duplicate database entries, this might not fix anything, but this will definitely more efficient.

@ksooo @Jalle19


Glenn-1990:
Yes, you are right, it's not save to do this.

Jalle19:
@Glenn-1990 should we close this one?

PR #178, opened 2016-01-25
Extend automatic build tests to run on clang and OS X too

Jalle19:
@ksooo funny how only OS X complains about the missing `IsRealTimeStream` symbol. I haven't had time to implement that part yet.

@MartijnKaijser @AlwinEsch the Travis configuration is now about as complete as it can get, feel free to adapt these to other binary addons.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,22 @@
-#
-# Travis defaults to building on Ubuntu Precise. We need Trusty
-# in order to get up to date versions of cmake and g++
-#
 language: cpp
-sudo: required
-dist: trusty
+
+#
+# Define the build matrix
+#
+# Travis defaults to building on Ubuntu Precise when building on 
+# Linux. We need Trusty in order to get up to date versions of 
+# cmake and g++.
+#
+matrix:
+  include:
+    - os: linux
+      dist: trusty    
+      sudo: required
+      compiler: gcc
+    - os: linux
+      dist: trusty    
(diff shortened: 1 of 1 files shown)


Jalle19:
@Memphiz do you think it would be beneficial to build using other Xcode versions as well?

Memphiz:
jenkins builds with xcode6.1 atm - what do you mean? Thats where it should build for sure.

Jalle19:
@Memphiz you pretty much answered my question. The main goal of using Travis here is to catch potential build issues before they make it to master and breaks the addon.

I'll add at least 6.1 to the build matrix then.

Memphiz:
Yep - after jarvis is out of the door i will bump to at least xcode7.2 (which is already installed side by side) ...

Memphiz:
Also you need to ensure to use the same deployment target as kodi uses (atm osx 10.7 and ios 5.1) - else compile results might differ ...

Jalle19:
@Memphiz I'm not sure if it's possible to control that on Travis.

MartijnKaijser:
@Memphiz what happend to the mythicbeast slave? Wasn't it planned for upgrade?

Jalle19:
Added xcode6.1 as a build target.

Memphiz:
@martijnkaijser - its updated to osx 10.7 - but its not a jenkins slave atm - i used it for checking 10.7 runtime with kodi...

PR #179, opened 2016-01-27
Updated to PVR API v4.2.0 (implemented IsRealTimeStream)

ksooo:
Implementation should basically do the same as pvr.vdr.vnsi: https://github.com/kodi-pvr/pvr.vdr.vnsi/commit/3b94c58114f93c42e38741fb1dd403858bfe04b2#diff-34f48ad91ac6c202ac60b0348ae90e30R862

@Jalle19 good to go?

### DIFF
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -231,6 +231,17 @@ void CHTSPDemuxer::SetStreamingProfile(const std::string &profile)
   m_subscription.SetProfile(profile);
 }
 
+bool CHTSPDemuxer::IsRealTimeStream() const
+{
+  if (GetTimeshiftTime() == 0)
+    return true;
+
+  if (GetTimeshiftBufferEnd() - GetTimeshiftTime() < 10)
+    return true;
+
+  return false;
+}
+
 /* **************************************************************************
  * Parse incoming data
  * *************************************************************************/
--- a/src/Tvheadend.h
+++ b/src/Tvheadend.h
@@ -265,6 +265,7 @@ public:
       return m_lastUse;
     return 0;
   }
+  bool IsRealTimeStream() const;
 
   /**
    * Tells each demuxer to use the specified profile for new subscriptions
@@ -273,9 +274,9 @@ public:
   void SetStreamingProfile(const std::string &profile);
 
 private:
-  P8PLATFORM::CMutex                        m_mutex;
+  P8PLATFORM::CMutex                      m_mutex;
   CHTSPConnection                        &m_conn;
-  P8PLATFORM::SyncedBuffer<DemuxPacket*>    m_pktBuffer;
+  P8PLATFORM::SyncedBuffer<DemuxPacket*>  m_pktBuffer;
   ADDON::XbmcStreamProperties             m_streams;
   std::map<int,int>                       m_streamStat;
   int64_t                                 m_seekTime;
@@ -568,6 +569,7 @@ public:
   int64_t      DemuxGetTimeshiftTime() const;
   int64_t      DemuxGetTimeshiftBufferStart() const;
   int64_t      DemuxGetTimeshiftBufferEnd() const;
+  bool         DemuxIsRealTimeStream() const;
 
   /*
    * VFS (pass-thru)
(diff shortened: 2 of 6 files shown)


src/client.cpp
```diff
@@ -301,6 +301,39 @@ bool IsTimeshifting(void)
...
 }
 
+static time_t ConvertMusecsToTime(int64_t musecs)
+{
+  // tvheadend reports microseconds for timeshifting values,
+  // Kodi expects it to be an absolute UNIX timestamp
+  return time(NULL) - static_cast<time_t>(static_cast<double>(musecs / 1000000));
+}
+
+time_t GetPlayingTime()
```
FernetMenta: this does not look thread safe. at least I don't see it

Jalle19: What do you mean? The code has simply been moved, it's been like this for quite some time.

FernetMenta: I did not say that it broke but that it does not look right. I have never looked into this before. Those methods here are called be gui but as far as I can see they go unprotected into demuxer.

ksooo: @FernetMenta you are right. Not introduced by this PR (at least not for the "old" functions), but should be fixed. Will do.

ksooo:
@Jalle19 looked almost first time into class CHTSPDemuxer. Is there any logic when member access is to protect by m_mutex? Cannot make any sense out of what I'm seeing there.

Jalle19:
There is some logic but I think some parts need some work. All the parse methods are called with mutex protection from ProcessMessage(). m_subscription is thread-safe. It's the rest I'm not really sure about.

ksooo:
@FernetMenta  Okay now? (At least the stuff touched by this PR?)

FernetMenta:
@ksooo not sure. where is the protection?

ksooo:
@FernetMenta => https://github.com/ksooo/pvr.hts/commit/a300442439a3df6fe565d5d64f63dac61a4de977#diff-fc67c1aab2e0d4a225199143a1d4be69R236

FernetMenta:
This does not help. GetBufferTimeStart is called by gui thread and accesses an unprotected parameter.

ksooo:
Sure. Sigh. I have the feeling I have to fix more and more stuff others implemented in a non-ideal way... m_dmx_active is accessed without protection "everywhere".

FernetMenta:
what is the story behind m_dxm_active? there's never more than a single demuxer at a given time.

Jalle19:
@FernetMenta yes there are, the active one points to the one that should be sending to Kodi.

@ksooo it should be enough to add mutex locks tp the various Get() methods in the demuxer.

FernetMenta:
@Jalle19 are you playing with demuxers just for fun? kodi never opens more than a single demuxer.

Jalle19:
@FernetMenta it's for an optional feature called predictive tuning.

Jalle19:
@ksooo I'm taking some notes and looking through the demuxer code right now, okay if I handle the threading issues?

ksooo:
Sure. Thx. So, okay to merge this PR now and to address the trhreading stuff in your followup-up PR?

Jalle19:
Yes, let's do that.

src/HTSPDemuxer.cpp
```diff
@@ -231,6 +231,19 @@ void CHTSPDemuxer::SetStreamingProfile(const std::string &profile)
   m_subscription.SetProfile(profile);
 }
 
+bool CHTSPDemuxer::IsRealTimeStream() const
+{
+  CLockObject lock(m_mutex);
```
Jalle19: @ksooo can you drop this before merging?

src/Tvheadend.h
```diff
@@ -273,9 +274,9 @@ class CHTSPDemuxer
   void SetStreamingProfile(const std::string &profile);
 
 private:
-  P8PLATFORM::CMutex                        m_mutex;
+  mutable P8PLATFORM::CMutex              m_mutex;
```
Jalle19: If you drop the lock you can drop this as well, I'll be re-introducing it in my PR.

ksooo: sure, will drop both.

PR #155, opened 2015-12-30
Timeshift: Seeking fixes

perexg:
Cherry picked changes from PR #146 . I've not tested compilation on my dev platform (using 15.2), so little tweaking may be required.

Note: Test with TVH v4.1-1264-g0f069c7 and later.
Note2: Another bunch of timeshift fixes TVH v4.1-1276-ga27ed2e (mainly wrong refcounting and reworked on-demand mode).
Note3: TVH v4.1-1299-g80a50ce seems to catch all major timeshift bugs.

PR #180, opened 2016-01-28
Fix some potential threading issues in the demuxer

Jalle19:
@adamsutton are you able to comment on whether https://github.com/kodi-pvr/pvr.hts/commit/e61ad9d5aec5e3582767d14452654b31a06de00d looks correct?

### DIFF
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -167,7 +167,9 @@ bool CHTSPDemuxer::Seek
   }
 
   /* Wait for time */
+  CLockObject lock(m_conn.Mutex());
   m_seekTime = 0;
+
   if (!m_seekCond.Wait(m_conn.Mutex(), m_seekTime, Settings::GetInstance().GetResponseTimeout()))
   {
     Logger::Log(LogLevel::LEVEL_ERROR, "failed to get subscriptionSeek response");
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -53,9 +53,7 @@ void CHTSPDemuxer::Connected ( void )
     m_subscription.SendSubscribe(0, 0, true);
     m_subscription.SendSpeed(0, true);
 
-    /* Reset status */
-    m_signalInfo.Clear();
-    m_sourceInfo.Clear();
+    ResetStatus();
   }
 }
 
@@ -96,8 +94,7 @@ bool CHTSPDemuxer::Open ( uint32_t channelId, enum eSubscriptionWeight weight )
   m_subscription.SendSubscribe(channelId, weight);
   
   /* Reset status */
-  m_signalInfo.Clear();
-  m_sourceInfo.Clear();
+  ResetStatus();
 
   /* Send unsubscribe if subscribing failed */
   if (!m_subscription.IsActive())
@@ -254,6 +251,14 @@ bool CHTSPDemuxer::IsRealTimeStream() const
     return true;
 
   return false;
+}
+
+void CHTSPDemuxer::ResetStatus()
+{
+  CLockObject lock(m_mutex);
+
+  m_signalInfo.Clear();
+  m_sourceInfo.Clear();
 }
 
 /* **************************************************************************
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -288,10 +288,14 @@ void CHTSPDemuxer::SetStreamingProfile(const std::string &profile)
 
 bool CHTSPDemuxer::IsRealTimeStream() const
 {
-  if (GetTimeshiftTime() == 0)
+  /* Avoid using the getters since they lock individually and
+   * we want the calculation to be consistent */
+  CLockObject lock(m_mutex);
+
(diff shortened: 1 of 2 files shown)


adamsutton:
@Jalle19 yes, I'm surprised that worked at all! Unless I'm missing a lock somewhere buried in the lib platform code (unlikely). Or there is a lock outside the Seek() call I'm missing. This just shouldn't work at all!

The mutex passed to cond_wait() MUST be locked before entry, and will be re-locked upon exit.

Jalle19:
> The mutex passed to cond_wait() MUST be locked before entry, and will be re-locked upon exit.

So it's not enough to lock it before `Wait()` like I do now, I'd also need to lock it afterwards when the value of `m_seekTime` is being read and transferred to `*startpts`? I'm beginning to think this is why people tend to get completely bogus seek times when they skip too quickly (`ParseSubscriptionSeek()` seems to have been able to overwrite the value of `m_seekTime` while the value is being read into `*startpts`).

adamsutton:
No, your code is correct. the sequence is like this:

1. lock mutex
2. call cond_wait()
2a. cond_wait will atomically unlock the mutex and wait for signal
2b. signal is received (or timeout) cond_wait will lock mutex
3. do work
4. unlock mutex

You're using a CLockObject() which is a helper C++'ism which will lock the mutex on creation and then auto-unlock on deletion (i.e. when it goes out of scope, in this case when the function returns).

But I think you also need to look at the registration (conn setup) condition, it seems to have the same flaw.

What I want to know is WTF was I thinking! and why doesn't this generate errors!

FernetMenta:
btw: there is no need to protect entire demuxer. only videoplayer thread calls demuxer methods. you only need to protect the timeshift parameters queried by gui.
vnsi demuxer is not protected. it only grabs a lock when writing timeshift values into a cache.

Jalle19:
@FernetMenta we have two separate mutexes, one which the demuxer owns and one which the connection owns. Some things need to be locked by one, some by the other.

Jalle19:
@ksooo rebased, let's see what Travis has to say.

Jalle19:
Dropped the first commit since it was unrelated.

src/HTSPDemuxer.cpp
```diff
@@ -226,6 +226,47 @@ PVR_ERROR CHTSPDemuxer::CurrentSignal ( PVR_SIGNAL_STATUS &sig )
   return PVR_ERROR_NO_ERROR;
 }
 
+int64_t CHTSPDemuxer::GetTimeshiftTime() const
```
FernetMenta: calling IsRealTimeStream ends up in grabbing the lock tree times. the results of the subtraction might be wrong too

````
bool CHTSPDemuxer::IsRealTimeStream() const
{
  if (GetTimeshiftTime() == 0)
    return true;

  if (GetTimeshiftBufferEnd() - GetTimeshiftTime() < 10)
    return true;

  return false;
}
```

Jalle19: True, but it shouldn't matter since I doubt the mutex will be locked for more than a split second. We could move the whole algorithm to a method inside CHTSPDemuxer though, then we could lock just once without having to make the mutex public (something I really want to avoid).

Jalle19:
@perexg requested #155 to be merged before this since it fixes another flaw in how seek time is handled. @ksooo any objections?

ksooo:
no ojections. except that travis-ci failed?

src/HTSPDemuxer.cpp
ksooo: really? I don't like this. really. because it is not correct.

ksooo: what's the problem with locking the mutex once on entry of CHTSPDemuxer::IsRealTimeStream()? Is it to expensive to reacquire the already locked mutex when entering GetTimeshiftTime() and GetTimeshiftBufferEnd()?

FernetMenta: in vnsi I only lock the cache because I don't want the gui to influence demuxer in any way. you don't want the gui lock out the demuxer a couple of times each cycle.

Jalle19:
Travis failed due to a network issue with one of the mirrors. I'll add a second commit before merging this.

Jalle19:
@ksooo @FernetMenta please see the latest commit. Are you okay with this?

src/HTSPDemuxer.cpp
```diff
@@ -247,15 +288,27 @@ void CHTSPDemuxer::SetStreamingProfile(const std::string &profile)
...
-  if (GetTimeshiftTime() == 0)
+  /* Avoid using the getters since they lock individually and
+   * we want the calculation to be consistent */
+  CLockObject lock(m_mutex);
+
+  if (m_timeshiftStatus.shift == 0)
     return true;
 
-  if (GetTimeshiftBufferEnd() - GetTimeshiftTime() < 10)
+  if (m_timeshiftStatus.start - m_timeshiftStatus.shift < 10)
```
ksooo: yeah that is exactly what I had in mind to do it "right". thx.

ksooo:
go head, shove it in.

Issue #188, opened 2016-01-31
New problems with timeshift in latest master/Jarvis


ksooo:
Using pvr.hts Jarvis with latest timeshift fixes from @perexg and unpatched tvh 4.0.8. Running now in several new trouble when seeking (timeshifting) while watching Live TV. Always getting log entries like:
 
11:35:25 T:140537336149760   ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - failed to get subscriptionSeek response
11:35:25 T:140537336149760 WARNING: error while seeking

 and picture freezes. Have to stop and restart the channel to fix.

 Most of the bugs are difficult to reproduce. But one is 100% reproducable. While watching Live TV change audio track (which in fact does a seek). 

@perexg would be great if you could investigate into this.

ksooo:
... or is the cause https://github.com/kodi-pvr/pvr.hts/pull/180/files#diff-fc67c1aab2e0d4a225199143a1d4be69R170 ? Just guessing, because it is around the failing htsp method call.

ksooo:
No, removing the mutex before "subscriptionseek" call does not help.

ksooo:
... but reverting #155 fixes my problems. :-/

Jalle19:
How long does it take after you trigger the seek until you get the "subscriptionSeek failed" message? Does it look like the `Wait()` call returns immediately?

Jalle19:
Think I see the problem, must be a merge issue when I did the Jarvis PR. Stand by...

Jalle19:
@ksooo please try https://github.com/kodi-pvr/pvr.hts/pull/189

ksooo:
Nope. As I said I tried with ad without that mutex. Makes no difference. Problem is definitely caused by by @perexg timeshift PR. If I completely revert #155 my problems are solved.

ksooo:
Here are two log snippets (in one file), first for working audio track switch with perexg changes reverted, second for failing track switch (freeze) with perexg changes applied. http://paste.ubuntu.com/14800426/

perexg:
@ksoo : could you provide debug for pvr.hts for both working / non-working code. But I guess that 4.0.8 has wrong replies to the seek request - it was ignored before.

Jalle19:
I think I'm still on 4.0.7, I'll try again later. Wasn't able to reproduce this with #189 though.

ksooo:
[...]

PR #189, opened 2016-01-31
Fix merge error which causes subscriptionSeek to fail

Jalle19:


Jalle19:
@ksooo haven't tried without this patch but at least with it timeshift works properly

ksooo:
Not related to #188, but definitely okay.

PR #191, opened 2016-01-31
Remove superfluous "attempting reconnect" log message

Jalle19:
It will be triggered during shutdown when it's obvious there will be no
reconnect attempt made.

@ksooo ?


ksooo:
+1

PR #195, opened 2016-02-07
Implement PVR API 5.0.0

ksooo:
Implementation of PVR Addon API 5.0.0 as defined here https://github.com/xbmc/xbmc/pull/8736

Note: SetEPGTimeFrae currently is a stub only, to be implemented later. Other stuff is fully implemented and runtime-tested for some weeks now.

@Jalle19 Your opinion?

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -93,7 +93,8 @@ void *CHTSPRegister::Process ( void )
 CHTSPConnection::CHTSPConnection ()
   : m_socket(NULL), m_regThread(this), m_ready(false), m_seq(0),
     m_serverName(""), m_serverVersion(""), m_htspVersion(0),
-    m_webRoot(""), m_challenge(NULL), m_challengeLen(0), m_suspended(false)
+    m_webRoot(""), m_challenge(NULL), m_challengeLen(0), m_suspended(false),
+    m_state(PVR_CONNECTION_STATE_UNKNOWN)
 {
 }
 
@@ -108,7 +109,7 @@ CHTSPConnection::~CHTSPConnection()
  * Info
  */
 
-std::string CHTSPConnection::GetWebURL ( const char *fmt, ... )
+std::string CHTSPConnection::GetWebURL ( const char *fmt, ... ) const
 {
   va_list va;
   const Settings &settings = Settings::GetInstance();
@@ -133,32 +134,38 @@ std::string CHTSPConnection::GetWebURL ( const char *fmt, ... )
 
 bool CHTSPConnection::WaitForConnection ( void )
 {
-  if (!m_ready) {
+  if (!m_ready)
+  {
     Logger::Log(LogLevel::LEVEL_TRACE, "waiting for registration...");
     m_regCond.Wait(m_mutex, m_ready, Settings::GetInstance().GetConnectTimeout());
   }
   return m_ready;
 }
 
-std::string CHTSPConnection::GetServerName ( void )
+int  CHTSPConnection::GetProtocol ( void ) const
+{
+  CLockObject lock(m_mutex);
+  return m_htspVersion;
+}
+
+std::string CHTSPConnection::GetServerName ( void ) const
 {
   CLockObject lock(m_mutex);
   return m_serverName;
 }
 
-std::string CHTSPConnection::GetServerVersion ( void )
+std::string CHTSPConnection::GetServerVersion ( void ) const
 {
   CLockObject lock(m_mutex);
   return StringUtils::Format("%s (HTSP v%d)", m_serverVersion.c_str(), m_htspVersion);
 }
 
-std::string CHTSPConnection::GetServerString ( void )
+std::string CHTSPConnection::GetServerString ( void ) const
 {
   const Settings &settings = Settings::GetInstance();
 
   CLockObject lock(m_mutex);
-  return StringUtils::Format("%s:%d [%s]", settings.GetHostname().c_str(), settings.GetPortHTSP(),
-             m_ready ? "connected" : "disconnected");
+  return StringUtils::Format("%s:%d", settings.GetHostname().c_str(), settings.GetPortHTSP());
 }
 
 bool CHTSPConnection::HasCapability(const std::string &capability) const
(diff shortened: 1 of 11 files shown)


ksooo:
The build errors are expected as ci builds against old API headers.

src/HTSPConnection.cpp
```diff
@@ -187,6 +194,32 @@ void CHTSPConnection::OnWake ( void )
...
+      m_state   = newState;
+
+      Logger::Log(LogLevel::LEVEL_DEBUG, "connection state change (%d -> %d)", prevState, newState);
+    }
+  }
+
+  if (prevState != newState)
+  {
+    /* Notify connection state change (callback!) */
+    PVR->ConnectionStateChange(GetServerString().c_str(), newState, NULL);
```
Jalle19: Couldn't this be done in the same if block as above?

src/HTSPConnection.cpp
```diff
@@ -362,7 +394,6 @@ htsmsg_t *CHTSPConnection::SendAndWait0 ( const char *method, htsmsg_t *msg, int
   if (!htsmsg_get_u32(msg, "noaccess", &noaccess) && noaccess)
   {
     // access denied
-    //XBMC->QueueNotification(QUEUE_ERROR, "Command %s failed: Access denied", method);
```
Jalle19: The acces denied state must be set here too, no?

src/HTSPConnection.cpp
```diff
@@ -157,8 +165,7 @@ std::string CHTSPConnection::GetServerString ( void )
   const Settings &settings = Settings::GetInstance();
 
   CLockObject lock(m_mutex);
-  return StringUtils::Format("%s:%d [%s]", settings.GetHostname().c_str(), settings.GetPortHTSP(),
-             m_ready ? "connected" : "disconnected");
+  return StringUtils::Format("%s:%d", settings.GetHostname().c_str(), settings.GetPortHTSP());
```
Jalle19: We should merge #161 first since it refactors e.g. this method.

src/HTSPTypes.h
```diff
@@ -91,17 +92,27 @@ enum eHTSPEventType
 struct SHTSPEvent
 {
   eHTSPEventType m_type;
-  uint32_t       m_idx;
```
Jalle19: Same thing here, I moved these events to a separate file.

src/HTSPTypes.h
```diff
@@ -91,17 +92,27 @@ enum eHTSPEventType
...
+
+  SHTSPEvent (eHTSPEventType type = HTSP_EVENT_NONE) :
     m_type(type),
-    m_idx (idx)
+    m_state(EPG_EVENT_CREATED)
   {
   }
-  
+
+  SHTSPEvent (eHTSPEventType type, const tvheadend::entity::Event &epg, EPG_EVENT_STATE state) :
```
Jalle19: With the code from #161 this special case event could be subclassed from the more generic base event which only needs a type member.

src/Tvheadend.cpp
```diff
@@ -1235,25 +1249,9 @@ PVR_ERROR CTvheadend::GetEpg
   /* Async transfer */
   if (Settings::GetInstance().GetAsyncEpg())
```
Jalle19: I would reverse this condition instead of having an empty code branch.

src/Tvheadend.h
```diff
@@ -178,7 +178,7 @@ class CHTSPConnection
   htsmsg_t *SendAndWait0    ( const char *method, htsmsg_t *m, int iResponseTimeout = -1);
   htsmsg_t *SendAndWait     ( const char *method, htsmsg_t *m, int iResponseTimeout = -1 );
 
-  inline int  GetProtocol      ( void ) const { return m_htspVersion; }
+  int  GetProtocol             ( void );
```
Jalle19: Make the mutex mutable instead of dropping const here.

L-S-D:
don't forget #169

ksooo:
@Jalle19 by any chance, I would like to get this PR merged before all the other refactoring PRs. The code is well-tested and I will have not much time during the next weeks for rebasing and retesting all this. What do you think?

Ofc, I will address the other - non-refactoring related - points you raised before merging this PR.

Jalle19:
@ksooo fine by me

ksooo:
@Jalle19 Thanks, Sam.

src/HTSPConnection.cpp
ksooo: No, callback must not called with mutex locked. https://github.com/kodi-pvr/pvr.hts/pull/195#discussion-diff-52122902R203

src/HTSPConnection.cpp
ksooo: No, as this is not a connection related error it does not alter the connection state. It's just a single operation (e.g. delete recording or timer) that failed.

src/Tvheadend.cpp
ksooo: Okay.

src/Tvheadend.h
ksooo: Okay.

src/HTSPConnection.cpp
Jalle19: You're right, just refreshed my memory by reading the wiki.

src/HTSPConnection.cpp
Jalle19: You're right, I missed the mutex part.

Jalle19:
@ksooo I guess it's good to go once the API bump in Kodi gets merged. @L-S-D #169 is not usable yet without more changes to the API (never had time to implement them since I'm not yet sure about how exactly it should be done).

ksooo:
>  I guess it's good to go once the API bump in Kodi gets merged.

 Cool. My plan is to merge the Kodi API bump PR mid to end of Feb.

Jalle19:
@ksooo wasn't it agreed that every addon maintainer has to implement addon changes on their own instead of having one guy do a reference implementation for all?

ksooo:
> wasn't it agreed that every addon maintainer has to implement addon changes on their own instead of having one guy do a reference implementation for all?

 I think if the implementation is trivial (in the sense of not requiring real knowledge of addon internals) if somebody steps up and implements everything (as I did here) all is fine.

 However, what we must avoid is providing implementations if we are not really know what we are doing - happened more than one time in the past with implementation of new functions - for instance with IsTimeshifting.

 That's why I refrained from implementing API 4.2.0 for pvr.dvlink and pvr.dvbviewer. These addons support timeshifting and I have no clue how to implement IsRealTimeStream for those.

Jalle19:
@ksooo fair point. I have nothing against "outsiders" keeping addons working, just wanted to check if I missed something in the previous discussion.

PR #203, opened 2016-03-01
Build pvr.hts branch Jarvis against updated p8platform

bas-t:
The build instructions in 09657735eae33bd85bb8324cb44f90de97e519a9 won't work on the Jarvis branch without this backport.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,10 +6,10 @@ enable_language(CXX)
 
 find_package(kodi REQUIRED)
 find_package(kodiplatform REQUIRED)
-find_package(platform REQUIRED)
+find_package(p8-platform REQUIRED)
 
 include_directories(${kodiplatform_INCLUDE_DIRS}
-                    ${platform_INCLUDE_DIRS}
+                    ${p8-platform_INCLUDE_DIRS}
                     ${KODI_INCLUDE_DIR}
                     ${PROJECT_SOURCE_DIR}/lib)
 
@@ -95,7 +95,7 @@ list(APPEND HTS_SOURCES
 
 add_subdirectory(lib/libhts)
 
-set(DEPLIBS ${platform_LIBRARIES} hts)
+set(DEPLIBS ${p8-platform_LIBRARIES} hts)
 if(WIN32)
   list(APPEND DEPLIBS ws2_32)
 endif()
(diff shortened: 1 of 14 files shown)


ksooo:
What is this good for? Why do you want to build Jarvis against p8-platform? Afaik p8-platform is for master only not for Jarvis.

bas-t:
Ah, I see...
This Travis system checks out specific commits before building stuff.
Well, that does not happen when building in a regular environment.

bas-t:
So the build fails.

ksooo:
Then Travis needs to be fixed, not kodi code.

bas-t:
Right, I did not know about travis before I  witnessed it failing this build.

bas-t:
BTW, p8 platform is just a renamed version of the former platform, right?
So I don't get it, you say it's ment for master only, but ,being the same code, only renamed, I disagree.

ksooo:
I'm not the deep into this topic but think it's not just about renaming. @opdenkamp should know better

opdenkamp:
it's a rename to make linux distributions (read: debian) happy. it does not change anything and is not needed for jarvis.

bas-t:
The only commit after renaming it to p8 platform is this one (apart from bumping the version)
https://github.com/Pulse-Eight/platform/commit/51e2b280b9704413050e01ce260dab1a1f8dc4cb
So I guess @opdenkamp knows his stuff, but seriously, the code is still the same.

bas-t:
I know it's not needed for Jarvis, it just breaks the build instructions in 0965773

opdenkamp:
yes, the code is still the same, what's your question? :)

ksooo:
> So I guess @opdenkamp knows his stuff, but seriously, the code is still the same.

 That's what he said. But what's your point? We need to fix the travis config, not pvr.hts Jarvis branch.

ksooo:
@opdenkamp hehe

bas-t:
Like I said before: I did not know about travis and could only build Jarvis without this backport.
That's all.

bas-t:
It gets worse, I'm making a total fool of myself... (yeah, I know, not the first time...)
I never noticed that this Travis system depends on clang....
(and I've never seen this new dependency published anywhere)
I guess Travis does not need fixing after all.

I'm very sorry for the noise

Issue #201, opened 2016-02-29
Recording lifetime "forever" does not work


ksooo:
At least with tvh 4.0.x (checked 4.0.8) this setting does quite the opposite of what it is supposed to do. The scheduled recording will be done just fine, but immediately after the recording is finished, the respective tvh database entry will be deleted. Seems, that it expires immediately (retention = 0 days). For the user it seems that the recording never happened as there is neither an entry in "finished recordings" nor in "failed recordings" and ofc no entry in Kodi timers UI. ;-)

I did experience this as well as a user in the forum. http://forum.kodi.tv/showthread.php?tid=262086

@Glenn-1990 maybe you can have a look? Maybe some sort of data overflow on tvh side? The "forever" value is quite big (2^31). Just guessing.

Glenn-1990:
I'll look into this, seems strange as I tested with 4.0 and 4.1

Glenn-1990:
It seems like a bug on the tvh side, we can just remove that option for 4.0? 
The retention/removal code was rewritten for 4.1, so that should just work.

ksooo:
> It seems like a bug on the tvh side, we can just remove that option for 4.0? 

 Yeah remove it for 4.0 as it actually is a fake anyway ("forever" by huge number of days, but technically not "forever").

 Please don't forget to change the default setting label like you already did for "until disk space needed" (tvh 4.1+ or so).

ksooo:
> It seems like a bug on the tvh side

Just a question. Have you debugged this? Is it a fact that it does not work for tvh 4.0 and that is does work for 4.1?

Glenn-1990:
I tested it partly, I have no acces to a tuner ATM, but everything is passed correctly to tvh.
Also I can see some possible time_t overflows in the 4.0 code:
https://github.com/tvheadend/tvheadend/blob/release/4.0/src/dvr/dvr_db.c#L1163

Is your server OS 64 or 32-bit?

ksooo:
OpenELEC x86_64 - sizeof(time_t) is 8 bytes - checked with a small test program.

Imo, changing https://github.com/tvheadend/tvheadend/blob/release/4.0/src/dvr/dvr_db.c#L1163 to <code>de->de_stop + dvr_entry_get_retention(de) * (time_t)86400</code>, like already in master (https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr_db.c#L417) should do the trick.

[...]

Issue #200, opened 2016-02-27
Hardcoded English strings in settings dialog


ksooo:
https://github.com/kodi-pvr/pvr.hts/blob/master/pvr.hts/resources/settings.xml#L30

Settings values are hardcoded English strings. They should be translatable.

@Glenn-1990 could you take over as you introduced those strings?

Glenn-1990:
I'll make them translatable ;-)

PR #205, opened 2016-03-08
Fix for #200 and #201

Glenn-1990:
https://github.com/kodi-pvr/pvr.hts/issues/201
Should be fixed in tvh master with this:
https://github.com/tvheadend/tvheadend/commit/055be40c0c3dffae117f1be2ce22885eb819f02c

So I disabled 'forever' for 4.0 and lower.

### DIFF
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,7 @@
+3.2.1
+- fixed: lifetime 'forever' should be disabled for HTSP v24 and lower 
+- improved: make lifetime settings translatable
+
 3.2.0
 - Updated to PVR API v5.1.0
 
--- a/pvr.hts/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.hts/resources/language/resource.language.en_gb/strings.po
@@ -297,7 +297,12 @@ msgctxt "#30388"
 msgid "Until space needed (tvh 4.1+)"
 msgstr ""
 
-#empty strings from id 30389 to 30399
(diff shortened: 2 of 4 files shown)


ksooo:
Looking good. Thanks.

Jalle19:
Travis will fail until https://github.com/kodi-pvr/pvr.hts/pull/206 is merged. Otherwise it looks good, thanks!

ksooo:
@MartijnKaijser wtf. Can't merge. Why? Thought this is only for non-team-members...

MartijnKaijser:
@ksooo it's an all or nothing switch. It must pass the build check for everyone. Not even direct pushes are allowed as all must go through PR. Up to you if you want to keep it.

ksooo:
> Up to you if you want to keep it.

 No, I don't want to keep it. This is too restrictive as long as builds are failing many times without a  reason that actually should prevent the merge.

MartijnKaijser:
Disabled again. Perhaps we should only do it on stable branches or something.

ksooo:
Thx @MartijnKaijser

ksooo:
@Glenn-1990 please rebase and bump version to 3.2.1.

Jalle19:
Can you add "fixes #foo" to the commit messages so the related issues will get closed automatically on merge?

Glenn-1990:
updated

Jalle19:
@ksooo your button

ksooo:
Thanks @Glenn-1990

PR #206, opened 2016-03-09
Implement PVR API 5.1.0

ksooo:
Refer to xbmc/xbmc#9290 and xbmc/xbmc#9295 for details.

@FernetMenta demuxer changes good to go?
@Jalle19 good to go?

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="3.1.0"
+  version="3.2.0"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
     <c-pluff version="0.1"/>
-    <import addon="xbmc.pvr" version="5.0.0"/>
+    <import addon="xbmc.pvr" version="5.1.0"/>
     <import addon="xbmc.codec" version="1.0.1"/>
   </requires>
   <extension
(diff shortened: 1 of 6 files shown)


ksooo:
@MilhouseVH could you do a runtime test?

FernetMenta:
@ksooo demuxer changes look ok but runtime test is more important in this case.

ksooo:
Hmm. Sounds like a general problem, not only pvr.hts related. Whatever you guys say makes sense.

I only would like to avoid another API bump right now. Regardless of the version number it's always PITA and I really would like somebody else to do the next monkey work, then.

ksooo:
Wouldn't proper <code>memset</code> before reassigning stream data in <code>ParseSubsriptionStart</code>solve the problem @Jalle19 mentioned?

Jalle19:
Yes a `memset()` would fix work, I just don't like polluting the code with C-isms just because we have a C API.

ksooo:
> Yes a memset() would fix work, I just don't like polluting the code with C-isms just because we have a C API.

 Fair enough. For **now** I would like to go with the memset approach. Okay?

MilhouseVH:
> @MilhouseVH could you do a runtime test?

To be honest I don't personally use any PVR functionality so runtime testing is difficult. I can build test it, that's about it - I leave runtime testing to the test users. :)

ksooo:
... now with memset.

ksooo:
Runtime tested. Works to some extend. One can watch tv, but on some channels audio stops after 3-5 seconds. Don't think this is an add-on problem.

@Jalle19 I suggest to merge this PR as it is now, so we have at least a compatible add-on again. Okay?

ksooo:
> I just don't like polluting the code with C-isms just because we have a C API.

 Lets fix this with the next API bump, which I'm pretty sure will come up before Krypton release anyway.

@Jalle19 okay to merge?

Jalle19:
Yeah go ahead, no point in having a broken addon.

Issue #214, opened 2016-04-17
predictive tuning crashes


nfedera:
If I enable predictive tuning Kodi (16.0, 16.1, master) always crashes (only tested this on osx):

```
Thread 38 Crashed:: DVDPlayer
0   pvr.hts.2.2.13.dylib          	0x000000011bc27423 tvheadend::ChannelTuningPredictor::PredictNextChannelId(unsigned int, unsigned int) const + 227
1   pvr.hts.2.2.13.dylib          	0x000000011bc1ce8b CTvheadend::PredictiveTune(unsigned int, unsigned int) + 59
2   pvr.hts.2.2.13.dylib          	0x000000011bc1d0c9 CTvheadend::DemuxOpen(PVR_CHANNEL const&) + 409
3   org.xbmc.kodi                 	0x000000010ea3088d PVR::CPVRClient::OpenStream(std::__1::shared_ptr<PVR::CPVRChannel> const&, bool) + 589
4   org.xbmc.kodi                 	0x000000010ea3720d PVR::CPVRClients::OpenStream(std::__1::shared_ptr<PVR::CPVRChannel> const&, bool) + 93
5   org.xbmc.kodi                 	0x000000010ea7d48b PVR::CPVRManager::OpenLiveStream(CFileItem const&) + 267
6   org.xbmc.kodi                 	0x000000010eaca092 XFILE::CPVRFile::Open(CURL const&) + 162
7   org.xbmc.kodi                 	0x000000010eacb794 CDVDInputStreamPVRManager::Open(char const*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool) + 228
8   org.xbmc.kodi                 	0x000000010e40a451 CDVDPlayer::OpenInputStream() + 465
9   org.xbmc.kodi                 	0x000000010e40cfdc CDVDPlayer::Process() + 28
10  org.xbmc.kodi                 	0x000000010e67973f CThread::Action() + 31
11  org.xbmc.kodi                 	0x000000010e678483 CThread::staticThread(void*) + 259
12  libsystem_pthread.dylib       	0x00007fff9629999d _pthread_body + 131
13  libsystem_pthread.dylib       	0x00007fff9629991a _pthread_start + 168
14  libsystem_pthread.dylib       	0x00007fff96297351 thread_start + 13
```

nfedera:
Fixed via #215 and #216

PR #215, opened 2016-04-17
Fix crash on initial predictive tuning

nfedera:
fixes issue #214

### DIFF
--- a/src/tvheadend/ChannelTuningPredictor.cpp
+++ b/src/tvheadend/ChannelTuningPredictor.cpp
@@ -74,7 +74,7 @@ uint32_t ChannelTuningPredictor::PredictNextChannelId(uint32_t tuningFrom, uint3
    * it will point at the channel we should tune to */
   std::set<ChannelPair>::iterator predictedIt = m_channels.cend();
 
-  if (std::next(fromIt, 1) == toIt || toIt->second == firstNum) {
+  if (fromIt == m_channels.cend() || std::next(fromIt, 1) == toIt || toIt->second == firstNum) {
     /* Tuning up or if we're tuning the first channel */
     predictedIt = ++toIt;
   }
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="3.2.2"
+  version="3.2.3"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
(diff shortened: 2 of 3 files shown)


Jalle19:
It might be because it's in the morning here but I don't think I completely understand this fix. Do you only have one channel or what is the scenario that is causing this crash?

src/tvheadend/ChannelTuningPredictor.cpp
```diff
@@ -74,7 +74,7 @@ uint32_t ChannelTuningPredictor::PredictNextChannelId(uint32_t tuningFrom, uint3
    * it will point at the channel we should tune to */
   std::set<ChannelPair>::iterator predictedIt = m_channels.cend();
 
-  if (std::next(fromIt, 1) == toIt || toIt->second == firstNum) {
+  if (fromIt == m_channels.cend() || std::next(fromIt, 1) == toIt || toIt->second == firstNum) {
```
Jalle19: If `fromIt` is equal to the end iterator it would mean that `GetIterator` doesn't find the channel in question which would mean the root cause lies somewhere else. `fromIt` and `toIt` being valid iterators are supposed to be invariants.

nfedera: @Jalle19 On my setup on the initial/first tune, the tuneFrom channelid parameter of `PredictNextChannelId` is zero which causes  `GetIterator` to return `end()` which in turn crashes if `std::next` is used on it. 
What should `tuningFrom` be set to initially ?

Jalle19: You're absolutely right, it seems the code runs into undefined behavior without this patch (advancing an iterator past the end). It seems to work on Linux and Windows though since no one has reported this yet.

@ksooo we'll have to backport this to the Jarvis branch.

nfedera: @Jalle19 @ksooo yes, would be great to get this into 16.1

Jalle19:
@ksooo any objections? Just waiting for AppVeyor to finish.

ksooo:
As always, I would like to see at least a changelog update, even better a bump of the addon version. We can then point to version x.y.z if users ask for a patch... to avoid confusion.

nfedera:
AppVeyor build in queue for 24 hours? Seems to be broken ...

Jalle19:
@nfedera just add a changelog entry and bump the version in addon.xml.in and it's good to go, I doubt the build will fail on Windows anyway.

nfedera:
@Jalle19 done. You want me to do an additional PR for the Jarvis branch?

ksooo:
> You want me to do an additional PR for the Jarvis branch?

yes, please.

nfedera:
Here you go: #216

PR #216, opened 2016-04-19
Fix crash on initial predictive tuning

nfedera:
Jarvis backport of commit 66b07a8a0f8aa7e65542a272fc1702e5d34e6269

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ language: cpp
 # Linux. We need Trusty in order to get up to date versions of 
 # cmake and g++.
 #
+env:
+  global:
+    - app_id=pvr.hts
+
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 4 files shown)


nfedera:
IMO the Travis cmake comand line requires `-DADDON_SRC_PREFIX=../../../..` instead of `-DADDON_SRC_PREFIX=../..`

Jalle19:
@nfedera the build was changed recently so it broke all older branches. @ksooo still good to go?

ksooo:
we should fix travis (cannot merge if failed, as this test is marked as required)

nfedera:
good to go now ?

ksooo:
Seems to work, but any reason for not doing it like in master? => https://github.com/kodi-pvr/pvr.hts/blob/master/.travis.yml#L45

nfedera:
@ksooo sure, I've adapted the travis config from master

ksooo:
Thanks.

PR #218, opened 2016-04-22
Rename addon title to "Tvheadend PVR Client Addon"

Hedda:
Suggest renaming of this addon title to "Tvheadend PVR Client Addon" as I think that it makes it more clear and less confusing if HTSP is removed and suffix "PVR Client Addon" is added, and the information about the HTSP protocol being used is being mentioned to the description instead.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -2,7 +2,7 @@
 <addon
   id="pvr.hts"
   version="3.2.3"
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
Adding "addon" to the name is stupid.

ksooo:
Makes no sense.

Jalle19:
@Hedda while I appreciate the effort I don't think it makes sense to change addon names, atleast not right now.

PR #226, opened 2016-05-17
Fix instant timers.

ksooo:
Instant timers are actually manual timers, esp. start and end time, no matter they are created from epg tags. => https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/timers/PVRTimerInfoTag.cpp#L790

@Jalle19 good to go?

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="3.4.0"
+  version="3.4.1"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+3.4.1
+- Fix instant timers
+
 3.4.0
 - Cmake: rename find_package kodi to Kodi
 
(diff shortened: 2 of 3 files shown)


Glenn-1990:
@ksooo When watching tv fullscreen and pressing the record button (on remote or in video osd), the current show will be recorded. Will this PR break this behavior?

ksooo:
@Glenn-1990 it will change the behavior to behave like desired (=> https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/timers/PVRTimerInfoTag.cpp#L790), not break it. Kodi schedules an instant timer, currently pvr.hts does ignore this and schedules an epg based timer for the current broadcasting.

I know that some people do want Kodi to record the current show, others want a fixed length timer. There is no "right" and "wrong" here. That's why I'm working on making the behavior configurable in Kodi.

Glenn-1990:
@ksooo 
Yes, changing the Kodi behavior seems the best approach to handle this case. 
Maybe an additional setting isn't needed at all and simply do:
1) epg presens -> record current show
2) no epg -> record as long as defined (instant timer duration setting)
If I remember correctly, that was the way mediaportal handles this case.

But that's all your choice of course ;-)

ksooo:
@Glenn-1990 it's not that easy. ;-) Currently, I'm aware of two threads in Kodi forum where people requesting exacly the opposite behavior. In one thread they want what you prefer, in the other thread quite the opposite. I want to make both parties happy, thus the approach with a new setting for Kodi.

Jalle19:
Personally I want the current show to be recorded when I press record. I wouldn't mind if the behavior changes, but if you have the instant recording duration set to e.g. 120 minutes and the current show is 140 minutes you may end up with a recording that ends prematurely.

I'm all for a setting though.

ksooo:
Working on the Kodi setting. 

Is this PR good to go? Strictly speaking it is an unrelated bug fix. The fact that it "breaks" your preferred behavior is just a coincidence.

Jalle19:
Yeah I guess so.

ksooo:
rebased

PR #230, opened 2016-06-14
HTSP v25 additions

Glenn-1990:
Changes:
1) bIsRadio -> channelType: channelType can have 'radio', 'tv' or 'other' assigned to it. By this way, we can hide all channels marked as 'other' from Kodi. With the current implementation, channels with data/download services where shown as tv channels in Kodi...
2) Recordings with a deleted channel are now correctly identified as radio/television and the channel name will still be present. 
The current implementation just maps everything to tv recordings in this case.

@ksooo Part 2 is the one you requested some time ago.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -195,7 +195,7 @@ PVR_ERROR CTvheadend::GetTags ( ADDON_HANDLE handle, bool bRadio )
       /* Does group contain channels of the requested type?             */
       /* Note: tvheadend groups can contain both radio and tv channels. */
       /*       Thus, one tvheadend group can 'map' to two Kodi groups.  */
-      if (!entry.second.ContainsChannelType(bRadio))
+      if (!entry.second.ContainsChannelType(bRadio ? CHANNEL_TYPE_RADIO : CHANNEL_TYPE_TV))
         continue;
 
       PVR_CHANNEL_GROUP tag;
@@ -245,7 +245,8 @@ PVR_ERROR CTvheadend::GetTagMembers
       {
         auto cit = m_channels.find(channelId);
 
-        if (cit != m_channels.cend() && cit->second.IsRadio() == group.bIsRadio)
+        if (cit != m_channels.cend() && cit->second.GetType() == (group.bIsRadio ?
+            CHANNEL_TYPE_RADIO : CHANNEL_TYPE_TV))
         {
           PVR_CHANNEL_GROUP_MEMBER gm;
           memset(&gm, 0, sizeof(PVR_CHANNEL_GROUP_MEMBER));
@@ -295,14 +296,14 @@ PVR_ERROR CTvheadend::GetChannels ( ADDON_HANDLE handle, bool radio )
     {
       const auto &channel = entry.second;
 
-      if (radio != channel.IsRadio())
+      if (channel.GetType() != (radio ? CHANNEL_TYPE_RADIO : CHANNEL_TYPE_TV))
         continue;
 
       PVR_CHANNEL chn;
       memset(&chn, 0 , sizeof(PVR_CHANNEL));
 
       chn.iUniqueId         = channel.GetId();
-      chn.bIsRadio          = channel.IsRadio();
+      chn.bIsRadio          = radio;
       chn.iChannelNumber    = channel.GetNum();
       chn.iSubChannelNumber = channel.GetNumMinor();
       chn.iEncryptionSystem = channel.GetCaid();
@@ -482,10 +483,22 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
         auto cit = m_channels.find(rec.iChannelUid);
         if (cit == m_channels.cend())
           rec.channelType = PVR_RECORDING_CHANNEL_TYPE_UNKNOWN;
-        else if (cit->second.IsRadio())
-          rec.channelType = PVR_RECORDING_CHANNEL_TYPE_RADIO;
         else
-          rec.channelType = PVR_RECORDING_CHANNEL_TYPE_TV;
+        {
+          switch (cit->second.GetType())
+          {
+            case CHANNEL_TYPE_TV:
+              rec.channelType = PVR_RECORDING_CHANNEL_TYPE_TV;
+              break;
+            case CHANNEL_TYPE_RADIO:
+              rec.channelType = PVR_RECORDING_CHANNEL_TYPE_RADIO;
+              break;
+            case CHANNEL_TYPE_OTHER:
+            default:
+              rec.channelType = PVR_RECORDING_CHANNEL_TYPE_UNKNOWN;
+              break;
+          }
+        }
       }
 
       recs.push_back(rec);
@@ -1811,17 +1824,26 @@ void CTvheadend::ParseChannelAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   {
     htsmsg_field_t *f;
     uint32_t caid  = 0;
-    bool     radio = false;
     HTSMSG_FOREACH(f, list)
     {
       if (f->hmf_type != HMF_MAP)
         continue;
 
-      /* Radio? */
-      if ((str = htsmsg_get_str(&f->hmf_msg, "type")) != NULL)
+      /* Channel type */
+      if (m_conn.GetProtocol() >= 25)
       {
-        if (!strcmp(str, "Radio"))
-          radio = true;
+        if (!htsmsg_get_u32(&f->hmf_msg, "content", &u32))
+          channel.SetType(u32);
+      }
+      else
+      {
+        if ((str = htsmsg_get_str(&f->hmf_msg, "type")) != NULL)
+        {
+          if (!strcmp(str, "Radio"))
+            channel.SetType(CHANNEL_TYPE_RADIO);
+          else
+            channel.SetType(CHANNEL_TYPE_TV);
+        }
       }
 
       /* CAID */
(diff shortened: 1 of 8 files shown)


src/Tvheadend.cpp
```diff
@@ -1827,7 +1827,7 @@ void CTvheadend::ParseChannelAddOrUpdate ( htsmsg_t *msg, bool bAdd )
         {
           if (!strcmp(str, "Radio"))
             channel.SetType(PVR_RECORDING_CHANNEL_TYPE_RADIO);
-          else
+          else if (!strcmp(str, "SDTV") || !strcmp(str, "HDTV"))
```
ksooo: What about "UHDTV"? => https://github.com/tvheadend/tvheadend/blob/7b28d753903d4e963965335932b568a1e20ecf1d/src/service.c#L1342

src/Tvheadend.cpp
```diff
@@ -482,10 +483,8 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
         auto cit = m_channels.find(rec.iChannelUid);
         if (cit == m_channels.cend())
           rec.channelType = PVR_RECORDING_CHANNEL_TYPE_UNKNOWN;
-        else if (cit->second.IsRadio())
-          rec.channelType = PVR_RECORDING_CHANNEL_TYPE_RADIO;
         else
-          rec.channelType = PVR_RECORDING_CHANNEL_TYPE_TV;
+          rec.channelType = static_cast<PVR_RECORDING_CHANNEL_TYPE>(cit->second.GetType());
```
Glenn-1990: Yes, you're right, will change this.

src/Tvheadend.cpp
Glenn-1990: "UHDTV" was added with htspv25, this code only gets called when htsp version < 25.

src/Tvheadend.cpp
```diff
@@ -1917,9 +1907,64 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
+          }
+        }
+      }
+      rec.SetChannelType(hasVideo ? PVR_RECORDING_CHANNEL_TYPE_TV :
+          (hasAudio ? PVR_RECORDING_CHANNEL_TYPE_RADIO : PVR_RECORDING_CHANNEL_TYPE_UNKNOWN));
+    }
+  }
+
+  /* Channel name fallback (in case channel was deleted) */
+  if (rec.GetChannelName().empty()  && m_conn.GetProtocol() >= 25)
```
ksooo: Please remove one of the two spaces between <code>empty()</code> and <code>&& m_conn</code>.

src/Tvheadend.cpp
```diff
@@ -1917,9 +1907,64 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
+  }
+
+  /* Channel type fallback (in case channel was deleted) */
+  if (!rec.GetChannelType() && m_conn.GetProtocol() >= 25)
+  {
+    htsmsg_t *files, *streams;
+    if ((files = htsmsg_get_list(msg, "files")) != NULL)
+    {
+      htsmsg_field_t *file, *stream;
+      uint32_t hasAudio = 0, hasVideo  = 0, u32;
```
ksooo: Please remove one of the two spaces between <code>hasVideo</code> and <code>= 0</code>.

src/Tvheadend.cpp
```diff
@@ -1917,9 +1907,64 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   rec.SetStop(stop);
 
   /* Channel is optional, it may not exist anymore */
-  if (!htsmsg_get_u32(msg, "channel", &channel))
+  if (!htsmsg_get_u32(msg, "channel", &channel) && channel > 0)
```
ksooo: Why the additional condition? Is 0 an invalid channel id?

src/Tvheadend.cpp
```diff
@@ -1917,9 +1907,64 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
+          HTSMSG_FOREACH(stream, streams) // Loop through all streams
+          {
+            if (stream->hmf_type != HMF_MAP)
+              continue;
+
+            if (!htsmsg_get_u32(&stream->hmf_msg, "audio_type", &u32)) // Only present for audio streams
+              hasAudio = 1;
+            if (!htsmsg_get_u32(&stream->hmf_msg, "aspect_num", &u32)) // Only present for video streams
+              hasVideo = 1;
+            if (hasAudio && hasVideo) // No need to parse the rest
```
ksooo: Hmm. The above lines look somehow hacky to me. Might work, but still...

src/Tvheadend.cpp
```diff
@@ -418,16 +418,17 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
...
-        strncpy(rec.strChannelName, cit->second.GetName().c_str(),
-                sizeof(rec.strChannelName) - 1);
-
         strncpy(rec.strIconPath, cit->second.GetIcon().c_str(),
                 sizeof(rec.strIconPath) - 1);
       }
 
+      /* Channel name */
+      strncpy(rec.strChannelName, recording.GetChannelName().c_str(),
+              sizeof(rec.strChannelName) - 1);
```
ksooo: This is a bad idea if recording.GetChannelName returns an empty string. Was not good in original code, btw.

src/Tvheadend.cpp
Glenn-1990: Yes, it's not that clean but perexg preferred this solution as it doesn't need much additional logic in tvheadend. Remember that this only handles recordings with a missing channel, tested with radio and tv recording (without channel ofc).

Check the first comment here:
https://github.com/tvheadend/tvheadend/pull/847

src/Tvheadend.cpp
Glenn-1990: @ksooo what's wrong with an empty string as long it's initialized?

Glenn-1990:
@ksooo updated

src/Tvheadend.cpp
ksooo: > what's wrong with an empty string as long it's initialized?

 Nothing. My fault, sorry. Mixed recording.GetChannelName() return value which can be empty with rec.strChannelName which cannot be zero lenght.

src/Tvheadend.cpp
ksooo: Ah, okay then.

src/tvheadend/entity/Channel.h
```diff
@@ -44,7 +44,7 @@ namespace tvheadend
       Channel() :
         m_num(0),
         m_numMinor(0),
-        m_radio(false),
+        m_type(0),
```
ksooo: CHANNEL_TYPE_OTHER ?

src/tvheadend/entity/Tag.cpp
```diff
@@ -90,7 +90,7 @@ bool Tag::ContainsChannelType(bool bRadio) const
   {
     if ((cit = channels.find(*it)) != channels.end())
     {
-      if (bRadio == cit->second.IsRadio())
+      if (cit->second.GetType() == (bRadio ? CHANNEL_TYPE_RADIO : CHANNEL_TYPE_TV))
```
ksooo: Shouldn't <code>Tag::ContainsChannelType(bool bRadio)</code> consequently be changed to <code>Tag::ContainsChannelType(channel_type_t eType)? Currently implmentation returns true for "other" channels if bRadio is false which is not correct.

src/Tvheadend.cpp
```diff
@@ -482,10 +483,21 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
...
+        {
+          switch (cit->second.GetType())
+          {
+            case CHANNEL_TYPE_TV:
+              rec.channelType = PVR_RECORDING_CHANNEL_TYPE_TV;
+              break;
+            case CHANNEL_TYPE_RADIO:
+              rec.channelType = PVR_RECORDING_CHANNEL_TYPE_RADIO;
+              break;
+            default:
```
ksooo: For the sake of completeness I would also handle CHANNEL_TYPE_OTHER explicitely here.

src/tvheadend/entity/Tag.cpp
Glenn-1990: If bRadio is false, the function returns true for tv channels only..
But changing the function to what you suggest makes sense.

Glenn-1990:
@ksooo updated

ksooo:
@Glenn-1990 many thanks. Could you please add changelog and version bump? Then we are ready to merge this.

Glenn-1990:
@ksooo version bump done

ksooo:
@Glenn-1990 please rebase. 3.4.2 already exists ;-)

Glenn-1990:
@ksoo, oops my bad, rebased now

ksooo:
@Glenn-1990 unrelated, but i just stumbled over the fact that DVR_RET_SPACE and DVR_RET_FOREVER are both are defined as INT32_MAX-1. This feels wrong.

https://github.com/kodi-pvr/pvr.hts/pull/230/files#diff-2db6f295170a206772114b26ca4174a0R79

Seems also be buggy in tvheadend.

https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr.h#L146

???

Glenn-1990:
@ksooo 
This is correct as dvr_retention_t and dvr_removal_t are combined to one.
retention INT32_MAX-1 = remove log file when dvr file gets removed.
removal INT32_MAX-1 = remove dvr file when space is needed.
For all other cases, remove and retention are similar.

ksooo:
okay, then. thanks for the clarification

PR #232, opened 2016-06-27
Fix setting channel type & setting epg uid

ksooo:
@Glenn-1990 fyi. for commit 1, please note my comment in the code regarding the evil of adding new functionality to htsp api without bumping htsp version. :-/

@seo @da-anda this brings pvr.hts back to life for you (and me) and probably many others using latest 'official' tvheadend raspberry build (dated january 2016 or so).

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="3.4.3"
+  version="3.4.4"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,7 @@
+3.4.4
+- fixed: setting channel type for 'early' htsp v25 servers
+- fixed: setting epg uid for timers
+
 3.4.3
 - improved: only show real radio and TV channels in Kodi
 - improved: recording handling when it's channel doesn't exist anymore (HTSP v25 and above)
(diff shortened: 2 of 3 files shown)


ksooo:
@MilhouseVH would be great if you include this into next test build as it fixes an ugly bug... will merge this PR in a second.

MilhouseVH:
Yep, now that it's merged it will be in tonight's build.

Glenn-1990:
@ksooo 
I don't get why we need this hack, we agreed some time ago that we could not support all tvheadend dev builds? The code change in tvh is already there for months.

ksooo:
Latest tvheadend builds for Raspbian that can be installed from official unstable tvheadend ppa (http://apt.tvheadend.org/unstable) are 4.1-1429 (jessie) and 4.1-1434 (wheezy), dated end of January 2016. Many people are using tvh on raspi, many use the builds mentioned above (simply, because there is nothing more recent available). Unfortunately, for these people, your change completely breaks Kodi Krypton PVR when used with pvr.hts 3.4.3. No channels will be reported to Kodi, because they are all recognised as neither radio nor TV... 

I think it worth the "hack" in this special case (low effort/many users happy). We will release Krypton Alpha the next days and I do not want to explain all the bleeding edge raspi kodi/tvh users that they cannot test Kodi Krypton PVR because they are using the "wrong" tvheadend server.

Generally, I consider it not good (because not save) to currently implement pvr.hts code that depends on htsp >= 25. As a matter of fact, we can only check for > 25 if we want to be sure that we get all 25 functionality from the server.

IMO, it is wrong not to bump htsp version if new htsp features get implemented. It's just an increasing number, nothing else. Not much effort, but definitely helpful/important for the clients. Don't get why tvh people do not just do it.

Jalle19:
Peehaps we can drop it before Krypton final?

Glenn-1990:
@ksooo 
I remember a conversation somewhere (maybe with you, don't kwow anymore) that bumping the htsp api for each small change was to much hassle, meaning that we only can support the stable and very recent builds. People with alpha/beta builds will probably also update tvheadend very often, when pvr.hts changes are weeks later than the tvheadend changes, less users will face this issue. 
Anyway, if @perexg is ok with it, I'll add a htsp bump when doing some htsp stuff next time ;-)
@perexg is it possible to update the http://apt.tvheadend.org/unstable builds more often to avoid incompatibilities as described here? Then this hack can be dropped sooner or later.

Issue #228, opened 2016-06-01
build instructions unclear


tmelin:
I think there is an issue with the build instructions. If one tries to build according to them, it doesn't work.
It will give:

CMake Error: The source directory "xbmc/xbmc/project/xbmc/project/cmake/addons" does not exist.

When building according to paths given in the instructions, the pvr.hts repo needs to be cloned inside the xbmc repo into folder xbmc/project.

Could the build instructions please be clarified on this point, or otherwise changed?

Also, I could never find any generated zip file after building, assuming the -DPACKAGE_ZIP=1 refers to that being requested.

Thanks,
Tomas

Jalle19:
I just copy-pasted the exact commands from the README and it built just fine. There's a missing `make install` in the end though, I'll make a PR for that.

Make sure your `pvr.hts` and `xbmc` directories are in the same directory.

tmelin:
It seems to work if doing the commands from the README, but not if checking out the version branch first. So e.g. in my case I first check out branch Jarvis and then run the commands. cmake fails on error.

Jalle19:
You need to build against the Jarvis branch of the addon if you're building against a Jarvis Kodi. Also check the README for that branch, it could be that the instructions have changed since then.

tmelin:
That is exactly what doesn't work for me if following the README instructions 

Following README without any checkout - build works
Following README and checking out Jarvis in both Kodi and addon repo - build fails.

Jalle19:
I guess I'll have to try building the Jarvis branch.

jiphex:
I had a similar problem building the Jarvis branch of this addon, cmake was complaining about a missing addon source directory of `xbmc/project/pvr.hts`. I solved it by doing:

    cd xbmc/project/
    ln -s ../../pvr.hts

After doing that I could build the Jarvis branch

Jalle19:
I tried the build instructions recently when building for master and they're still correct. A slight change is needed if you're building Kodi with CMake, that will be fixed by #239

PR #239, opened 2016-09-18
Minor README updates

Jalle19:


Jalle19:
Okay so apparently we do use AppVeyor still. @MartijnKaijser can you tell me the correct badge URL? I can't determine it because it contains some token only visible from the project settings.

Jalle19:
@MartijnKaijser ?

Jalle19:
We'll re-add the badge later, the README changes fix https://github.com/kodi-pvr/pvr.hts/issues/228

PR #242, opened 2016-09-22
Fix Travis XCode 6.1 build by upgrading CMake

Jalle19:
tinyxml requires at least CMake 3.1


ksooo:
Should this fix be applied to all our pvr addons? ping @hudokkow

Jalle19:
@ksooo this is only required for the XCode 6.1 build and only some addons require a newer CMake version than that platform ships with. It would probably not hurt to have the fix everywhere though, as soon as I can come up with something that actually works.

hudokkow:
We should eventually require 3.1 for all add-ons (same as core) but no need to rush if travis is happy with lower. btw, one of the
> <code>script: make</code>

lines can go.

PR #249, opened 2016-11-22
Bugfix/epg loading time improvements

adamsutton:
When through a few variations in trying to figure out the performance issues and then improve, but in the end it would have probably been simpler just to use the solution from the final commit and just that.

But hey ho, it's MUCH faster than before and "almost" tolerable now. I think any additional delays are out of pvr.hts control.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="3.4.7"
+  version="3.4.8"
   name="Tvheadend HTSP Client"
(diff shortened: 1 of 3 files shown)


adamsutton:
There you go, much simpler.

ksooo:
Thanks @adamsutton could you add add-on version bump and changelog entry, please?

Jalle19:
I was about to ask the same thing.

ksooo:
@adamsutton thanks
@Jalle19 good to go?

ksooo:
@adamsutton i just runtime tested this. EPG loading is so much faster now! Thanks you so much for this fix.

adamsutton:
@ksooo glad to be of service, it's been annoying the hell out of me for ages! Just not had the time to look at it.

PR #251, opened 2016-11-27
3.4.10. Adpated to pvr addon api 5.2.1 (3.4.9 was just to ake it compile again).

ksooo:
@Jalle19 we have to cleanup to actually benefit from the API change.

### DIFF
--- a/src/tvheadend/Subscription.cpp
+++ b/src/tvheadend/Subscription.cpp
@@ -173,7 +173,7 @@ void Subscription::SendUnsubscribe(void)
   Logger::Log(LogLevel::LEVEL_DEBUG, "demux successfully unsubscribed from channel id %d, subscription id %d", GetChannelId(), GetId());
 }
 
-bool Subscription::SendSeek(int time)
+bool Subscription::SendSeek(double time)
 {
   /* Build message */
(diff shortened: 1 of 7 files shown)


src/tvheadend/Subscription.cpp
```diff
@@ -173,12 +173,12 @@ void Subscription::SendUnsubscribe(void)
...
 }
 
-bool Subscription::SendSeek(int time)
+bool Subscription::SendSeek(double time)
 {
   /* Build message */
   htsmsg_t *m = htsmsg_create_map();
   htsmsg_add_u32(m, "subscriptionId", GetId());
-  htsmsg_add_s64(m, "time",           static_cast<int64_t>(time * 1000LL));
+  htsmsg_add_s64(m, "time",           time * 1000LL);
```
Jalle19: Why no cast anymore? `htsmsg_add_s64` still expects an `int64_t`, not a double, right?

ksooo: Good point. Will revert this line.

ksooo:
@Jalle19 addressed your review comment. Good to go now?

Jalle19:
+1

PR #257, opened 2016-12-13
HTSP v26 support

Glenn-1990:
Htsp v26 support.

@ksooo @Jalle19 

Edit: removed WIP

### DIFF
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -495,6 +495,7 @@ bool CHTSPConnection::SendHello ( void )
 bool CHTSPConnection::SendAuth
   ( const std::string &user, const std::string &pass )
 {
+  uint32_t u32;
   htsmsg_t *msg = htsmsg_create_map();
   htsmsg_add_str(msg, "username", user.c_str());
 
@@ -513,7 +514,33 @@ bool CHTSPConnection::SendAuth
   /* Send and Wait */
   msg = SendAndWait0("authenticate", msg);
 
-  return (msg != NULL);
+  if (msg == NULL)
+    return 0;
+
+  if (m_htspVersion >= 26)
+  {
+    /* Log received permissions */
+    Logger::Log(LogLevel::LEVEL_INFO, "  Received permissions:");
+    if (!htsmsg_get_u32(msg, "admin", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  administrator              : %i", u32);
+    if (!htsmsg_get_u32(msg, "streaming", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  HTSP streaming             : %i", u32);
+    if (!htsmsg_get_u32(msg, "dvr", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  HTSP DVR                   : %i", u32);
+    if (!htsmsg_get_u32(msg, "faileddvr", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  Failed/aborted DVR         : %i", u32);
+    if (!htsmsg_get_u32(msg, "anonymous", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  anonymous HTSP only        : %i", u32);
+    if (!htsmsg_get_u32(msg, "limitall", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  global connection limit    : %i", u32);
+    if (!htsmsg_get_u32(msg, "limitdvr", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  DVR connection limit       : %i", u32);
+    if (!htsmsg_get_u32(msg, "limitstreaming", &u32))
+      Logger::Log(LogLevel::LEVEL_INFO, "  streaming connection limit : %i", u32);
+  }
+
+  htsmsg_destroy(msg);
+  return 1;
 }
 
 /**
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -1826,20 +1826,12 @@ void CTvheadend::ParseChannelAddOrUpdate ( htsmsg_t *msg, bool bAdd )
         continue;
 
       /* Channel type */
-      bool bGotContent = false;
-      if (m_conn.GetProtocol() >= 25)
+      if (m_conn.GetProtocol() >= 26)
       {
         if (!htsmsg_get_u32(&f->hmf_msg, "content", &u32))
-        {
           channel.SetType(u32);
-          bGotContent = true;
-        }
       }
-
-      // The 'content' htsp method field was added to tvheadend htsp api without htsp version bump.
-      // Unfortunately, there are many semi-official tvheadend builds with htsp version 25 in the wild which
-      // do not support the 'content' htsp attribute. Just checking htsp api version is not sufficient. :-/
-      if (!bGotContent)
+      else
       {
         if ((str = htsmsg_get_str(&f->hmf_msg, "type")) != NULL)
         {
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -1908,31 +1908,27 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
     return;
   }
 
-  if (htsmsg_get_s64(msg, "start", &start) && bAdd)
-  {
-    Logger::Log(LogLevel::LEVEL_ERROR, "malformed dvrEntryAdd: 'start' missing");
-    return;
-  }
-
-  if (htsmsg_get_s64(msg, "stop", &stop) && bAdd)
-  {
-    Logger::Log(LogLevel::LEVEL_ERROR, "malformed dvrEntryAdd: 'stop' missing");
-    return;
-  }
-
-  if (((state = htsmsg_get_str(msg, "state")) == NULL) && bAdd)
-  {
-    Logger::Log(LogLevel::LEVEL_ERROR, "malformed dvrEntryAdd: 'state' missing");
-    return;
-  }
-
   /* Get entry */
   Recording &rec = m_recordings[id];
   Recording comparison = rec;
   rec.SetId(id);
   rec.SetDirty(false);
-  rec.SetStart(start);
-  rec.SetStop(stop);
+
+  if (!htsmsg_get_s64(msg, "start", &start))
+    rec.SetStart(start);
+  else if (bAdd)
+  {
+    Logger::Log(LogLevel::LEVEL_ERROR, "malformed dvrEntryAdd: 'start' missing");
+    return;
+  }
+
+  if (!htsmsg_get_s64(msg, "stop", &stop))
+    rec.SetStop(stop);
+  else if (bAdd)
+  {
+    Logger::Log(LogLevel::LEVEL_ERROR, "malformed dvrEntryAdd: 'stop' missing");
+    return;
+  }
 
   /* Channel is optional, it may not exist anymore */
   if (!htsmsg_get_u32(msg, "channel", &channel))
@@ -2052,9 +2048,9 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
     return;
   }
 
-  if (state != NULL)
+  /* Parse state */
+  if ((state = htsmsg_get_str(msg, "state")) != NULL)
   {
-    /* Parse state */
     if      (strstr(state, "scheduled") != NULL)
       rec.SetState(PVR_TIMER_STATE_SCHEDULED);
     else if (strstr(state, "recording") != NULL)
@@ -2065,6 +2061,11 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
       rec.SetState(PVR_TIMER_STATE_ERROR);
     else if (strstr(state, "invalid") != NULL)
       rec.SetState(PVR_TIMER_STATE_ERROR);
+  }
+  else if (bAdd)
+  {
+    Logger::Log(LogLevel::LEVEL_ERROR, "malformed dvrEntryAdd: 'state' missing");
+    return;
   }
 
   /* Add optional fields */
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -452,6 +452,10 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
       /* Lifetime (based on retention or removal) */
       rec.iLifetime = recording.GetLifetime();
 
+      /* Play status */
+      rec.iPlayCount = recording.GetPlayCount();
+      rec.iLastPlayedPosition = recording.GetPlayPosition();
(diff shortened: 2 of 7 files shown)


ksooo:
> The pvr addon capabilities are requested by Kodi before we can determine the htsp version. That's why there is a setting for now.

The setting should not be nessecary. Yes, on first call of getAddonCapabilites we are not connected (most probably), but addon capabilites are re-read directly after the addon state changes to CONNECTED. => https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/addons/PVRClients.cpp#L1536 If this does not work, it's broken.

ksooo:
> "GetTimerTypes" does several htsp version checks, but this is somehow broken.
Htsp version is not determined yet when calling this function. The connection state is still "PVR_CONNECTION_STATE_CONNECTING" then.

 Are you sure it does not work? GetTimerTypes should be called not once once, but on demand, for instance everytime timer settings dialog is opened, and there is no caching of the types in Kodi PVR core. So, it should work. However, if there is a call to GetTimerTypes that comes to "early", this could be a bug in Kodi PVR core which has no serious consequences, though. I will look inti this.

Glenn-1990:
@ksooo Good to know that GetTimerTypes gets called every time, that pointed me to problem and solution ;-)

Is rereading the addon settings after connecting a new thing? I always remembered that this was a limitation, I'll see if I can get it working without the setting then :-)

ksooo:
So, both "problems" solved? 

Rereading add-on capabilities (not settings, that's different!) is there since day 1 of async connect, means since very early Krypton alphas.

ksooo:
Just looked it up in the code: both addon capabilities and timer types are reread if connection state changes to CONNECTED. So, timer types are not always fetched from the backend, like I said earlier, but only on (re)connect, exactly like addon capabilities.

Glenn-1990:
Oops, closed by accident.

@ksooo Both issues resolved, this async stuff is really an improvement ;-)

Jalle19:
@Glenn-1990 the "delete channels without services" can be dropped now that you did https://github.com/tvheadend/tvheadend/commit/e73a8b455a5e2449281f66f6816fef0891c2cc6e right?

Glenn-1990:
@Jalle19 indeed ;-)

piotrasd:
as the tvheadend getting ecm time and encoding system from oscam could we ship that over htsp to kodi ? sometimes we just could know why we lost picture, beucase of signal problem/lost or oscam problem, or maybe privileges on card gone ... etc. 

Thx

ksooo:
@piotrasd may i kindly ask you not to hijack a random PR for completely unrelated feature requests. Thank you.

Glenn-1990:
@ksooo @Jalle19 Updated some commits

src/HTSPConnection.cpp
```diff
@@ -495,6 +495,7 @@ bool CHTSPConnection::SendHello ( void )
 bool CHTSPConnection::SendAuth
   ( const std::string &user, const std::string &pass )
 {
+  uint32_t access, u32;
```
Jalle19: `access` should be bool

src/HTSPConnection.cpp
```diff
@@ -513,7 +514,43 @@ bool CHTSPConnection::SendAuth
   /* Send and Wait */
   msg = SendAndWait0("authenticate", msg);
 
-  return (msg != NULL);
+  // <= htsp v25, msg present = no access, msg missing = access
+  // >= htsp v26, msg should be present in any case
+  if (m_htspVersion <= 25) //
+    access = msg ? 0 : 1;
```
Jalle19: I would return early end skip the `else` clause, but it's up to you

src/HTSPConnection.cpp
```diff
@@ -513,7 +514,43 @@ bool CHTSPConnection::SendAuth
...
+  else
+  {
+    if (msg == NULL || (!htsmsg_get_u32(msg, "noaccess", &u32) && u32))
```
Jalle19: nullptr

src/HTSPConnection.cpp
```diff
@@ -513,7 +514,43 @@ bool CHTSPConnection::SendAuth
...
+      Logger::Log(LogLevel::LEVEL_DEBUG, "authentication failed");
+    }
+    else
+    {
+      access = 1;
+      Logger::Log(LogLevel::LEVEL_DEBUG, "authentication successful");
+
+      /* Log received permissions */
+      if (!htsmsg_get_u32(msg, "admin", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "admin permissions: %i", u32);
```
Jalle19: Can you log these lines with two spaces extra indentation, like we do for demuxer streams? I find it makes the log more readable.

src/HTSPConnection.cpp
```diff
@@ -513,7 +514,43 @@ bool CHTSPConnection::SendAuth
...
+      if (!htsmsg_get_u32(msg, "streaming", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "streaming permissions: %i", u32);
+      if (!htsmsg_get_u32(msg, "dvr", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "dvr permissions: %i", u32);
```
Jalle19: DVR in uppercase, same below for "dvr connection limit"

src/HTSPConnection.cpp
```diff
@@ -513,7 +514,43 @@ bool CHTSPConnection::SendAuth
...
+      if (!htsmsg_get_u32(msg, "anonymous", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "anonymous htsp access: %i", u32);
```
Jalle19: HTSP should be capitalized

src/HTSPConnection.cpp
```diff
@@ -513,7 +514,43 @@ bool CHTSPConnection::SendAuth
...
+      if (!htsmsg_get_u32(msg, "limitall", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "global connection limit: %i", u32);
+      if (!htsmsg_get_u32(msg, "limitdvr", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "dvr connection limit: %i", u32);
+      if (!htsmsg_get_u32(msg, "limitstreaming", &u32))
+        Logger::Log(LogLevel::LEVEL_INFO, "sreaming connection limit: %i", u32);
```
Jalle19: typo (should be streaming)

src/Tvheadend.cpp
```diff
@@ -598,6 +602,43 @@ PVR_ERROR CTvheadend::RenameRecording ( const PVR_RECORDING &rec )
...
+  htsmsg_add_u32(m, "playposition", playposition);
+  return SendDvrUpdate(m);
+}
+
+int CTvheadend::GetPlayPosition ( const PVR_RECORDING &rec )
+{
+  if (m_conn.GetProtocol() < 26)
+    return PVR_ERROR_NOT_IMPLEMENTED;
+
+  /* Seems a bit odd */
```
Jalle19: What does this mean?

src/Tvheadend.cpp
```diff
@@ -1367,7 +1408,7 @@ bool CTvheadend::Connected ( void )
   m_asyncState.SetState(ASYNC_CHN);
   
   msg = htsmsg_create_map();
-  if (Settings::GetInstance().GetAsyncEpg())
+  if (Settings::GetInstance().GetAsyncEpg() || m_conn.GetProtocol() >= 26)
```
Jalle19: You can drop this commit now

src/Tvheadend.cpp
```diff
@@ -598,6 +602,43 @@ PVR_ERROR CTvheadend::RenameRecording ( const PVR_RECORDING &rec )
   return SendDvrUpdate(m);
 }
 
+PVR_ERROR CTvheadend::SetPlayCount ( const PVR_RECORDING &rec, int playcount )
```
Jalle19: You use `playCount` in another place, I suggest to use it here too for consistency

src/Tvheadend.cpp
```diff
@@ -598,6 +602,43 @@ PVR_ERROR CTvheadend::RenameRecording ( const PVR_RECORDING &rec )
...
+  Logger::Log(LogLevel::LEVEL_DEBUG, "Setting playcount to %i for recording %s", playcount, rec.strRecordingId);
+
+  /* Build message */
+  htsmsg_t *m = htsmsg_create_map();
+  htsmsg_add_u32(m, "id",        atoi(rec.strRecordingId));
+  htsmsg_add_u32(m, "playcount", playcount);
+  return SendDvrUpdate(m);
+}
+
+PVR_ERROR CTvheadend::SetPlayPosition ( const PVR_RECORDING &rec, int playposition )
```
Jalle19: You use `playPosition` when calling this method, I suggest you name this the same for consistency

src/Tvheadend.cpp
Glenn-1990: Kodi calls this function to request the play position, although the play position was already transferred to Kodi as a member of PVR_RECORDING.

Glenn-1990:
@Jalle19 @ksooo Updated and removed WIP

Some fixes are preferable for Krypton, should I separate the fixes in another PR apart from the HTSP v26 stuff?

ksooo:
Lets finalize this PR first before opening one for Krypton, to prevent unneeded resync.

ksooo:
I still had no time to take a closer look at your changes.

ksooo:
Have you seen and addressed my comments?

ksooo:
At least CTvheadend::GetPlayPosition needs still to be fixed. 

Will take a closer look at the rest and comment back as soon as i find some time.

Glenn-1990:
@ksooo Yes I addressed you comments about add-on capabilities and timertypes.

ksooo:
What about getplayposition?

Glenn-1990:
@ksooo GetPlayPosition() is working (when on htsp v26 or up) but I'm still not sure if this is how it should be. If you could take a look :-)

ksooo:
GetPlayPosition() should not return what came in with rec.iLastPlayedPosition, but obtain and return latest position stored in backend.

Glenn-1990:
Well as we are working async, rec.iLastPlayedPosition will always hold the correct value. We push an Kodi update every time the position on the backend changes. 
https://github.com/kodi-pvr/pvr.hts/blob/master/src/Tvheadend.cpp#L1895

But ofc we could also use our local playposition (from m_recordings)?

ksooo:
> We push an Kodi update every time the position on the backend changes.

 How often is this in worst case? can this cause performance problems?

> But ofc we could also use our local playposition (from m_recordings)?

 I think we should go for this. "Feels" better.

src/Tvheadend.cpp
```diff
@@ -665,14 +706,20 @@ PVR_ERROR CTvheadend::GetTimerTypes ( PVR_TIMER_TYPE types[], int *size )
 
   /* PVR_Timer.iPreventDuplicateEpisodes values and presentation.*/
   static std::vector< std::pair<int, std::string> > deDupValues;
-  if (deDupValues.size() == 0)
```
Jalle19: Is it okay to remove this check? Won't this mean we may accidentally `push_back` duplicates?

ksooo:
I still want to do a review before we merge this. Will do as soon as I find the time. No nned to rush, imo.

Jalle19:
@ksooo sure

src/Tvheadend.cpp
Glenn-1990: This will not work as we do a version check a few lines below, and htspversion = 0 the first time. 
It's working for the other methods, so I guess Kodi will reject duplicates.

Glenn-1990:
> How often is this in worst case? can this cause performance problems? Not sure we actually should do this, especially as we're getting asked (GetPlayPosition) when kodi needs an update!
@ksooo Only when a user presses "stop" on other clients. 
What's the use of "iLastPlayedPosition" as we have "GetPlayPosition()" ? The API is unclear about it. If we don't need "iLastPlayedPosition", the code can be simplified.

ksooo:
> What's the use of "iLastPlayedPosition" as we have "GetPlayPosition()" ? The API is unclear about it. If we don't need "iLastPlayedPosition", the code can be simplified.

I have not invented this API, but I guess GetPlayPosition() is there to have a lightweight method to be able to directly and quickly (and frequently?) pull the play position data for a particular recording. We only have bulk query methods otherwise and refreshing data for all recordings (GetRecordings) might be too heavyweight.

src/Tvheadend.cpp
Jalle19: IMO it's better to be safe than sorry, we cannot rely on Kodi doing the "right thing". Is there a way to change this to non-static so this thing wouldn't be an issue?

Jalle19:
@Glenn-1990 @ksooo would it perhaps be a good idea to separate the "Fixed" commits to a separate PR while we grind this one?

ksooo:
+1 for doing separate PR for fixes and features. Makes the review easier.

src/client.cpp
```diff
@@ -217,6 +217,8 @@ PVR_ERROR GetAddonCapabilities(PVR_ADDON_CAPABILITIES* pCapabilities)
   pCapabilities->bHandlesInputStream         = true;
   pCapabilities->bHandlesDemuxing            = true;
   pCapabilities->bSupportsRecordingEdl       = true;
+  pCapabilities->bSupportsRecordingPlayCount = tvh->GetProtocol() >= 26 ? true : false;
+  pCapabilities->bSupportsLastPlayedPosition = tvh->GetProtocol() >= 26 ? true : false;
```
ksooo: these two lines look a little bit overcomplicated. Why not just
<pre>
 pCapabilities->bSupportsRecordingPlayCount = (tvh->GetProtocol() >= 26);
 pCapabilities->bSupportsLastPlayedPosition = (tvh->GetProtocol() >= 26);
</pre>

ksooo:
@Glenn-1990 good job, thanks. Could you please add changelog and version bump. We are ready to then, imo.

Jalle19:
+1 from me too if you add the change log entry etc.

Glenn-1990:
Changelog added

Jalle19:
@ksooo good to go?

ksooo:
Yep, good to go.

Does anything speak against a Krypton backport @glenn-1990?

Jalle19:
@ksooo Krypton is still unreleased so everything should be backported to it IMO.

ksooo:
> @ksooo Krypton is still unreleased so everything should be backported to it IMO.

 Yep, but we need to hurry then, because we want to release Krypton really very soon, @Glenn-1990 ;-)

Glenn-1990:
@ksooo A backport would be nice, maybe someone else can do a runtime test first? Especially the play count and play position stuff.

ksooo:
That's problematic, due to lack of time on my end. If we are not sure that it actually works, no backport now, please. 

However, ofc a Krypton PR is welcome. I will test asap.

Glenn-1990:
Under some circumstances the playposition seems to be wrong as Kodi uses "-1" for fully watched. Only affects htsp v26 and up.  I'll come up with a PR, already addressed in the Krypton PR.

PR #261, opened 2016-12-30
4.0.2: Enable async epg transfer by default.

ksooo:


Glenn-1990:
@ksooo this can cause issues with older backends. "epgMaxTime" is not implemented correctly here. This was fixed with commit https://github.com/tvheadend/tvheadend/commit/0de7b5c4ea478821527e66fed45c6bf059cb4e4b

So the available options are:
1) This PR, but users on < htsp v25 will run out epg data after some days.
2) Enable async epg by default but don't set the epg window (epgMaxTime) when connected to older backends. This should not have any side expect beside a small increase in loading time.
3) The auto option like explained in the other PR.

I vote for option 2 :-)

ksooo:
I like option 2)

What do you suggest to detect "older backends"?

Glenn-1990:
@ksooo Something like this?
https://github.com/Glenn-1990/pvr.hts/commit/e6ea7d4d1251774114f12b18f55918e573cb5723

ksooo:
@Glenn-1990 see my latest commit. Is this okay?

src/Tvheadend.cpp
```diff
@@ -2158,6 +2165,23 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
+  // To workaround this, we must request all epg data and filter it here.
+  if (m_conn.GetProtocol() < 25 &&
+      Settings::GetInstance().GetAsyncEpg() &&
+      m_epgMaxDays > EPG_TIMEFRAME_UNLIMITED)
+  {
+    // check event's stop time against m_epgMaxDays, return if to large
+    if (stop > static_cast<int64_t>(time(NULL) + m_epgMaxDays * int64_t(24 * 60 *60)))
+      return false;
+  }
+
```
Glenn-1990: Not parsing the event causes the same problem as the buggy "epgMaxTime". An event will be sent once in async mode, if we ignore an event, it will be missing forever (without restarting the htsp connction). It's better to just ignore the day setting for the affected servers or would this lead in an unlimited tv guide on the Kodi side?

ksooo: Then, this approach is a nogo. People with low power machines use the respective kodi setting to reduce the number of epg events to fetch in order to prevent performance problems (due to low memory, etc.). Ignoring this setting and pushing everything tvheadend sends to kodi is a nogo.

Glenn-1990: I would leave it as it is for now, Especially for Krypton. For v18 we can remove the setting entirely to avoid confusion and incorrect behaviour when enabling the setting for older backends.

ksooo: Okay, I will close the PRs.

PR #260, opened 2016-12-30
Fix libXBMC_codec.h include path

Kwiboo:
This PR changes the include path for `libXBMC_codec.h` to make it easier to build this binary addon when Kodi's app name have been changed.

`${KODI_INCLUDE_DIR}` @ https://github.com/kodi-pvr/pvr.hts/blob/Krypton/CMakeLists.txt#L13 already points to the `kodi` folder and to an app-named include folder in my case.

PR #271, opened 2017-01-04
Fix libXBMC_pvr.h include path

Kwiboo:
Same as #260 but for kodi-agile branch


FernetMenta:
no need for any other branch than Krypton, because addon system is completely reworked and this change is obsolete.

Kwiboo:
Okay, closing :)

PR #272, opened 2017-01-04
Store streams in a vector instead of a C array

### LINKED ISSUE #204: Use PVR_STREAM_PROPERTIES directly
Now that consecutive order if streams is no longer mandatory we should stop using the wrapper for it and just push the streams directly to Kodi.

Related: https://github.com/kodi-pvr/pvr.vdr.vnsi/pull/60

### LINKED ISSUE #264: Crash when too many subtitle streams
I was looking into crash reported here:
https://discourse.osmc.tv/t/tv-channel-crashes/21206/15

I think this is the key bit of info:
> Is it possible that the channel has too many streams in it (24)? There are many subtitle streams (16-18).
I've played around with TVHeadend and figured out, that the problem is indeed the large number of Subtitle tracks. If I filter some of them out on the server side, then Kodi will play it nicely.
It is not a problem when recording to a file, because TVHeadend detects that there is no more stream space available and trims some of the subtitle streams.

I can reproduce issue and crash occurs in pvr.hts (built from current master of this repo):
```Program terminated with signal SIGSEGV, Segmentation fault.
#0  _M_lower_bound (this=<optimized out>, __k=@0x619feba0: 94210, __y=0xee2b5a06, __x=<optimized out>)
    at /home/dc4/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/c++/4.9.3/bits/stl_tree.h:1261
1261	/home/dc4/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/c++/4.9.3/bits/stl_tree.h: No such file or directory.
(gdb) bt
#0  _M_lower_bound (this=<optimized out>, __k=@0x619feba0: 94210, __y=0xee2b5a06, __x=<optimized out>)
    at /home/dc4/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/c++/4.9.3/bits/stl_tree.h:1261
#1  lower_bound (__k=@0x619feba0: 94210, this=0x7460bb0c) at /home/dc4/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/c++/4.9.3/bits/stl_tree.h:927
#2  lower_bound (__x=@0x619feba0: 94210, this=0x7460bb0c) at /home/dc4/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/c++/4.9.3/bits/stl_map.h:902
#3  operator[] (__k=<unknown type in /opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/addons/pvr.hts/pvr.hts.so.4.0.1, CU 0x3a553, DIE 0x54955>, this=0x7460bb0c)
    at /home/dc4/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/c++/4.9.3/bits/stl_map.h:516
#4  CHTSPDemuxer::ParseSubscriptionStart (this=this@entry=0x7460b5b0, m=0x17002, m@entry=0x68416880) at /home/dc4/xbmc_holder/pvr.hts/src/HTSPDemuxer.cpp:456
#5  0x70832e50 in CHTSPDemuxer::ProcessMessage (this=this@entry=0x7460b5b0, method=method@entry=0x67332980 "subscriptionStart", m=m@entry=0x68416880)
    at /home/dc4/xbmc_holder/pvr.hts/src/HTSPDemuxer.cpp:336
#6  0x7083a0b4 in CTvheadend::ProcessMessage (this=<optimized out>, method=method@entry=0x67332980 "subscriptionStart", msg=msg@entry=0x68416880) at /home/dc4/xbmc_holder/pvr.hts/src/Tvheadend.cpp:1406
#7  0x70827bc8 in CHTSPConnection::ReadMessage (this=0x92c, this@entry=0x74603c18) at /home/dc4/xbmc_holder/pvr.hts/src/HTSPConnection.cpp:324
#8  0x708291f4 in CHTSPConnection::Process (this=0x74603c18) at /home/dc4/xbmc_holder/pvr.hts/src/HTSPConnection.cpp:650
#9  0x7082cf44 in P8PLATFORM::CThread::ThreadHandler (_thread=0x74603c18) at /opt/xbmc-cmake/arm-linux-gnueabihf/include/p8-platform/threads/threads.h:65
#10 0x76dd4e90 in start_thread (arg=0x619ff410) at pthread_create.c:311
```
(this is the line m_streamStat[idx] = 0;)

I added some logging to see how far it gets: http://paste.ubuntu.com/23729089/
```
Abort0: 0/0/178956970
ParseSubscriptionStart: clear: 0/0/178956970
ParseSubscriptionStart: zero: 2/0/178956970
ParseSubscriptionStart: zero: 1/1/178956970
ParseSubscriptionStart: zero: 3/2/178956970
ParseSubscriptionStart: zero: 4/3/178956970
ParseSubscriptionStart: zero: 5/4/178956970
ParseSubscriptionStart: zero: 6/5/178956970
ParseSubscriptionStart: zero: 7/6/178956970
ParseSubscriptionStart: zero: 8/7/178956970
ParseSubscriptionStart: zero: 9/8/178956970
ParseSubscriptionStart: zero: 10/9/178956970
ParseSubscriptionStart: zero: 11/10/178956970
ParseSubscriptionStart: zero: 12/11/178956970
ParseSubscriptionStart: zero: 13/12/178956970
ParseSubscriptionStart: zero: 14/13/178956970
ParseSubscriptionStart: zero: 15/14/178956970
ParseSubscriptionStart: zero: 16/15/178956970
ParseSubscriptionStart: zero: 17/16/178956970
ParseSubscriptionStart: zero: 18/17/178956970
ParseSubscriptionStart: zero: 19/18/178956970
ParseSubscriptionStart: zero: 20/19/178956970
ParseSubscriptionStart: zero: 21/20/178956970
ParseSubscriptionStart: zero: 22/0/178956970
```
It feels like there is some limit of around 20 streams before memory gets corrupted.

Jalle19:
Also cap the streams at PVR_STREAMS_MAX_STREAMS. 

Closes #204, fixes #264.

@ksooo for review
@popcornmix @gkovacsp this should fix the issue you reported, can you test it?

### DIFF
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -74,7 +74,7 @@ void CHTSPDemuxer::Close0 ( void )
 void CHTSPDemuxer::Abort0 ( void )
 {
   CLockObject lock(m_mutex);
-  m_streams.iStreamCount = 0;
+  m_streams.clear();
   m_streamStat.clear();
   m_seeking = false;
   m_speedChange = false;
@@ -210,12 +210,13 @@ void CHTSPDemuxer::Weight ( enum eSubscriptionWeight weight )
 PVR_ERROR CHTSPDemuxer::CurrentStreams ( PVR_STREAM_PROPERTIES *props )
 {
   CLockObject lock(m_mutex);
-  for (int i = 0; i < m_streams.iStreamCount; i++)
+
+  for (size_t i = 0; i < m_streams.size(); i++)
   {
-    memcpy(&props->stream[i], &m_streams.stream[i], sizeof(PVR_STREAM_PROPERTIES::PVR_STREAM));
+    memcpy(&props->stream[i], &m_streams.at(i), sizeof(PVR_STREAM_PROPERTIES::PVR_STREAM));
   }
 
-  props->iStreamCount = m_streams.iStreamCount;
+  props->iStreamCount = static_cast<unsigned int>(m_streams.size());
   return PVR_ERROR_NO_ERROR;
 }
 
@@ -436,10 +437,11 @@ void CHTSPDemuxer::ParseSubscriptionStart ( htsmsg_t *m )
     return;
   }
   m_streamStat.clear();
-  m_streams.iStreamCount = 0;
+  m_streams.clear();
+
+  Logger::Log(LogLevel::LEVEL_DEBUG, "demux subscription start");
 
   /* Process each */
-  int count = 0;
   HTSMSG_FOREACH(f, l)
   {
     uint32_t   idx, u32;
@@ -454,82 +456,87 @@ void CHTSPDemuxer::ParseSubscriptionStart ( htsmsg_t *m )
 
     /* Find stream */
     m_streamStat[idx] = 0;
+    PVR_STREAM_PROPERTIES::PVR_STREAM stream = {};
+
     Logger::Log(LogLevel::LEVEL_DEBUG, "demux subscription start");
     
     CodecDescriptor codecDescriptor = CodecDescriptor::GetCodecByName(type);
     xbmc_codec_t codec = codecDescriptor.Codec();
 
-    memset(&m_streams.stream[count], 0, sizeof(m_streams.stream[count]));
-
     if (codec.codec_type != XBMC_CODEC_TYPE_UNKNOWN)
     {
-      m_streams.stream[count].iCodecType = codec.codec_type;
-      m_streams.stream[count].iCodecId   = codec.codec_id;
-      m_streams.stream[count].iPID       = idx;
+      stream.iCodecType = codec.codec_type;
+      stream.iCodecId   = codec.codec_id;
+      stream.iPID       = idx;
 
       /* Subtitle ID */
-      if ((m_streams.stream[count].iCodecType == XBMC_CODEC_TYPE_SUBTITLE) &&
+      if ((stream.iCodecType == XBMC_CODEC_TYPE_SUBTITLE) &&
           !strcmp("DVBSUB", type))
       {
         uint32_t composition_id = 0, ancillary_id = 0;
         htsmsg_get_u32(&f->hmf_msg, "composition_id", &composition_id);
         htsmsg_get_u32(&f->hmf_msg, "ancillary_id"  , &ancillary_id);
-        m_streams.stream[count].iSubtitleInfo = (composition_id & 0xffff)
-                                              | ((ancillary_id & 0xffff) << 16);
+        stream.iSubtitleInfo = (composition_id & 0xffff) | ((ancillary_id & 0xffff) << 16);
       }
 
       /* Language */
-      if (m_streams.stream[count].iCodecType == XBMC_CODEC_TYPE_SUBTITLE ||
-          m_streams.stream[count].iCodecType == XBMC_CODEC_TYPE_AUDIO)
+      if (stream.iCodecType == XBMC_CODEC_TYPE_SUBTITLE ||
+          stream.iCodecType == XBMC_CODEC_TYPE_AUDIO)
       {
         const char *language;
         
         if ((language = htsmsg_get_str(&f->hmf_msg, "language")) != NULL)
-          strncpy(m_streams.stream[count].strLanguage, language, sizeof(m_streams.stream[count].strLanguage) - 1);
+          strncpy(stream.strLanguage, language, sizeof(stream.strLanguage) - 1);
       }
 
       /* Audio data */
-      if (m_streams.stream[count].iCodecType == XBMC_CODEC_TYPE_AUDIO)
+      if (stream.iCodecType == XBMC_CODEC_TYPE_AUDIO)
       {
-        m_streams.stream[count].iChannels
-          = htsmsg_get_u32_or_default(&f->hmf_msg, "channels", 2);
-        m_streams.stream[count].iSampleRate
-          = htsmsg_get_u32_or_default(&f->hmf_msg, "rate", 48000);
+        stream.iChannels = htsmsg_get_u32_or_default(&f->hmf_msg, "channels", 2);
+        stream.iSampleRate = htsmsg_get_u32_or_default(&f->hmf_msg, "rate", 48000);
       }
 
       /* Video */
-      if (m_streams.stream[count].iCodecType == XBMC_CODEC_TYPE_VIDEO)
+      if (stream.iCodecType == XBMC_CODEC_TYPE_VIDEO)
       {
-        m_streams.stream[count].iWidth  = htsmsg_get_u32_or_default(&f->hmf_msg, "width", 0);
-        m_streams.stream[count].iHeight = htsmsg_get_u32_or_default(&f->hmf_msg, "height", 0);
+        stream.iWidth  = htsmsg_get_u32_or_default(&f->hmf_msg, "width", 0);
+        stream.iHeight = htsmsg_get_u32_or_default(&f->hmf_msg, "height", 0);
         
         /* Ignore this message if the stream details haven't been determined 
            yet, a new message will be sent once they have. This is fixed in 
            some versions of tvheadend and is here for backward compatibility. */
-        if (m_streams.stream[count].iWidth == 0 || m_streams.stream[count].iHeight == 0)
+        if (stream.iWidth == 0 || stream.iHeight == 0)
         {
           Logger::Log(LogLevel::LEVEL_DEBUG, "Ignoring subscriptionStart, stream details missing");
           return;
         }
         
         /* Setting aspect ratio to zero will cause XBMC to handle changes in it */
-        m_streams.stream[count].fAspect = 0.0f;
+        stream.fAspect = 0.0f;
         
         if ((u32 = htsmsg_get_u32_or_default(&f->hmf_msg, "duration", 0)) > 0)
         {
-          m_streams.stream[count].iFPSScale = u32;
-          m_streams.stream[count].iFPSRate  = DVD_TIME_BASE;
+          stream.iFPSScale = u32;
+          stream.iFPSRate  = DVD_TIME_BASE;
         }
       }
-        
-      Logger::Log(LogLevel::LEVEL_DEBUG, "  id: %d, type %s, codec: %u", idx, type, m_streams.stream[count].iCodecId);
-      count++;
+
+      /* We can only use PVR_STREAM_MAX_STREAMS streams */
+      if (m_streams.size() < PVR_STREAM_MAX_STREAMS)
+      {
+        Logger::Log(LogLevel::LEVEL_DEBUG, "  id: %d, type %s, codec: %u", idx, type, stream.iCodecId);
+        m_streams.push_back(stream);
+      }
+      else
+      {
+        Logger::Log(LogLevel::LEVEL_INFO, "Maximum stream limit reached ignoring id: %d, type %s, codec: %u", idx, type,
+                    stream.iCodecId);
+      }
     }
   }
 
   /* Update streams */
   Logger::Log(LogLevel::LEVEL_DEBUG, "demux stream change");
-  m_streams.iStreamCount = count;
   pkt = PVR->AllocateDemuxPacket(0);
   pkt->iStreamId = DMX_SPECIALID_STREAMCHANGE;
   m_pktBuffer.Push(pkt);
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.0.1"
+  version="4.0.2"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,9 @@
+4.0.2
+- updated build instructions in the README
+- fixed Travis CI and AppVeyor automated builds
+- fixed a bunch of cast warnings
+- cleanup stream handling somewhat
+
 4.0.1
 - updated language files from Transifex
 
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+4.0.1
+- updated language files from Transifex
+
 4.0.0
 - Initial Kodi v18 version
 
--- a/src/Tvheadend.h
+++ b/src/Tvheadend.h
@@ -256,7 +256,7 @@ private:
   mutable P8PLATFORM::CMutex              m_mutex;
   CHTSPConnection                        &m_conn;
   P8PLATFORM::SyncedBuffer<DemuxPacket*>  m_pktBuffer;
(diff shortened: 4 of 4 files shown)


src/HTSPDemuxer.cpp
```diff
@@ -454,82 +458,80 @@ void CHTSPDemuxer::ParseSubscriptionStart ( htsmsg_t *m )
 
     /* Find stream */
     m_streamStat[idx] = 0;
+    PVR_STREAM_PROPERTIES::PVR_STREAM stream;
```
ksooo: You need to zero memset <code>stream</code>, or is it guaranteed that every member will be set in every iteration? Even if this is the case, I would memset to be on the safe side, for instance if the struct gets extended in the future.

src/HTSPDemuxer.cpp
```diff
@@ -210,12 +210,16 @@ void CHTSPDemuxer::Weight ( enum eSubscriptionWeight weight )
 PVR_ERROR CHTSPDemuxer::CurrentStreams ( PVR_STREAM_PROPERTIES *props )
 {
   CLockObject lock(m_mutex);
-  for (int i = 0; i < m_streams.iStreamCount; i++)
+
+  /* Truncate the list of streams to prevent overflows */
+  m_streams.resize(PVR_STREAM_MAX_STREAMS);
+
+  for (size_t i = 0; i < m_streams.size() && i < PVR_STREAM_MAX_STREAMS; i++)
```
ksooo: The <code>resize</code> some lines above is imo not needed, as you already prevent overflows here.

src/HTSPDemuxer.cpp
```diff
@@ -210,12 +210,16 @@ void CHTSPDemuxer::Weight ( enum eSubscriptionWeight weight )
...
   {
-    memcpy(&props->stream[i], &m_streams.stream[i], sizeof(PVR_STREAM_PROPERTIES::PVR_STREAM));
+    memcpy(&props->stream[i], &m_streams.at(i), sizeof(PVR_STREAM_PROPERTIES::PVR_STREAM));
   }
 
-  props->iStreamCount = m_streams.iStreamCount;
+  props->iStreamCount = static_cast<unsigned int>(m_streams.size());
```
ksooo: If you remove the <code>resize</code> like I suggested, you need to adjust this line as well! 
=> <code>m_streams.size() <= PVR_STREAM_MAX_STREAMS ? m_streams.size() : PVR_STREAM_MAX_STREAMS </code>

src/HTSPDemuxer.cpp
```diff
@@ -454,82 +458,80 @@ void CHTSPDemuxer::ParseSubscriptionStart ( htsmsg_t *m )
...
-          m_streams.stream[count].iFPSScale = u32;
-          m_streams.stream[count].iFPSRate  = DVD_TIME_BASE;
+          stream.iFPSScale = u32;
+          stream.iFPSRate  = DVD_TIME_BASE;
         }
       }
         
-      Logger::Log(LogLevel::LEVEL_DEBUG, "  id: %d, type %s, codec: %u", idx, type, m_streams.stream[count].iCodecId);
+      Logger::Log(LogLevel::LEVEL_DEBUG, "  id: %d, type %s, codec: %u", idx, type, stream.iCodecId);
       count++;
```
ksooo: The variable <code>count</code> can be removed. You're not doing anything useful with it anymore. ;-)

src/HTSPDemuxer.cpp
Jalle19: The loop condition is a leftover from before I found resize(), I'll drop it.

src/HTSPDemuxer.cpp
Jalle19: Isn't a newly allocated struct guaranteed to be zero?

src/HTSPDemuxer.cpp
Jalle19: Indeed

src/HTSPDemuxer.cpp
ksooo: Why the resize? Why do you put more than elements PVR_STREAM_MAX_STREAM into the vector, solely for dropping them later? If you do never add > PVR_STREAM_MAX_STREAMS to the vector, you do not need to resize and you do not need to check anything else than size() here.

src/HTSPDemuxer.cpp
ksooo: No (in this case), as you did not specify an initilaizer. => http://stackoverflow.com/questions/1069621/are-members-of-a-c-struct-initialized-to-0-by-default

src/HTSPDemuxer.cpp
Jalle19: Okay, I'll change it like that instead, then we can also log streams that were ignored.

Jalle19:
@ksooo updated

popcornmix:
Okay, the crash has moved with this PR:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  CDVDFileInfo::DemuxerToStreamDetails (pInputStream=0x7ed03a20, pInputStream@entry=0x73d10098, pDemux=0x0, pDemux@entry=0x73d06098, details=..., 
    path=<error reading variable: Cannot access memory at address 0xfffffff4>) at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/DVDFileInfo.cpp:392
#1  0x006c01f8 in CDVDFileInfo::DemuxerToStreamDetails (pInputStream=0x73d10098, pDemuxer=0x73d06098, subs=std::vector of length 0, capacity 0, details=...)
    at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/DVDFileInfo.cpp:372
#2  0x006db004 in CVideoPlayer::GetStreamDetails (this=0x55d3148, details=...) at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/VideoPlayer.cpp:5237
#3  0x00bb3fa4 in CApplicationPlayer::GetStreamDetails (this=<optimized out>, details=...) at /home/dc4/xbmc_holder/xbmc/xbmc/ApplicationPlayer.cpp:502
#4  0x00ba6ac8 in CApplication::UpdateFileState (this=this@entry=0x4408160) at /home/dc4/xbmc_holder/xbmc/xbmc/Application.cpp:3814
#5  0x00baf1b0 in CApplication::ProcessSlow (this=this@entry=0x4408160) at /home/dc4/xbmc_holder/xbmc/xbmc/Application.cpp:4548
#6  0x00baf614 in CApplication::Process (this=0x4408160) at /home/dc4/xbmc_holder/xbmc/xbmc/Application.cpp:4514
#7  0x00c5da44 in CXBApplicationEx::Run (this=0x4408160, playlist=...) at /home/dc4/xbmc_holder/xbmc/xbmc/XBApplicationEx.cpp:94
#8  0x0090f2bc in XBMC_Run (renderGUI=<optimized out>, playlist=...) at /home/dc4/xbmc_holder/xbmc/xbmc/platform/xbmc.cpp:89
#9  0x00456eec in main (argc=5, argv=0x7ed04134) at /home/dc4/xbmc_holder/xbmc/xbmc/platform/posix/main.cpp:79
```
Does Kodi itself have a similar issue with more than 20 streams?

popcornmix:
Also seen it crash here
```
#0  CSelectionStreams::Update (this=0x617fe998, this@entry=0x4fcbee8, input=input@entry=0x73c0dfd0, demuxer=0x4fcbee8, demuxer@entry=0x73c13e48, filename2="")
    at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/VideoPlayer.cpp:570
#1  0x006da768 in CVideoPlayer::CheckStreamChanges (this=this@entry=0x4fcb6c0, current=..., stream=0x73c0cab0) at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/VideoPlayer.cpp:1767
#2  0x006daa50 in CVideoPlayer::ProcessVideoData (this=0x4fcb6c0, pStream=<optimized out>, pPacket=0x72a59248) at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/VideoPlayer.cpp:1818
#3  0x006e12f0 in CVideoPlayer::Process (this=0x4fcb6c0) at /home/dc4/xbmc_holder/xbmc/xbmc/cores/VideoPlayer/VideoPlayer.cpp:1688
#4  0x008dab00 in CThread::Action (this=0x4fcb6d0) at /home/dc4/xbmc_holder/xbmc/xbmc/threads/Thread.cpp:221
#5  0x008dadcc in CThread::staticThread (data=0x4fcb6d0) at /home/dc4/xbmc_holder/xbmc/xbmc/threads/Thread.cpp:131
#6  0x76dd5e90 in start_thread (arg=0x617ff410) at pthread_create.c:311
#7  0x7610d598 in ?? () at ../ports/sysdeps/unix/sysv/linux/arm/nptl/../clone.S:92 from /lib/arm-linux-gnueabihf/libc.so.6
```
It seems ``for (auto stream : demuxer->GetStreams())`` is returning a ``stream==nullptr`` which isn't liked.

Jalle19:
I think I know what's going on, but I have to think about how we'll actually fix that.

Jalle19:
@popcornmix can you post a full trace from all threads?

Jalle19:
And preferably a debug log from Kodi too so I can see if the streams are actually getting ignored.

Jalle19:
@popcornmix I don't think it fixes anything but can you try the latest commit?

popcornmix:
(without new commit)
Log: http://paste.debian.net/906325/
gdb is crashing on some of the threads backtraces. Here is a shallow backtrace - I can get deeper ones from specific tasks if you need them.
http://paste.debian.net/906324/

popcornmix:
No change to crash with new commit

Jalle19:
@popcornmix I can't find this commit anywhere so the line numbers don't make any sense. Do you have a link?

popcornmix:
By new commit I mean: https://github.com/kodi-pvr/pvr.hts/pull/272/commits/dfe9bbb65096fdfd833996118a32e87a84c46ce8

popcornmix:
The kodi tree I am using is https://github.com/popcornmix/xbmc/tree/newclock5

Jalle19:
@ksooo you're probably right, although I think it's the right thing to do either way

ksooo:
@Jalle19 just for testing, replace the initializer with memset, please. => https://github.com/kodi-pvr/pvr.hts/pull/272/files#diff-fc67c1aab2e0d4a225199143a1d4be69R457 I'm not sure the initializer is called on every iteration.

Jalle19:
The contents of `GetStreams()` comes from https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp#L337, where shared pointers are turned into naked pointers. `m_streams` is updated here: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp#L538 and cleared in a few other places, so it could be that the pointer gets deleted behind the back.

Jalle19:
@popcornmix @ksooo added `memset` call

Jalle19:
another another logging statement too

popcornmix:
No change with latest PR (inc memset and logging)
http://paste.debian.net/906328/

Jalle19:
I'm wondering if the "packet doesn't belong to a valid stream" is related

ksooo:
> No change with latest PR (inc memset and logging)

 Yeah, that's what I thought, just wanted to make 100% sure memset does not make a difference. 

 OTOH, I'm clueless what's going wrong. IMO, the changes @Jalle19 did are looking good.

popcornmix:
Might be worth taking @gkovacsp up on his offer so you can see the issue directly.
He has a tvheadend backend. Select the travel channel and I suspect you'll see the problem.

gkovacsp:
Use this config to access the server:

`<settings>
    <setting id="autorec_approxtime" value="0" />
    <setting id="autorec_maxdiff" value="15" />
    <setting id="connect_timeout" value="10" />
    <setting id="dvr_dubdetect" value="0" />
    <setting id="dvr_lifetime" value="8" />
    <setting id="dvr_priority" value="2" />
    <setting id="epg_async" value="false" />
    <setting id="host" value="address" />
    <setting id="htsp_port" value="15002" />
    <setting id="http_port" value="15001" />
    <setting id="pass" value="osmc" />
    <setting id="pretuner_closedelay" value="10" />
    <setting id="pretuner_enabled" value="false" />
    <setting id="response_timeout" value="5" />
    <setting id="streaming_profile" value="" />
    <setting id="total_tuners" value="2" />
    <setting id="trace_debug" value="false" />
    <setting id="user" value="osmc" />
</settings>`

Jalle19:
@popcornmix @gkovacsp none of those channels (4 in total visible) contain that many streams so I can't reproduce (https://paste.ubuntu.com/23739729/)

popcornmix:
Something has changed - I was seeing around 100 channels and the travel channel was the only one with the bug. I'm also seeing 4 channels currently.

gkovacsp:
Sorry, I've messed up something, Travel is back on the list.

popcornmix:
Yes, I can provoke the crash with "Travel" channel again.

ksooo:
The minimalistic fix seems to be:

<pre>
diff --git a/src/HTSPDemuxer.cpp b/src/HTSPDemuxer.cpp
index 12efa20..a65507c 100644
--- a/src/HTSPDemuxer.cpp
+++ b/src/HTSPDemuxer.cpp
@@ -524,6 +524,9 @@ void CHTSPDemuxer::ParseSubscriptionStart ( htsmsg_t *m )
         
       Logger::Log(LogLevel::LEVEL_DEBUG, "  id: %d, type %s, codec: %u", idx, type, m_streams.stream[count].iCodecId);
       count++;
+
+      if (count == PVR_STREAM_MAX_STREAMS)
+        break;
     }
   }

</pre>

I applied this to current pvr.hts master (without the changes done in this PR) and 'Travel' channel works.

popcornmix:
Yes, can confirm minimal patch avoids the crash on Travel for me too.

ksooo:
> Yes, can confirm minimal patch avoids the crash on Travel for me too.

 well, and it actually makes the "Travel" channel work, right?

popcornmix:
Yes, travel channel is working for me. Audio video and subs are fine.
(although presumably there are some languages that didn't make the cut).

Jalle19:
Yeah I guess, I would like to get to the bottom of why my version doesn't work though.

Jalle19:
Squashed the fixups and dropped the temporary stuff. @ksooo can you do a new review?

ksooo:
Sorry, overlooked that we have a fix already.

I think this fix justifies an addon version bump and a respective changelog entry. ;-)

Jalle19:
@ksooo the real fix was done in Kodi so I changed the label on this one from Fix to Cleanup. Also added the changelog entry.

Good to go?

I would like to get the frame sanitization crap merged too before we start backporting the latest changes to Krypton. Is that okay with you? The sanitization PR stalled previously because of this very same crash.

ksooo:
Thanks @Jalle19. Here we go...

PR #275, opened 2017-01-17
Some fixes

Glenn-1990:
@ksooo @Jalle19 
Fixes in a separate PR now.


ksooo:
nice. will review as soon as i find some time.

Jalle19:
In that case +1. @ksooo?

src/HTSPDemuxer.cpp
```diff
@@ -309,13 +309,8 @@ bool CHTSPDemuxer::IsRealTimeStream() const
...
 
-  if (m_timeshiftStatus.shift == 0)
-    return true;
-
-  if (m_timeshiftStatus.start - m_timeshiftStatus.shift < 10)
-    return true;
-
-  return false;
+  /* Handle as real time when reading close to the EOF (10000000�s - 10s) */
+  return (m_timeshiftStatus.shift < 10000000);
```
ksooo: I'm not at all sure about this change. A stream that is 10 secs off real time is considered 'real time'? Really? What does this change fix?

@FernetMenta what's your take on this?

FernetMenta: demux buffers are 8 secs so if you have less than this in your timeshift buffer, kodi's demux buffers can't fill. I don't know what the variables here mean and what the units are. vnsi considers a stream realtime if less that 10 secs are in the buffer.

Glenn-1990: @ksooo Now pvr.hts should behave the same as pvr.vnsi. When a stream is "real time" kodi applies an other A/V sync method if I'm right. Not using this method can lead to A/V sync issues when timeshifting close to the live point. Also tvheadends "shift" is equal to "end time - play time" in vdr.
https://github.com/kodi-pvr/pvr.vdr.vnsi/blob/Krypton/src/client.cpp#L870

ksooo: Okay, then. Thanks for explaining.

Glenn-1990:
@ksooo changelog added
Haven't included the "speed is signed" commit in the changelog as this is cosmetic only.

ksooo:
Nice. Thanks.

Could you maybe also create the corresponding Krypton PR (and also for your last merged PR for which we also do not have a Krypton backport yet, iirc )?

Glenn-1990:
Hmm, I'm having issues with pushing commits to the Krypton branch after updating git tools, I'll see if I can fix this later this week.

PR #277, opened 2017-01-23
Fix CTvheadend::DemuxIsTimeShifting()

ksooo:
Fix CTvheadend::DemuxIsTimeShifting() after #275

@Glenn-1990 okay?


Glenn-1990:
Can't believe I missed this. The fix is correct, thanks :-)
@ksooo I'm curious how you discovered this as estuary does not have a timeshift GUI.

ksooo:
Krypton has, master does not (yet), but will with next Estuary sync PR.

ksooo:
Runtime-tested. Seems to do now what it is supposed to do.

PR #276, opened 2017-01-22
Backport of #275

Glenn-1990:
@ksooo @Jalle19

PR #278, opened 2017-01-23
Krypton: Fix CTvheadend::DemuxIsTimeShifting()

ksooo:
Fix CTvheadend::DemuxIsTimeShifting() after #276

@Glenn-1990 okay?


Jalle19:
Good catch, not sure how we missed this.

ksooo:
Runtime-tested. Seems to do now what it is supposed to do.

PR #279, opened 2017-01-30
Krypton HTSP v26 support

Glenn-1990:
Partially backport of https://github.com/kodi-pvr/pvr.hts/pull/257
@ksooo @Jalle19

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="3.4.16"
+  version="3.4.17"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,7 @@
+3.4.17
+- added HTSP v26 support 
+- added server based playstatus for recordings (HTSP v26 and up)
+
 3.4.16
 - fix CTvheadend::DemuxIsTimeShifting
 
--- a/src/HTSPConnection.cpp
+++ b/src/HTSPConnection.cpp
@@ -495,6 +495,7 @@ bool CHTSPConnection::SendHello ( void )
 bool CHTSPConnection::SendAuth
   ( const std::string &user, const std::string &pass )
 {
+  uint32_t u32;
   htsmsg_t *msg = htsmsg_create_map();
   htsmsg_add_str(msg, "username", user.c_str());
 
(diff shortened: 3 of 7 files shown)


Jalle19:
@ksooo good to go?

ksooo:
Without runtime test, no. Sorry. Personally I don't have the time for this, currently. 

By all means, I do want to avoid merging followed by forced rush to fix regressions.

Jalle19:
Sure thing, let's wait. I should upgrade to Krypton at home at some point so I can test this properly too.

ksooo:
@Glenn-1990 i just noticed on my tvheadend server (4.1.2415), that my recordings have play counts > 0, although I never used them with pvr.hts containing server-side playcount support? It seems, whenever I start playing a recording (even if I play it only for a second), the counter gets increased by one. Is this automatically done by tvheadend? This seems to be wrong. Imo, the counter should never be increased by tvheadend itself, because there is no way of synchronizing the automatically generated counts with the client supplied counts. What do you think?

Glenn-1990:
Yes, that's done automatically by tvheadend, this was changed with commit: https://github.com/tvheadend/tvheadend/commit/fb5ed85435394164027df6a2d9b09d49e93b706b. 
I guess @perexg added this in order to be consistent with other clients and the webinterface?

ksooo:
> Yes, that's done automatically by tvheadend, this was changed with commit: tvheadend/tvheadend@fb5ed85.
I guess @perexg added this in order to be consistent with other clients and the webinterface?

 If you ask me, this renders the functionality useless. :-(

Glenn-1990:
@ksooo There is also room for improvement in tvheadend but it all depends on what we call the correct behaviour. 

Tvheadend increases the playcount even after playing even 1 second, this is the biggest problem I guess? Or the fact that this is done automatically?

Changing the playcount in tvheadend is user based, so If the user watches a recording in no matter what client, the recording will be marked as watched. This is desired from the server point of view as the user actually watched the recording. The webui will show the recording as watched, so if Kodi will do the same, they will be in sync :-).

> This seems to be wrong. Imo, the counter should never be increased by tvheadend itself, because there is no way of synchronizing the automatically generated counts with the client supplied counts.
Can you explain this a bit more, why synchronizing? Playcount updates from the backend are alredy forwarded to Kodi.

I'm wondering If it's possible to only use server side playposition and not playcount?
These are 2 separate support bools in the PVR API.

ksooo:
> Changing the playcount in tvheadend is user based, so If the user watches a recording in no matter what client, the recording will be marked as watched.

 This clashes with kodi's current implementation. Example: user starts playing a recording thru kodi, one second later tvh increases the playcount, let's say to 1. User continues watching the recording in kodi to the end. After playback finished, kodi increases playcount and sends the new count to tvh, which increases server-side playcount to 2, which is wrong.

Glenn-1990:
@ksooo I tested on that behaviour some time ago and that was not the case. 
I believe it was something like this:
1) User starts playing recording
2) Kodi caches the current playcount
3) Tvheadend increases the playcount
4) User stops watching and Kodi saves the new playcount (cached value + 1), at this moment this is a NOP as both values are the same.

I believe that there is something wrong with the Kodi part of playcount as well as Kodi calls the playcount function on the start of playback as well when playback ends. Sometimes Kodi doesn't even increase the number while the function is called. Seems to be related to when either stop is pressed or EOF is reached. This is with master. I'll see if I can dig a bit deeper into this. 

Another idea is to let tvheadend do the playcount+1 and igore the playcount+1 (except from mark watched/unwatched) from Kodi.

ksooo:
Cannot say anything more accurate, because I have not debugged it. I did some changes to the playcount logic in Kodi for v18, no functional change intended, but maybe I broke something, don't know. Unfortunately I do not have enough time to deal with this, atm.

> 
    User starts playing recording
    Kodi caches the current playcount
    Tvheadend increases the playcount
    User stops watching and Kodi saves the new playcount (cached value + 1), at this moment this is a NOP as both values are the same.

 Relying on exactly this behavior of kodi is very error-prone. Maybe I recently introduced one call to obtain server-side playcount while playing the recording ... and things break!

 Again, either server or kodi must increase the count, but never both. This cannot work! It is a conceptional problem which cannot be hacked away.

Glenn-1990:
@ksooo Kodi says "set watched" but the client gets playcount "0", this might explain the non increasing playcount. This is when clicking on "mark as watched" in the GUI. 
`21:58:25.207 T:139904437470976   DEBUG: CPVRRecordings - ChangeRecordingsPlayCount - will set watched for 1 items
21:58:25.207 T:139904437470976   DEBUG: CPVRRecordings - ChangeRecordingsPlayCount - setting watched for item 0
21:58:25.207 T:139904437470976   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - Setting play count to 0 for recording 1557456253`

I'll look into a way to properly implement playcount.

ksooo:
> I'll look into a way to properly implement playcount.

 Imo, you can't (!) do it right as long as both client and server are increasing the count for the same playback of one recording. I think the whole concept is flawed and all you can come up with are - sorry guy - hacks.

Glenn-1990:
@ksooo I can If tvheadend allows clients to handle the playcount ;-) At this moment tvheadend is inconsistent as it handles playcount by itself and playposition not. It was never my intention to make it that way btw :-)

ksooo:
> I can If tvheadend allows clients to handle the playcount

 That's what I'm saying. Stop tvh from modifying the playcount on it's own and we are fine.

Jalle19:
If something doesn't work, revert it. We can figure out a proper implementation once we have reached consensus on how we'd like tvheadend to work (ideally).

ksooo:
master: yep, we should revert the respective commit (for now).

Krypton: nothing to revert as the commit in question is part of this PR. we must take care not to accidentally merge this.

Goal should be to first fix tvheadend, then we can look what is missing in Kodi and pvr.hts

Glenn-1990:
@ksooo @Jalle19 Feel free to comment in my wip branch here: 
https://github.com/Glenn-1990/tvheadend/commit/cecfc1164b532e1c47ab90767442d94842fab7d8

I'm okay with the revert as I don't want to rush with tvheadend changes.

ksooo:
@Glenn-1990 Could you please come up with a PR containing the revert (+ addon version bump)? Thanks.

ksooo:
@Glenn-1990 can we close this one?

Glenn-1990:
@ksooo Yes, forgot this one

PR #284, opened 2017-02-05
3.4.17: fixed duplicate 'prevent duplicate episodes' timer settings v…

ksooo:
…alues.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -664,7 +664,8 @@ PVR_ERROR CTvheadend::GetTimerTypes ( PVR_TIMER_TYPE types[], int *size )
   }
 
   /* PVR_Timer.iPreventDuplicateEpisodes values and presentation.*/
-  static std::vector< std::pair<int, std::string> > deDupValues;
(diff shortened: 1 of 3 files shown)


src/Tvheadend.cpp
```diff
@@ -664,7 +664,8 @@ PVR_ERROR CTvheadend::GetTimerTypes ( PVR_TIMER_TYPE types[], int *size )
   }
 
   /* PVR_Timer.iPreventDuplicateEpisodes values and presentation.*/
-  static std::vector< std::pair<int, std::string> > deDupValues;
+  std::vector< std::pair<int, std::string> > deDupValues;
```
Jalle19: I commented on this back when this was introduced but @Glenn-1990 said this method was only ever called once or something like that.

ksooo: he was wrong.

Glenn-1990: Damn, somehow I didn't experienced this while testing. 
@ksooo Thanks for the fix! And indeed I was wrong :-)

PR #285, opened 2017-02-09
Revert "[htsp v26] implement server based playcount and playposition"

Glenn-1990:
@ksooo Revert playcount/playposition


ksooo:
Thanks.

PR #287, opened 2017-02-18
Workaround Slow Htps on LAN

jonudewux:


ksooo:
@FernetMenta this will no longer work with Leia, right?

PR #289, opened 2017-02-22
Fix htsp

jonudewux:
Better solution I think

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -411,12 +411,16 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
 {
   if (!m_asyncState.WaitForState(ASYNC_EPG))
     return PVR_ERROR_FAILED;
-  
+
   std::vector<PVR_RECORDING> recs;
   {
     CLockObject lock(m_mutex);
     Channels::const_iterator cit;
     char buf[128];
+
+    bool _http = Settings::GetInstance().GetHttPlayRec();
+    int port = Settings::GetInstance().GetPortHTTP();
+    std::string host = Settings::GetInstance().GetHostname();
 
     for (const auto &entry : m_recordings)
     {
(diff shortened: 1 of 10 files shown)


src/Tvheadend.cpp
```diff
@@ -478,6 +482,11 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
         }
       }
 
+      /* Fill recording url */
+      if (_http) {
+        snprintf(rec.strStreamURL, sizeof(rec.strStreamURL), "http://%s:%d/dvrfile/%s",host.c_str(), port, rec.strRecordingId);
```
ksooo: As I said before, I'm afraid that using PVR_RECORDING.strStreamUrl will not work any longer with Kodi v18 Leia. Thus, this approach could be a dead end. @FernetMenta please comment.

Jalle19:
We removed the circular buffer because it was causing some weird issues and slow seeking on recordings. This issue has been brought up on Kodi's bug tracker and @FernetMenta acknowledged that the way Kodi currently reads data from addons is the root cause here. I don't think re-introducing a custom buffer in the addon is the right way to go since it just adds needless complexity.

The addon doesn't use HTTP for recordings because Kodi doesn't handle seeking past the end of the recording (it will stop playback). I'm not sure if this has changed recently though. Maybe @FernetMenta knows.

@FernetMenta in general, do you think there's a better way to "feed" recordings to Kodi going forward (i.e. v18 and onward)?

jonudewux:
@Jalle19 
Are you serious, simplicity vs functionality ?

src/Tvheadend.cpp
FernetMenta: streamURL is deprecated, see forum

FernetMenta:
I don't know what kind of issues you are facing with tvh. I don't see any issues with vnsi.

Jalle19:
@FernetMenta I'm referring to http://trac.kodi.tv/ticket/16915, which is the issue here.

FernetMenta:
@Jalle19 as long streaming api is not split from pvr, caching has to be handled by the addon. even after the split we would need some means for kodi to tell whether the recording is accessed on a LAN or WAN

Jalle19:
Superseded by https://github.com/kodi-pvr/pvr.hts/pull/366

PR #290, opened 2017-02-24
fix install path

mdeguzis:
This also affects Krypton. After view, please merge into the Krypton branch as well.


Jalle19:
Why do you want to do this? It doesn't look right.

mdeguzis:
Source needs the Kodi piece, but use lib as destination. Sorry, will fix and squash the commits.

Jalle19:
@wsnipex is this correct?

wsnipex:
yes, that is ok

PR #293, opened 2017-04-05
4.0.9: Fix timeshift state and data calculation if demuxer is not active. Fixes #292

### LINKED ISSUE #292: Support audio passthrough in Recordings on Kodi 17
Background: [Audio Passthrough on LiveTV](http://forum.kodi.tv/showthread.php?tid=284580)

On Kodi 17, it looks like the intention is that Live TV will now always decode to PCM, but that Recordings should support audio passthrough. Currently (LibreElec 8.0.1), the tvheadend add-on is decoding to PCM for both Live TV and Recordings. I have no knowledge of the Kodi PVR API, but the above thread indicates an update may be required. Thanks!

ksooo:
This should fix #292 

I runtime tested this on macOS, latest kodi master.

@Jalle19 mind doing the review.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.0.8"
+  version="4.0.9"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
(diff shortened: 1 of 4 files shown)


ksooo:
> Does this fix passthrough for recordings or just the timeshift bug when you switch from a timeshifted channel to a recording?

It should fix both issues, but I currently cannot runtime test the passthrough issue.

ksooo:
> Does this fix passthrough for recordings or just the timeshift bug when you switch from a timeshifted channel to a recording?

 Runtime-tested. Fixes both issues. :-)

PR #297, opened 2017-04-29
libXBMC_codec.h removed and changed to libXBMC_pvr.h

AlwinEsch:
Related to Kodi Pull Request xbmc/xbmc#12028 where libXBMC_codec.h becomes removed.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -6,8 +6,7 @@
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>
     <c-pluff version="0.1"/>
-    <import addon="xbmc.pvr" version="5.2.1"/>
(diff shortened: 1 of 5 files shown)


ksooo:
Please immediately update the add-on version numbers. Otherwise we cannot even tell the users which version will be the one first supporting  the new pvr api version.

ksooo:
Next nightly build will have many no longer working pvr add-ons, because you did not even inform the add-on maintainers that they have to update there add-ons. I hope you will support the users / maintainers in the forum?

ksooo:
Last, many if the add-ons are not pulled any longer from kodi-pvr, so, some if the chars you dud there have no effect. Instead, you need to look up the right upstream repos in repo-binary-adsons and to inform the right maintainers to do the change.

ksooo:
And please also update changelog to document the changes for the new add-on version. :-/

ksooo:
To sum up: this could have been done  much better. I'm really disappointed.

PR #308, opened 2017-06-03
[htsp v27] implement server based playcount and playposition

Glenn-1990:
It's been a while, tvheadend 4.2 is out and it includes improved playstatus support.

Kodi master does have an issue that needs to be fixed as well before this code will work as expected.
https://github.com/xbmc/xbmc/pull/12226

The big advantage: 
Start watching a recording in the living room and resume watching in the bedroom etc.

@ksooo

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -452,6 +452,10 @@ PVR_ERROR CTvheadend::GetRecordings ( ADDON_HANDLE handle )
       /* Lifetime (based on retention or removal) */
       rec.iLifetime = recording.GetLifetime();
 
+      /* Play status */
+      rec.iPlayCount = recording.GetPlayCount();
+      rec.iLastPlayedPosition = recording.GetPlayPosition();
+
       /* Directory */
       // TODO: Move this logic to GetPath(), alternatively GetMangledPath()
       if (recording.GetPath() != "")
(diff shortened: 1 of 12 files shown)


src/Tvheadend.cpp
```diff
@@ -598,6 +602,49 @@ PVR_ERROR CTvheadend::RenameRecording ( const PVR_RECORDING &rec )
...
+  htsmsg_t *m = htsmsg_create_map();
+  htsmsg_add_u32(m, "id",           atoi(rec.strRecordingId));
+  htsmsg_add_u32(m, "playposition", playPosition >= 0 ? playPosition : 0); // Kodi uses -1 when fully watched
+  return SendDvrUpdate(m);
+}
+
+int CTvheadend::GetPlayPosition ( const PVR_RECORDING &rec )
+{
+  if (m_conn.GetProtocol() < 27)
+    return PVR_ERROR_NOT_IMPLEMENTED;
```
ksooo: According to the API documentation, this line should read <code>return -1;</code>.

src/Tvheadend.cpp
```diff
@@ -598,6 +602,49 @@ PVR_ERROR CTvheadend::RenameRecording ( const PVR_RECORDING &rec )
...
+
+  const auto &it = m_recordings.find(atoi(rec.strRecordingId));
+  if (it != m_recordings.end() && it->second.IsRecording())
+  {
+    Logger::Log(LogLevel::LEVEL_DEBUG, "Getting play position %i for recording %s", it->second.GetPlayPosition(), rec.strTitle);
+    return it->second.GetPlayPosition();
+  }
+
+  return PVR_ERROR_INVALID_PARAMETERS;
```
ksooo: According to the API documentation, this line should read <code>return -1;</code>.

ksooo:
@Glenn-1990 do you think it would be a good idea to let the user chooses (addon settings) whether server-side playcount/Position shall be used?

Also, version bump and changelog is missing.

Glenn-1990:
@ksooo Addressed your comments.

I also added an advanced setting in case users don't like the new behaviour.

ksooo:
Nice. Thank you very much for this contribution.

PR #312, opened 2017-06-06
Update HTSPTypes.h

DocMAX:
See

https://github.com/tvheadend/tvheadend/commit/33cc05b37ee982c796936d7d143321f95e365724


ksooo:
Can you please choose a more meaningful title for the commit and PR title? 

Moreover, why is the update needed? Does it fix a certain problem?

ksooo:
Okay, meanwhile I've seen the issue you reported. Linking from the PR to this issue would have helped here to avoid confusion. ;-)

Well, your fix is okay in principle, but there are some parts missing. As it takes more time to explain what else is to be done than to fix it, I suggest that I come up with a complete fix later today.

DocMAX:
thank you

PR #313, opened 2017-06-07
Add support for 'default' recording priority setting value. Fixes #311, closes #312

### LINKED ISSUE #311: Some recordings not visible
Hello,

some recordings aren't visible (59 out of 81).
No clue why. I already checked the "owner" of the recordings. They are all the same.

DocMAX

ksooo:
Fixes #311, superseeds #312 

@DocMAX fyi, maybe you can even runtime test this?

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.0.14"
+  version="4.0.15"
   name="Tvheadend HTSP Client"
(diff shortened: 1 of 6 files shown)


DocMAX:
works fine, but needs to go into all branches

ksooo:
Thanks for testing and no, does not need to go into all branches. 

Using stable Kodi with unstable tvheadend is not supported. Too much effort, no gain. Unstable tvheadend can break at any time and it's not worth the hassle to maintain multiple pvr.hts branches with potentially fragile tvh bleeding edge support. Costs to much time and nerves. For instance, a follow-up task for me now is to fix tvheadend, because I discovered a bug there that was introduced with the change that caused the problem we just fixed here.

PR #316, opened 2017-06-17
Add support for genre to recordings.

ksooo:
Genre is available via HTSP since "ever", but it seems we forgot to implement it.

Strange thing is that for epg events both major catagories and sub categories are delivered by tvh, but for recordings only the major category, although a quick look at the tvh code reveals that it should be very easy to support sub categories also for recordings (@Glenn-1990 maybe something for you to add to tvh? htsp version bump required!)

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.0.15"
+  version="4.0.16"
   name="Tvheadend HTSP Client"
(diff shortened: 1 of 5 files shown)


ksooo:
@Jalle19 good to go?

Jalle19:
Bogged down with work at the moment, haven't looked at Github this week until now. Feel free to merge stuff without comments from me!

Glenn-1990:
@ksooo I'll put it on my list ;-)

ksooo:
@Glenn-1990 cool, thanks. BTW, you got our invitation? Welcome aboard, if you like.

PR #320, opened 2017-07-23
api 5.4.0 - drop SwitchChannel

FernetMenta:
https://github.com/xbmc/xbmc/pull/12552


Jalle19:
So only OpenLiveStream will be used from now on?

FernetMenta:
right, SwitchChannel is not called anymore in Kodi master

Jalle19:
Alright, thanks

ksooo:
As https://github.com/xbmc/xbmc/pull/12552 is now in, this one should go in as well.

PR #323, opened 2017-08-20
3.4.27: Add support for 'default' recording priority setting value.

ksooo:
Fixes tvh 4.3 incompatibility with recordings not appearing in Kodi. Reported 'a hundred times' in different forums.

Krypton backport of #313

@Jalle19 mind taking a look.


ksooo:
https://github.com/xbmc/repo-binary-addons/pull/91

PR #330, opened 2017-09-15
Feature/refactor settings remove singleton

stuartjbrown:
This PR refactors the settings logic to remove the singleton and instead add the settings object as a property of the CTvheadend object.  This allows the use of multiple pvr.hts clients within Kodi (which I know is a hack but is very useful for some people).  If Kodi ever allows multiple instances of the same PVR client this is also a step towards that.  In addition I believe that passing dependencies e.g. settings into the dependant class is better than relying on a global instance for the same.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -50,6 +50,7 @@ CHelper_libXBMC_addon *XBMC      = NULL;
 CHelper_libXBMC_pvr   *PVR       = NULL;
 PVR_MENUHOOK          *menuHook  = NULL;
 CTvheadend            *tvh       = NULL;
+tvheadend::Settings              *settings  = NULL;
 
 /* **************************************************************************
  * ADDON setup
@@ -59,7 +60,8 @@ extern "C" {
 
 void ADDON_ReadSettings(void)
 {
-  Settings::GetInstance().ReadSettings();
+  settings = new tvheadend::Settings();
+  settings->ReadSettings();
 }
 
 ADDON_STATUS ADDON_Create(void* hdl, void* props)
@@ -70,7 +72,7 @@ ADDON_STATUS ADDON_Create(void* hdl, void* props)
   /* Instantiate helpers */
   XBMC  = new CHelper_libXBMC_addon;
   PVR   = new CHelper_libXBMC_pvr;
-  
+
   if (!XBMC->RegisterMe(hdl) ||
       !PVR->RegisterMe(hdl))
   {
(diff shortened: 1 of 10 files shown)


src/client.cpp
```diff
@@ -99,7 +101,7 @@ ADDON_STATUS ADDON_Create(void* hdl, void* props)
     }
 
     /* Don't log trace messages unless told so */
-    if (level == LogLevel::LEVEL_TRACE && !Settings::GetInstance().GetTraceDebug())
```
ksooo: We introduced this setting for a reason. Just deleting the code is for sure not the solution for solving the problem not to have have a `CTvheadend` instance at hand here.

src/client.cpp
```diff
@@ -50,6 +50,7 @@ CHelper_libXBMC_addon *XBMC      = NULL;
 CHelper_libXBMC_pvr   *PVR       = NULL;
 PVR_MENUHOOK          *menuHook  = NULL;
 CTvheadend            *tvh       = NULL;
+tvheadend::Settings              *settings  = NULL;
```
ksooo: I thought you were going to eliminate the settings singleton? Instead you are introducing another one here? This does not make sense at all, sorry.

src/client.cpp
```diff
@@ -59,7 +60,8 @@ extern "C" {
 
 void ADDON_ReadSettings(void)
 {
-  Settings::GetInstance().ReadSettings();
+  settings = new tvheadend::Settings();
```
ksooo: Despite from having another settings singleton now (which makes no sense if your goal is to eleiminate it), this introduces a memory leak in case ADDON_ReadSettings gets called more than once. And even it is called only only once, where gets the global `settings` instance deleted?

src/client.cpp
```diff
@@ -109,9 +111,10 @@ ADDON_STATUS ADDON_Create(void* hdl, void* props)
 
   Logger::Log(LogLevel::LEVEL_INFO, "starting PVR client");
 
-  ADDON_ReadSettings();
+  tvheadend::Settings *set = new tvheadend::Settings();
+  set->ReadSettings();
 
-  tvh = new CTvheadend(reinterpret_cast<PVR_PROPERTIES *>(props));
+  tvh = new CTvheadend(reinterpret_cast<PVR_PROPERTIES *>(props), set);
```
ksooo, line 122: As long as you still have a settings singleton (global variable `settings`) it does make NO sence to hold it additionally as a member of `CTvheadend`.

ksooo (changes requested):
While I like the general idea, the implementation can not get my Go, sorry. What you actually do is to introduce another settings singleton (global variable `settings` in client.cpp). Also you just hack away stuff that does not 'fit', like `Settings::GetTraceDebug()`. Finally, it would be great if the whitespace changes could be put into a separate commit.

Jalle19:
Aren't all addons run completely separately anyway? I had no idea that code style could affect the ability to run two addons side by side.

src/client.cpp
FernetMenta: the addon stays a singleton as long those globals are here. there is work in progress to change this. addon like inputstream are already changed to new design.

ksooo:
Please reopen for Matrix if you feel this is still of interest.

PR #332, opened 2017-09-28
Add build instructions for regular users.

bas-t:


README.md
```diff
@@ -26,6 +26,18 @@ suggested by the README.
 The addon files will be placed in `../../xbmc/kodi-build/addons` so if you build Kodi from source and run it directly 
 the addon will be available as a system addon.
 
+### Linux for regular users
```
ksooo: What is a "regular user", compared to a developer? I would not make assumptions who is using this, but write for what it can be used.

README.md
```diff
@@ -26,6 +26,18 @@ suggested by the README.
...
+
+The following instructions assume you will have installed Kodi already in either /usr or /usr/local
```
ksooo: I doubt it is Kodi that is needed to be installed for this, actually it's one or more of it's dev packages (I don't know the details)? You should write here which packages are actually needed. I think kodi package alone will not be sufficient.

README.md
bas-t: Good point. How would you like me to rephrase this?

README.md
bas-t: The list of build-deps was never part of the README.
I fail to see how adding such a list suddenly has become my task.

ksooo: If we add instructions, they have to be clear, otherwise we will not improve anything, but add another source of confusion.

> I fail to see how adding such a list suddenly has become my task.

 Well, do what you like, but if you want this merged we all need to be friendly and constructive.

README.md
ksooo: Make a suggestion, please.

bas-t: How about 'for system wide installations, also suitable for packaging'?

bas-t: However, I don't have a clue as to what the original instructions should be called to distinguish them in the right way from the new ones...

PR #333, opened 2017-10-02
Misc fixes.

ksooo:
Some trivial, but useful/important stuff.

@Jalle19 mind taking a look?

### DIFF
--- a/src/tvheadend/utilities/LocalizedString.h
+++ b/src/tvheadend/utilities/LocalizedString.h
@@ -1,0 +1,61 @@
+#pragma once
+
+/*
+ *      Copyright (C) 2017 Team Kodi
+ *      http://kodi.tv
+ *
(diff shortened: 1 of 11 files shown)


src/tvheadend/utilities/LocalizedString.h
```diff
@@ -0,0 +1,61 @@
...
+#include "../../client.h"
+
+namespace tvheadend
+{
+  namespace utilities
+  {
+    /**
+     * Encapsulates an localized string.
+     */
+    class CLocalizedString
```
Jalle19: Too late now (didn't see this earlier) but there's no point in prefixes namespaced classes like this. AFAIK the other namespaced classes don't have prefixes either.

ksooo: Okay, will change in a follow-up PR. Thanks for spotting.

PR #334, opened 2017-10-04
Source cleanup

ksooo:
@Jalle19 I was in a code cleanup rage and want to share the results. Tried to keep changes review-able by splitting things up into several commits. Comments welcome. :-)


Jalle19:
Great stuff, thanks for picking up! Had a quick look through and IMO it looks good. There are still things that could be done but this is a great start.

ksooo:
Absolutely, this is only the beginning... ;-)

PR #346, opened 2017-12-21
Implement API function GetStreamTimes & PVR Addon API 5.8.0

ksooo:
Implement API function GetStreamTimes (for tv/radion steam, recordings can be added later - currentyl tvh does not give access to recording times).

Remove implementation of deprecated API functions GetBufferTime(Start|End), GetPlayingTime, POsition(Recorded|Live)Stream, MoveChannel. 

Add support for PVR_CHANNEL_GROUP_MEMBER::iSubChannelNumber"

Details: https://github.com/xbmc/xbmc/pull/13228

### DIFF
--- a/src/tvheadend/HTSPDemuxer.cpp
+++ b/src/tvheadend/HTSPDemuxer.cpp
@@ -41,7 +41,8 @@ HTSPDemuxer::HTSPDemuxer ( HTSPConnection &conn )
   : m_conn(conn), m_pktBuffer((size_t)-1),
     m_seekTime(INVALID_SEEKTIME),
     m_seeking(false), m_speedChange(false),
-    m_subscription(conn), m_lastUse(0)
+    m_subscription(conn), m_lastUse(0),
+    m_startTime(0)
 {
 }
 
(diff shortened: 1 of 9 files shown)


src/tvheadend/HTSPDemuxer.cpp
```diff
@@ -272,41 +273,43 @@ PVR_ERROR HTSPDemuxer::CurrentDescrambleInfo ( PVR_DESCRAMBLE_INFO *info )
...
 
   CLockObject lock(m_mutex);
-  if (m_timeshiftStatus.shift != 0)
-    return true;
 
-  return false;
+  times->startTime = m_startTime;
+  times->ptsStart = 0;
+  times->ptsBegin = static_cast<int64_t>(static_cast<double>(m_timeshiftStatus.start / 1000000)) * DVD_TIME_BASE;
+  times->ptsEnd = static_cast<int64_t>(static_cast<double>(m_timeshiftStatus.end / 1000000)) * DVD_TIME_BASE;
```
Jalle19: Remember the "Note: start/end mismatch is not a bug. tvh uses inversed naming logic here!" comment

ksooo: TBH, I never understood this. As a matter of fact, it's working now - it's carefully runtime tested. Too lazy to think through this, tbh. I take it like I observed it.

Jalle19: If it works then it's all good, it could be that the hack we had in place is some old stuff that is no longer relevant in newer tvheadend versions.

PR #347, opened 2018-01-04
4.2.10: Implement GetStreamTimes for recordings

ksooo:
This PR implements the API function `GetStreamTimes` for recordings. Previous PR only supported TV/radio streams.

Remark: For best results, a small tvheadend modification (https://github.com/tvheadend/tvheadend/pull/1076) is needed, but it works okayish (better as before), also with older tvheadend versions.

Finally, with the above mentioned tvh improvement, even playtimes for in-progress recordings do work (first time ever).

@Jalle19 ping.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -667,6 +667,8 @@ int CTvheadend::GetPlayPosition ( const PVR_RECORDING &rec )
   if (m_conn->GetProtocol() < 27 || !Settings::GetInstance().GetDvrPlayStatus())
     return -1;
 
+  CLockObject lock(m_mutex);
+
   const auto &it = m_recordings.find(atoi(rec.strRecordingId));
   if (it != m_recordings.end() && it->second.IsRecording())
   {
@@ -1226,6 +1228,8 @@ PVR_ERROR CTvheadend::DeleteTimer(const PVR_TIMER &timer, bool)
            (timer.iTimerType == TIMER_ONCE_CREATED_BY_AUTOREC))
   {
     /* Read-only timer created by autorec or timerec */
+    CLockObject lock(m_mutex);
+
     const auto &it = m_recordings.find(timer.iClientIndex);
     if (it != m_recordings.end() && it->second.IsRecording())
     {
@@ -1264,6 +1268,8 @@ PVR_ERROR CTvheadend::UpdateTimer ( const PVR_TIMER &timer )
     }
     else
     {
+      CLockObject lock(m_mutex);
+
       const auto &it = m_recordings.find(timer.iClientIndex);
       if (it == m_recordings.end())
       {
@@ -1322,6 +1328,8 @@ PVR_ERROR CTvheadend::UpdateTimer ( const PVR_TIMER &timer )
     if (m_conn->GetProtocol() >= 23)
     {
       /* Read-only timer created by autorec or timerec */
+      CLockObject lock(m_mutex);
+
       const auto &it = m_recordings.find(timer.iClientIndex);
       if (it != m_recordings.end() &&
           (it->second.IsEnabled() == (timer.state == PVR_TIMER_STATE_DISABLED)))
@@ -1514,10 +1522,15 @@ bool CTvheadend::Connected ( void )
     entry.second.SetDirty(true);
   for (auto &entry : m_tags)
     entry.second.SetDirty(true);
-  for (auto &entry : m_recordings)
-    entry.second.SetDirty(true);
   for (auto &entry : m_schedules)
     entry.second.SetDirty(true);
+
+  {
+    CLockObject lock(m_mutex);
+
+    for (auto &entry : m_recordings)
+      entry.second.SetDirty(true);
+  }
 
   /* Request Async data, first is channels */
   m_asyncState.SetState(ASYNC_CHN);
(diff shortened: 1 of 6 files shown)


src/Tvheadend.cpp
```diff
@@ -1586,12 +1586,27 @@ void CTvheadend::OnWake()
...
-  return m_vfs->Open(rec);
+  bool ret = m_vfs->Open(rec);
+
+  if (ret)
+  {
+    const auto &it = m_recordings.find(atoi(rec.strRecordingId));
+    if (it != m_recordings.end())
+    {
+      CLockObject lock(m_mutex);
+      m_playingRecording = &(it->second);
```
Jalle19: Not sure if it's possible but what if the currently playing recording is deleted? Won't m_playingRecording point to a stale pointer and GetStreamTimes will crash?

ksooo: It would, but imo the recording that is currently played cannot be deleted.

ksooo: In my first version I had a member storing the recording id and every GetStreamTimes included a map lookup to find the recording. I thought this would be to expensive.

Jalle19: I guess this is fine then

src/Tvheadend.cpp
```diff
@@ -1586,12 +1586,27 @@ void CTvheadend::OnWake()
...
 {
-  return m_vfs->Open(rec);
+  bool ret = m_vfs->Open(rec);
+
+  if (ret)
+  {
+    const auto &it = m_recordings.find(atoi(rec.strRecordingId));
+    if (it != m_recordings.end())
+    {
+      CLockObject lock(m_mutex);
```
Jalle19: Should the lock protect access to `m_recordings` too?

src/Tvheadend.cpp
```diff
@@ -2231,17 +2254,28 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
-            if (hasAudio && hasVideo) // No need to parse the rest
-              break;
+              hasVideo = true;
           }
         }
       }
-      rec.SetChannelType(hasVideo ? CHANNEL_TYPE_TV :
-          (hasAudio ? CHANNEL_TYPE_RADIO : CHANNEL_TYPE_OTHER));
+
+      if (!htsmsg_get_s64(&file->hmf_msg, "start", &s64) && (start == 0 || start > s64))
```
Jalle19: Could be that I'm reading this the wrong way, but if a recording happens to be split into two files, shouldn't "start" get set to the value of the first part (i.e. `start < s64`, the opposite of what is done here)? Same with stop.

src/Tvheadend.cpp
```diff
@@ -2209,20 +2226,26 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
-  if (!rec.GetChannelType() && m_conn->GetProtocol() >= 25)
+  htsmsg_t *files, *streams;
+  if ((files = htsmsg_get_list(msg, "files")) != NULL)
   {
-    htsmsg_t *files, *streams;
-    if ((files = htsmsg_get_list(msg, "files")) != NULL)
+    htsmsg_field_t *file, *stream;
+    uint32_t u32;
+    int64_t s64;
+    bool needChannelType = !rec.GetChannelType() && m_conn->GetProtocol() >= 25;
```
Jalle19: Is there any point in doing this optionally? We used to call `SetChannelType` unconditionally.

src/Tvheadend.cpp
ksooo: Anyway, I will check again and maybe try to make it more failsafe, just to be sure.

src/Tvheadend.cpp
ksooo: I store  the last value of start, if I get a new start value, I check whee l whether the previous value is larger than the new value. If so, I store the be value. This, I should end up with start holding the lowest value when I'm done with iterating all files. Did I code this wrong?

src/Tvheadend.cpp
ksooo: > We used to call SetChannelType unconditionally.

 Not correct. We called it only if we had no channeltype.

ksooo: was:

<pre>
if (!rec.GetChannelType() && m_conn->GetProtocol() >= 25)
{
  htsmsg_t *files, *streams;
  if ((files = htsmsg_get_list(msg, "files")) != NULL)
...
</pre>

is:

<pre>
htsmsg_t *files, *streams;
if ((files = htsmsg_get_list(msg, "files")) != NULL)
{
...
  bool needChannelType = !rec.GetChannelType() && m_conn->GetProtocol() >= 25;
</pre>

becasue we now need to always parse the "files" map, but still only want to set the channel type according to "files" information if we do not already have a channel type (exactly like before).

Jalle19: Sorry, the diff was a bit hard to read

src/Tvheadend.cpp
Jalle19: No you're right, I read it as `s64 > start` but obviously it's the other way around.

src/Tvheadend.cpp
ksooo: np

ksooo:
@Jalle19 updated and added a new commit to add missing `m_recordings` locks.

PR #348, opened 2018-01-08
Fixed crashes on addon destruction.

ksooo:
I noticed often crashes when deinstalling pvr.hts.

This PR fixes those crashes by:
- closing demuxers before closing tvh connection and stopping tvh worker thread
- ensuring CTvheadend thread worker function is exited when destructing the CTvheadend instance: `StopThread(0)` == "wait infinite" vs. `StopThread()`== "wait 5 secs" (5 secs are not enough sometimes). 

The rest of the code changes is small cosmetical stuff, like 
- introduce `CTvheadend::Stop` as counterpart to existing `CTvheadend::Start`
- do not check for `tvh`instance presence where not needed
- more intermediate `IsStopped`check points in `CTvheadend::Process`

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -57,12 +57,8 @@ CTvheadend::CTvheadend(PVR_PROPERTIES *pvrProps)
 CTvheadend::~CTvheadend()
 {
   for (auto *dmx : m_dmx)
-  {
     delete dmx;
-  }
 
-  m_conn->Stop();
-  StopThread();
   delete m_conn;
   delete m_vfs;
 }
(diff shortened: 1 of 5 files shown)


ksooo:
@Jalle19 mind taking a look?

src/Tvheadend.cpp
```diff
@@ -73,6 +69,15 @@ void CTvheadend::Start ( void )
   m_conn->Start();
 }
 
+void CTvheadend::Stop ( void )
+{
+  for (auto *dmx : m_dmx)
+    dmx->Close();
```
Jalle19: Shouldn't we also do a `delete`?

ksooo: No, delete's are only done in the dtor. Stop just "resets".

Jalle19: Of course

PR #370, opened 2018-07-27
Add setting to ignore duplicate scheduled recordings (tvh4.3+) 

ksooo:
See https://github.com/tvheadend/tvheadend/commit/30f9f7f2a62cab74f7c508849b8ccc4126791811

If the setting is activated (and tvh version is recent enough) only events actually scheduled for recording will be listed in Kodi. Otherwise, also duplicates appear. Functionality is similar to "show/hide duplicates" button in tvh 4.3 web ui:

screen shot 2018-07-27 at 09 03 09

@Jalle19 mind taking a look.

### DIFF
--- a/pvr.hts/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.hts/resources/language/resource.language.en_gb/strings.po
@@ -100,7 +100,15 @@ msgctxt "#30058"
 msgid "Defaults for new timers"
 msgstr ""
 
-#empty strings from id 30059 to 30099
+msgctxt "#30059"
+msgid "Scheduled recordings"
+msgstr ""
+
+msgctxt "#30060"
+msgid "Ignore duplicates (tvh 4.3+)"
+msgstr ""
+
+#empty strings from id 30061 to 30099
 
 msgctxt "#30100"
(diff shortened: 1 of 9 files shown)


pvr.hts/resources/language/resource.language.en_gb/strings.po
```diff
@@ -100,7 +100,15 @@ msgctxt "#30058"
 msgid "Defaults for new timers"
 msgstr ""
 
-#empty strings from id 30059 to 30099
+msgctxt "#30059"
+msgid "Scheduled recordings"
+msgstr ""
+
+msgctxt "#30060"
+msgid "Ignore duplicates (tvh 4.3+)"
```
Jalle19: I'd use Tvheadend instead of tvh

ksooo: Then, the string gets to long and wont fit the settings dialog anymore - at least for more 'comprehensive' languages like French and German. And, we use 'tvh' already for another version dependent setting string.

Jalle19: Okay, then I guess it's fine

PR #374, opened 2018-09-10
Add support for images for recordings via dvrEntryAdd message.

azlm8t:
This patch makes the recording image available when you browse recordings, similar to the existing EPG images.

Tvheadend supports one optional image URL per recording which is sent in the dvrEntryAdd message. Tvheadend gets this image URL via some xmltv programs.

We copy this in to all of the Kodi image paths (icon, thumbnail, fanart) since Kodi seems to scale/zoom the images. With the artwork I've tried, this seems reasonable and looks better than a blank background. But perhaps some xmltv sources only has very low-res images so we should exclude fanart?

For reference: the existing EPG code only has icon images and previously in recordings just the channel icon would appear next to the recordings with no fanart.

When you hit playback, and then press "info", the default skin displays a video camera icon instead of the recording artwork. I have the same behaviour with pvr.mythtv on Kodi master and Kodi 17.6. So I don't believe this patch is missing an extra line where the image path should be set.

I'm happy for the patch to just be modified to whatever is decided (e.g., not setting fanart), or I can re-submit.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.3.7"
+  version="4.3.8"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+4.3.8
+- Add support for thumbnail image for recordings if supplied by tvheadend (tvh4.3+).
+
 4.3.7
 - Fixed an edge case where wrong times were displayed in Kodi video/audio OSD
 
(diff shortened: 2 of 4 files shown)


ksooo:
Thanks for your contribution.

We should not set icon and fanart, only thumbnail.

The API documentation is not very clear on that, but icon is for that channel logo and fan art is "semantically" different from thumbnail.

Furthermore, please increase add-on micro version number and please add a changelog entry.

I will look at the Kodi core issue you described as soon as I find the time.

azlm8t:
Thanks. I've made those changes and confirmed the recording image still appears in the corner of the recordings view, in the "recent recordings" on the TV tab, and when you press info on a recording. The background fanart is no longer displayed when viewing a recording.

I don't know if I should explicitly mention in the changelog that it only works with dev-branch tvheadend (not the current stable), but then tvh dev will become stable at some time. So I said "if supplied by Tvheadend".

ksooo:
> I don't know if I should explicitly mention in the changelog that it only works with dev-branch tvheadend
 
 For other changes of this kind we added simply a `(tvh4.3+)` in the respective line.

azlm8t:
Aha, I see the tvh4.3+. I've updated the changelog accordingly.

PR #376, opened 2018-09-20
PVR Addon API 5.10.2: Implemented support for PVR_ADDON_CAPABILITIES.…

ksooo:
…bSupportsAsyncEPGTransfer

For details see https://github.com/xbmc/xbmc/pull/14458

(Kodi now calls GetEPGForChannel only if it needs sync epg data. This happens when loading single epg events for timers outside configured epg window or after manual epg refresh for a channel was requested by the user.)

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -1458,55 +1458,47 @@ PVR_ERROR CTvheadend::GetEPGForChannel
   Logger::Log(LogLevel::LEVEL_DEBUG, "get epg channel %d start %ld stop %ld", chn.iUniqueId,
            (long long)start, (long long)end);
 
+  /* Build message */
+  htsmsg_t *msg = htsmsg_create_map();
+  htsmsg_add_u32(msg, "channelId", chn.iUniqueId);
+  htsmsg_add_s64(msg, "maxTime",   end);
 
-  /* Note: Nothing to do if "async epg transfer" is enabled as all changes are pushed live to Kodi, then. */
-  if (!Settings::GetInstance().GetAsyncEpg())
+  /* Send and Wait */
   {
-    /* Build message */
-    htsmsg_t *msg = htsmsg_create_map();
-    htsmsg_add_u32(msg, "channelId", chn.iUniqueId);
-    htsmsg_add_s64(msg, "maxTime",   end);
+    CLockObject lock(m_conn->Mutex());
 
-    /* Send and Wait */
-    {
-      CLockObject lock(m_conn->Mutex());
-      
-      if ((msg = m_conn->SendAndWait0("getEvents", msg)) == NULL)
-        return PVR_ERROR_SERVER_ERROR;
-    }
-
-    /* Process */
-    htsmsg_t *l;
-
-    if (!(l = htsmsg_get_list(msg, "events")))
-    {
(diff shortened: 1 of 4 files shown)


src/Tvheadend.cpp
```diff
@@ -1458,55 +1458,47 @@ PVR_ERROR CTvheadend::GetEPGForChannel
   Logger::Log(LogLevel::LEVEL_DEBUG, "get epg channel %d start %ld stop %ld", chn.iUniqueId,
            (long long)start, (long long)end);
 
+  /* Build message */
+  htsmsg_t *msg = htsmsg_create_map();
+  htsmsg_add_u32(msg, "channelId", chn.iUniqueId);
+  htsmsg_add_s64(msg, "maxTime",   end);
 
-  /* Note: Nothing to do if "async epg transfer" is enabled as all changes are pushed live to Kodi, then. */
-  if (!Settings::GetInstance().GetAsyncEpg())
```
Jalle19: Maybe we should remove the non-asynchronous EPG support while we're at it and simply pass set `bSupportsAsyncEPGTransfer` to true?

ksooo: Maybe later, @Glenn-1990 reported that async epg will not work with older tvh versions. Unfortunately I forgot what version got the fix. That's why the default for async epg is still "off".

ksooo: Found it: https://github.com/tvheadend/tvheadend/commit/0de7b5c4ea478821527e66fed45c6bf059cb4e4b

Fixed for tvh > 4.0.x

Shall we drop support for 4.0.x or is this to early?

Jalle19:
@ksooo dropping < 4.x support wouldn't be such a bad idea. LibreElec and practically all distributions ship 4.0 these days

ksooo:
@Jalle19 we dropped support for tvh < 4.x long time ago. :-)

Jalle19:
Oh you meant dropping support for 4.0. That is a bit too soon I think, I only recently upgraded to 4.2 myself :D

src/Tvheadend.cpp
Glenn-1990: @ksooo, it's been a while, but according to the tvh commit you picked (yes, that's the right one), it is fixed for 4.2 and un, not 4.0?

Glenn-1990:
@ksooo How does this work?
Kodi sets the async timespan at startup end tvh starts pushing epg events?
What happens if epg is requested for timers outside of the initial time window?
Just asking because htsp also has a "getEvent" method.

ksooo:
Hi @glenn-1999, nice to get a sign of live from you.

In case an event is not found in Kodi's EPG container, Kodi will now call GetEPGForChannel at the add-on with a very small timeframe that should match only the one event to look up. pvr.hts sends a "getEvents" to tvh to fetch the missing event. 

Not perfect, but I was looking for a simple solution that does not require to introduce a new API function. The approach I have chosen has the benefit that it should work for every add-on OOTB, excect those implementing async epg which is only hts, maybe zattoo and teleboy. That's why I had to do small adjustments for pvr.hts

Glenn-1990:
@ksooo Yea I've been busy with life lately :-)
Your change looks good enough to me.

PR #377, opened 2018-09-26
Add support for fanart images for recordings via dvrEntryAdd message.

azlm8t:
Tvheadend (4.3+) supports one optional fanart image URL per recording
which is sent in the dvrEntryAdd message. We copy this in to the Kodi fanart path.

Currently setting the fanart in Tvheadend for a recording is a manual process
via a manually run support script, but it will become automatic in the lifetime
of Kodi 18.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.3.9"
+  version="4.3.10"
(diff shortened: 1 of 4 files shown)


azlm8t:
To test, either add a field "fanart_image" with a URL in the dvr/log of Tvheadend and bounce the server.

Alternatively, in Tvheadend, there is support/tvhmeta which is a proof of concept of setting the fanart.
`tvhmeta --uuid 1111111... --fanart-url http://.....`
That script is just to show fanart can be set, and to use as a basis for a proper lookup script.

PR #381, opened 2018-10-23
4.4.0: RDS support and minor changes

ksooo:
@Jalle19 mind doing a quick code review?

@da-anda fyi - adding this was by far not so complicated as I feared before. ;-)


src/tvheadend/HTSPDemuxer.cpp
```diff
@@ -384,6 +386,39 @@ bool HTSPDemuxer::ProcessMessage ( const char *method, htsmsg_t *m )
...
+      if (!pkt)
+        return;
+
+      uint8_t* rdsdata = new uint8_t[rdslen];
+
+      // Reassemble UECP block. mpeg stream contains data in reverse order!
+      for (int i = offset - 2, j = 0; i > 3 && i > offset - 2 - rdslen; i--, j++)
+        rdsdata[j] = data[i];
+
+      memcpy(pkt->pData, rdsdata, rdslen);
```
Jalle19: Is rdslen always equal to the actual length of rdsdata? My brain melts when I look at the loop above.

src/tvheadend/HTSPDemuxer.cpp
```diff
@@ -32,6 +32,8 @@
 #define INVALID_SEEKTIME (-1)
 #define SPEED_NORMAL (1000) // x1 playback speed
 
+#define TVH_STREAM_INDEX_OFFSET (1000)
```
Jalle19: What is this?

ksooo: In difference to all others, rds stream is not delivered by tvh, but created by pvr.hts
We need unique stream indexes among all streams - tvh-supplied and pvr.hts created. Easiest way is to add a fixed offset to all streams delivered by tvh and to use numbers less than TVH_STREAM_INDEX_OFFSET for streams created by tvh.

src/tvheadend/HTSPDemuxer.cpp
ksooo: :-) yes.

src/tvheadend/HTSPDemuxer.cpp
Jalle19: Ah, that explains it!

src/tvheadend/HTSPDemuxer.cpp
Jalle19: Hehe good

PR #392, opened 2019-01-07
remove space so markdown link works in github preview

l3d00m:
This is just a really minor "fix", but I wanted to also use this as way to say thank you to you guys! I am relatively new to Kodi, previously had tvheadend running in VLC, and it was mind-blowing how good this addon is integrated into Kodi. It feels so polished with all the things like recording, dvr guides, timeshifting, etc and it works really well too, which is a rare combination :)

Kudos to you


ksooo:
Thanks.

PR #401, opened 2019-02-04
Fix missing description in the EPG and DVR

Verequies:
Some channels store the description in the channel subtitle in HTSP v32+. 
These commits fix the issue and update pvr.hts to 4.4.13.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -2603,13 +2603,6 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
     evt.SetSummary(str);
   if ((str = htsmsg_get_str(msg, "description")) != NULL)
     evt.SetDesc(str);
-  if (evt.GetDesc().empty() && m_conn->GetProtocol() >= 32)
-  {
-    // If no description is available, use summary as description. This was done by tvh prior to htsp version 32.
-    evt.SetDesc(evt.GetSummary());
-    // No duplicate description/summary
-    evt.SetSummary("");
-  }
   if ((str = htsmsg_get_str(msg, "image")) != NULL)
     evt.SetImage(GetImageURL(str));
   if (!htsmsg_get_u32(msg, "nextEventId", &u32))
@@ -2634,6 +2627,27 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
     evt.SetYear(u32);
   if (!htsmsg_get_u32(msg, "dvrId", &u32))
     evt.SetRecordingId(u32);
+
+  if (m_conn->GetProtocol() >= 32) {
+    if (evt.GetDesc().empty()) {
+      /* 
+        Due to changes in HTSP v32, if the description is empty, try
+        to use the summary as the description. If the summary is empty, 
+        try to use the subtitle as the description. Also clear the
+        respective entries to avoid duplicate information being displayed.
+
+        This was done by TVHeadend prior to HTSP v32.
+      */
+      if (!evt.GetSummary().empty()) {
+        evt.SetDesc(evt.GetSummary());
+        evt.SetSummary("");
+      }
+      else if (!evt.GetSubtitle().empty()) {
+        evt.SetDesc(evt.GetSubtitle());
+        evt.SetSubtitle("");
+      }
+    }
+  }
   
   htsmsg_t *l;
   if ((l = htsmsg_get_map(msg, "credits")) != nullptr)
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -2470,6 +2470,20 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   if ((str = htsmsg_get_str(msg, "fanartImage")) != NULL)
     rec.SetFanartImage(GetImageURL(str));
 
+  if (m_conn->GetProtocol() >= 32) {
+    if (rec.GetDescription().empty() && !rec.GetSubtitle().empty()) {
+      /* 
+        Due to changes in HTSP v32, if the description is empty, try
+        to use the subtitle as the description. Clear the subtitle
+        afterwards to avoid duplicate information being displayed.
+
+        This was done by TVHeadend prior to HTSP v32.
+      */
+      rec.SetDescription(rec.GetSubtitle());
+      rec.SetSubtitle("");
+    }
+  }
+
   /* Error */
   if ((str = htsmsg_get_str(msg, "error")) != NULL)
   {
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.4.12"
+  version="4.4.13"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+4.4.13
+- Fixed missing description in EPG and DVR tags.
+
 4.4.12
 - Fixed missing summary in EPG tags.
 


src/Tvheadend.cpp
```diff
@@ -2459,6 +2459,21 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
     rec.SetDescription(str);
+  /* 
+    Due to changes in HTSP v32, try to use subtitle 
+    as the description if the description and summary
+    failed. Clear the subtitle afterwards  to avoid
+    duplicate information being displayed.
+
+    This was done by TVHeadend prior to HTSP v32.
+  */
+  else if (rec.GetDescription().empty() && m_conn->GetProtocol() >= 32)
```
ksooo: Why `else if`and not just `if`? you might have gotten empty desc or summary above...

src/Tvheadend.cpp
```diff
@@ -2459,6 +2459,21 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
+  {
+    if ((str = htsmsg_get_str(msg, "subtitle")) != NULL) {
```
ksooo: You already obtained subtitle some lines above, so there is no need to get it from the htsp msg a second time. => `if (!rec.GetSubtitle().empty()) `

src/Tvheadend.cpp
```diff
@@ -2603,12 +2618,24 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
-    evt.SetSummary("");
+  /* 
+    Due to changes in HTSP v32, try to use the summary
+    as the description. If the summary is empty, try to use
+    the subtitle as the description. Also clear the respective
+    entries to avoid duplicate information being displayed.
+
+    This was done by TVHeadend prior to HTSP v32.
+  */
+  else if (evt.GetDesc().empty() && m_conn->GetProtocol() >= 32)
```
ksooo: Why else ifand not just if? you might have gotten empty desc above...

src/Tvheadend.cpp
```diff
@@ -2603,12 +2618,24 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
+  {
+    if ((str = htsmsg_get_str(msg, "summary")) != NULL) {
```
ksooo: `if (!evt.GetSummary().empty())`

src/Tvheadend.cpp
```diff
@@ -2603,12 +2618,24 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
+      evt.SetDesc(str);
+      evt.SetSummary("");
+    }
+    else if ((str = htsmsg_get_str(msg, "subtitle")) != NULL) {
```
ksooo: `if (!evt.GetSubtitle().empty())`. You may have gotten an empty summary above.

src/Tvheadend.cpp
```diff
@@ -2603,12 +2618,24 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
+    as the description. If the summary is empty, try to use
+    the subtitle as the description. Also clear the respective
+    entries to avoid duplicate information being displayed.
+
+    This was done by TVHeadend prior to HTSP v32.
+  */
+  else if (evt.GetDesc().empty() && m_conn->GetProtocol() >= 32)
+  {
+    if ((str = htsmsg_get_str(msg, "summary")) != NULL) {
+      evt.SetDesc(str);
```
ksooo: `evt.SetDesc(evt.GetSummary())`

src/Tvheadend.cpp
```diff
@@ -2603,12 +2618,24 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
+      evt.SetSummary("");
+    }
+    else if ((str = htsmsg_get_str(msg, "subtitle")) != NULL) {
+      evt.SetDesc(str);
```
ksooo: `evt.SetDesc(evt.GetSubtitle())`

src/Tvheadend.cpp
```diff
@@ -2459,6 +2459,21 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
+    as the description if the description and summary
+    failed. Clear the subtitle afterwards  to avoid
+    duplicate information being displayed.
+
+    This was done by TVHeadend prior to HTSP v32.
+  */
+  else if (rec.GetDescription().empty() && m_conn->GetProtocol() >= 32)
+  {
+    if ((str = htsmsg_get_str(msg, "subtitle")) != NULL) {
+      rec.SetDescription(str);
```
ksooo: `rec.SetDescription(evt.GetSubttitle())`

Jalle19:
I'm wondering if it would make more sense to use the "default" HTSP fields first, then after all that there would be the protocol version check, which if true fixes up fields that need fixing? This thing is becoming a bit unreadable.

ksooo:
I agree.

src/Tvheadend.cpp
Verequies: I had it as else if since we only really want the code to happen if the description is empty. Its only going to do one or the other and it'd be less efficient in my opinion to have it run two checks.

src/Tvheadend.cpp
Verequies: Keeping the else as we don't want to override the summary as description with the subtitle as description. That would cause issues.

src/Tvheadend.cpp
```diff
@@ -2470,6 +2470,24 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   if ((str = htsmsg_get_str(msg, "fanartImage")) != NULL)
     rec.SetFanartImage(GetImageURL(str));
 
+  /* Extra checks for HTSP v32+ */
```
ksooo: Please remove this comment. It is more then obvious what the following code line does.

src/Tvheadend.cpp
```diff
@@ -2634,6 +2645,28 @@ bool CTvheadend::ParseEvent ( htsmsg_t *msg, bool bAdd, Event &evt )
...
+  if (m_conn->GetProtocol() >= 32) {
+    /* 
+      Due to changes in HTSP v32, try to use the summary
+      as the description. If the summary is empty, try to use
+      the subtitle as the description. Also clear the respective
+      entries to avoid duplicate information being displayed.
+
+      This was done by TVHeadend prior to HTSP v32.
+    */
```
ksooo: This comment block fits best after `if (evt.GetDesc().empty()) {`, not before. Please move the comment.

src/Tvheadend.cpp
```diff
@@ -2470,6 +2470,24 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
   if ((str = htsmsg_get_str(msg, "fanartImage")) != NULL)
     rec.SetFanartImage(GetImageURL(str));
 
+  /* Extra checks for HTSP v32+ */
+  if (m_conn->GetProtocol() >= 32) {
+    /* 
+      Due to changes in HTSP v32, try to use subtitle 
+      as the description if the description and summary
+      failed. Clear the subtitle afterwards  to avoid
```
ksooo: This comment is wrong as it refeneces "summary" which is not avail for recordings.

src/Tvheadend.cpp
```diff
@@ -2470,6 +2470,24 @@ void CTvheadend::ParseRecordingAddOrUpdate ( htsmsg_t *msg, bool bAdd )
...
+      duplicate information being displayed.
+
+      This was done by TVHeadend prior to HTSP v32.
+    */
```
ksooo: This comment block fits best after `if (rec.GetDescription().empty()) {`, not before. Please move the comment.

src/Tvheadend.cpp
Verequies: Should this not be removed as well then? Thats why I referenced it.

https://github.com/kodi-pvr/pvr.hts/blob/dc35ade7b4afd0edd65c4170aab981218bad79d2/src/Tvheadend.cpp#L2460-L2461

Verequies:
Fixed up those issues, should be good now. Unless https://github.com/kodi-pvr/pvr.hts/pull/401#discussion_r253854402 needs to be removed?

Verequies:
Fixed that up and amended a comment.

ksooo:
Thanks for your contribution.

Verequies:
No problem, happy to help :)

dkonerma:
Hmm,

assuming we have e.g. an series episode that is named by subtitle.
Then we clear out the valid and usable subtitle and move that to description, just because description is empty otherwise? Or do I get that wrong? ;-)

Cheers
Dietmar.

Verequies:
It attempts to use the summary as the description first, and if thats empty then it attemps to use the subtitle as the description. In my testing TV episodes such as "The Big Bang Theory" on 9 Go, work fine. They show the episode name as the subtitle, however they have a description. But for the majority of channels, the description is in the subtitle.

ksooo:
> assuming we have e.g. an series episode that is named by subtitle.
Then we clear out the valid and usable subtitle and move that to description, just because description is empty otherwise? Or do I get that wrong? ;-)

 We only restore functionality that was "always" in tvheadend, until it got removed there recently. So, there should be no new issues, because functunality stays the same from Kodi user's perspective.

PR #403, opened 2019-02-17
4.4.14 Fix HTSPDemuxer::IsRealTimeStream to always return true if subscription is active.

ksooo:
Details can be found here: https://github.com/kodi-pvr/pvr.hts/commit/f6d248a766fe1b6af318d7e66448e23f9bdb9565#r32349769

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.4.13"
+  version="4.4.14"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>@ADDON_DEPENDS@</requires>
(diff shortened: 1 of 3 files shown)


FernetMenta:
looks like tvh does support reading faster than 1x by using subscription_set_speed

ksooo:
Which means what for implemrntation of  IsRealTimeStream?

FernetMenta:
Not for IsRealTimeStream but related. There is a whole bunch of things going wrong in the addon regarding streaming. But as long as you block required changes like https://github.com/xbmc/xbmc/pull/14488 those can't be fixed. The linked change and this PR for example prevent tv from going 1.5x of normal speed in the timeshift buffer.
In other words: currently things get worse instead of better

ksooo:
Rainer, you know me and you know that I have nearly zero knowledge when it comes to streaming.

I blocked this PR because it made things worse for pvr.hts and even though the root cause might be in pvr.hts itself, with nobody around with the skills to fix pvr.hts this would have "killed" one of the most used pvr addons - which is imo a nono.

I would be more than happy if somebody with your skills would step up to fix those streaming bugs in pvr.hts. And I do really hope, you will do. ;-)

perexg:
Okay, TVH returns packets based on PTS timing, but it seems that there may be an issue for video streams where PTS/DTS differs a lot (to save the bandwidth) so the player might go to the unsync state, because some frames are missing. We need probably determine those cases. The SPEED command is handled in the code, but I need to know the exact frame timing problems. For TVH, you may dump the frames using '--trace timeshift'.

PR #407, opened 2019-02-27
4.4.15: demuxer timeshift improvements

ksooo:
Requires Kodi fix: https://github.com/xbmc/xbmc/pull/14488

Code taken from https://github.com/FernetMenta/pvr.hts/commit/9999beb87dab709689384f1fe03b65e8f2a01a63 Details discussed there.

### DIFF
--- a/src/tvheadend/HTSPDemuxer.cpp
+++ b/src/tvheadend/HTSPDemuxer.cpp
@@ -46,7 +46,7 @@ using namespace tvheadend::utilities;
 HTSPDemuxer::HTSPDemuxer ( HTSPConnection &conn )
   : m_conn(conn), m_pktBuffer((size_t)-1),
     m_seekTime(INVALID_SEEKTIME),
-    m_seeking(false), m_speedChange(false),
+    m_seeking(false),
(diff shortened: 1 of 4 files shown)


ksooo:
Will runtime-test tonight...

src/tvheadend/HTSPDemuxer.cpp
```diff
@@ -297,7 +297,13 @@ bool HTSPDemuxer::IsTimeShifting() const
 
 bool HTSPDemuxer::IsRealTimeStream() const
 {
-  return m_subscription.IsActive();
+  if (!m_subscription.IsActive())
+    return false;
+
+  if (m_subscription.GetSpeed() != SPEED_NORMAL)
```
FernetMenta: this is wrong. where did you get this from?

ksooo: Ooops, this was there in the code before the  last change of this method and slipped in accidentally. I will remove that.

ksooo:
@FernetMenta I removed the code piece that slipped in accidentally.

PR #408, opened 2019-02-28
4.4.16: Fix auto close of timeshifted subscriptions if predictive tuning is activated.

### LINKED ISSUE #406: Subscription not closed when stopping timeshifted playback with predictive tuning enabled
**Steps to reproduce**

1. Configure Tvheadend to enable time shifting - for example:

image

2. Enable the Predictive Tuning options in the add-on settings.

    On my setup I have:

    Number of subscriptions to use: 2
    Unused subscription close delay (seconds): 20   (AKA pretuner_closedelay)

3. Start playback of a tv channel

4. Wait 1 minute

5. Hit Step Back to shift back 10 seconds

6. Hit Stop

7. Check in the Tvheadend web interface, and you'll see the subscription is never closed. (I even waited several hours overnight - it really is never closed)

image

**Log Files**

Here are some logfiles when re-producing the issue using the steps above:
* kodi debug log: [kodi.no-sub-end.log](https://github.com/kodi-pvr/pvr.hts/files/2910504/kodi.no-sub-end.log)
* kodi debug log with pvr.hts tracing enabled:  [kodi.no-sub-end-trace.log](https://github.com/kodi-pvr/pvr.hts/files/2910502/kodi.no-sub-end-trace.log)

**Description**

My Setup is:

* Kodi 18.1 running on a Vero4K+
* pvr.hts version 4.4.14
* Tvheadend 4.2.8 running on a Debian 9.8 server

I have found that this only occurs when predictive tuning is enabled, and only when hitting stop whilst in time-shifted playback. At this point, the only way I've found to close the subscription is to restart Kodi.

I also noticed that when stopping live tv (non-timeshifted) with predictive tuning enabled, it takes _{pretuner_closedelay}_ seconds before the subscription ends, and during that time you get a large number of debug lines in kodi.log as follows (not sure if this is expected):

`DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - Dropped packet with unknown stream index xxxx`

If you hit stop whilst in time-shifted playback with predictive tuning enabled, then these errors continue indefinitely and the subscription is never closed.

ksooo:
Fixes #406

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.4.15"
+  version="4.4.16"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+4.4.16
+- Fix auto close of timeshifted subscriptions if predictive tuning is activated
+
 4.4.15
 - Fix/improve HTSPDemuxer timeshift support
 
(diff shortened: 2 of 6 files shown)


Jalle19 (approved):

ksooo:
@Jalle19 needed to revise as initial approach was not working for all use cases...

PR #410, opened 2019-03-03
API function FillBuffer.

ksooo:


ksooo:
Ooops, this was not ment to be PRed (yet).

Issue #379, opened 2018-10-17
HTSPVFS and in progress recordings


Glenn-1990:
HTSPVFS sends incomplete packets when reaching the end of the file, also when the file is still growing (in progress recordings). This is the root cause of a longstanding issue when trying to watch recordings at the live point. However, not all hw suffers from this. Typically this leads to stuttering or stopping playback. Another problem here is that kodi will not correctly buffer, actually there is no buffer at all.

Incomplete packets are not supported by Kodi, see https://github.com/xbmc/xbmc/blob/389f30da825bbf4baf82c63d758f66966fa0fa13/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamPVRBase.cpp#L65

However I'm wondering how to fix this issue, a few possibilities:
1) Discharge incomplete packets, this doesn't work well as the filepointer on the tvheadend side moves forward on every read.  
2) Let HTSPVFS::Size() return a smaller value (filesize-kodibuffersize), this works but I consider this as a workaround.
3) return the read only if the read is complete, this will introduce a delay. I'm not sure if this is acceptable from the kodi point of view. I did a quick test over here: https://github.com/Glenn-1990/pvr.hts/commit/88b7c7eb8ebe952e93cbce0e7165efc68a049541
4) Make kodi support incomplete reads.

Any opinions ?

Just opening an issue here before I'm putting effort in a PR :-)

FernetMenta:
> Incomplete packets are not supported by Kodi, see 

you misinterpreted this comment. this function is before demuxer. at this time there is no packets (in terms of frames). the linked function can return any number of bytes. zero means eof.

Glenn-1990:
@FernetMenta What I actually mean are incomplete packets of binary data (from the VFS), so requesting x number of bytes but getting less back. This is not supported by imputstream/demuxer right?

FernetMenta:
wrong. as mentioned above: we speak about packets after demuxer. before demuxer it is just a stream that can deliver any number of bytes. the requested number of bytes varies with the underlying protocol, i.e. nfs, smb, etc

Glenn-1990:
@FernetMenta Ok I get that, I will stop speaking about packets :-)
The requested number of bytes is 64k (default addon setting), when reaching eof, the returned number of bytes by prv.hts can be less. My guess is that Kodi can't cope with this? If not, there is an issue somewhere else.

FernetMenta:
Again, returning less is not a problem

[...]

PR #413, opened 2019-04-12
Fix playback when skipping an in progress recording to real-time

jahutchi:
## Description

This PR resolves issue #379: HTSPVFS and in progress recordings

## Motivation and Context

Currently, if you skip right to the end of an in progress recording, then the playback completely chokes; i.e. both the audio and video break up badly, the video freezes, and the audio makes popping noises.

On debugging, I found that the main issue was that Tvheadend can briefly return 0 bytes when playing an in-progress recording towards EOF; the demuxer expects at least 1 bytes to be returned.

My PR resolves this issue by adjusting HTSPVFS::Read() to retry when 0 bytes are returned for an in-progress recording.

After fixing this problem, the other issue I found was that Kodi would buffer for around 10 seconds when skipping fwd to real-time, before eventually resuming normal playback. After some debugging and research I landed on the IsRealTimeStream API call. This PR returns true for IsRealTimeStream when playback is within the last 10 seconds of an in-progress recording.

## How Has This Been Tested?

My testing was over a Gigabit Ethernet using both HD and SD channels.

Backend: TVheadend (v4.2.8) running on a 10yr old HTPC (Acer Revo 3700) running Debian Stretch
Kodi: Running on a Vero4K+ using OSMC Leia nightlies with pvr.hts v4.4.16 (+ this patch).

In this setup, I have conducted several tests as follows:

* Watched a normal, completed recording from start to finish to ensure no regressions were introduced for normal recordings. Also added extra debug logging to show the value of IsRealTimeStream, and could see it was always set to false as it always has been.
* Played an in-progress recording and skipped way beyond real-time playback. This caused the seek to 'snap back' to the end of file i.e. "Real-time" playback.  Playback started almost immediately (buffering circle appeared for only a split second), and playback was fine thereafter.
* Played an in-progress recording (1hr long) and skipped to real-time playback. Then watched the recording right to the end. No stutters or playback issues were observed, and when the show finished the playback ended immediately.
* With extra debugging in place, I monitored the value of IsRealTimeStream on an in-progress recording. I performed several small skip operations, and IsRealTimeStream was only set to true when nearing the end of the file (within last 10 seconds).
* Played LiveTV and also tried skipping backwards and forwards in the timeshift buffer to ensure no regressions were introduced in this area.
* Repeated the tests of skipping an in-progress recording to real-time playback using different chunk sizes. I have tried with both the minimum (4KB) and maximum (512KB) and playback was fine in all cases.
* Tried starting a recording and playing it almost immediately after. Playback was fine and I skipped forward (+2-3 seconds) to real-time playback without any problems / buffering /stuttering.

New tests added (16/04):
* Skip forward to real-time playback, then hit pause for 10 seconds, hit play and watch the logfile to ensure that IsRealTimeStream is set to false
* Skip forward to real-time playback, hit pause, then hit skip back (10 seconds) followed by play. Watch the logfile to ensure that IsRealTimeStream is set back to false.
* Pause an in progress recording (non real-time); whilst paused, skip forward to real-time then hit play. Ensure that IsRealTimeStream is set to true.
* Pause an in progress recording (non real-time); whilst paused, skip forward to real-time, then wait >10 seconds, then hit play. Ensure that IsRealTimeStream is set to false.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -1651,17 +1651,31 @@ void CTvheadend::VfsClose()
 
 ssize_t CTvheadend::VfsRead(unsigned char *buf, unsigned int len)
 {
-  return m_vfs->Read(buf, len);
+  return m_vfs->Read(buf, len, VfsIsActiveRecording());
 }
 
 long long CTvheadend::VfsSeek(long long position, int whence)
 {
-  return m_vfs->Seek(position, whence);
+  return m_vfs->Seek(position, whence, VfsIsActiveRecording());
 }
 
 long long CTvheadend::VfsSize()
 {
   return m_vfs->Size();
+}
+
+void CTvheadend::VfsPauseStream(bool paused)
+{
+  if (VfsIsActiveRecording())
+    m_vfs->PauseStream(paused);
+}
+
+bool CTvheadend::VfsIsRealTimeStream()
+{
+  if (VfsIsActiveRecording())
+    return m_vfs->IsRealTimeStream();
+  else
+    return false;
 }
 
 /* **************************************************************************
--- a/src/Tvheadend.h
+++ b/src/Tvheadend.h
@@ -185,6 +185,11 @@ private:
   void ParseEventDelete          ( htsmsg_t *m );
   bool ParseEvent                ( htsmsg_t *msg, bool bAdd, tvheadend::entity::Event &evt );
 
+  /*
+   * VFS
+   */
+  bool VfsIsActiveRecording() const { return m_playingRecording && m_playingRecording->GetState() == PVR_TIMER_STATE_RECORDING; }
+
 public:
   /*
    * Connection (pass-thru)
@@ -223,6 +228,8 @@ public:
   ssize_t VfsRead(unsigned char *buf, unsigned int len);
   long long VfsSeek(long long position, int whence);
   long long VfsSize();
+  void VfsPauseStream(bool paused);
+  bool VfsIsRealTimeStream();
 
   /*
    * stream times (live streams and recordings)
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -286,7 +286,10 @@ bool IsTimeshifting(void)
 
 bool IsRealTimeStream()
 {
-  return tvh->DemuxIsRealTimeStream();
+  if (tvh->m_playingRecording)
+    return tvh->VfsIsRealTimeStream();
+  else
+    return tvh->DemuxIsRealTimeStream();
 }
 
 bool OpenLiveStream(const PVR_CHANNEL &channel)
@@ -307,6 +310,12 @@ bool SeekTime(double time,bool backward,double *startpts)
 void SetSpeed(int speed)
 {
   tvh->DemuxSpeed(speed);
+}
+
+void PauseStream(bool paused)
+{
+  if (tvh->m_playingRecording)
+    tvh->VfsPauseStream(paused);
 }
 
 PVR_ERROR GetStreamProperties(PVR_STREAM_PROPERTIES* pProperties)
@@ -541,10 +550,6 @@ PVR_ERROR OpenDialogChannelSettings(const PVR_CHANNEL&)
 PVR_ERROR OpenDialogChannelAdd(const PVR_CHANNEL&)
 {
   return PVR_ERROR_NOT_IMPLEMENTED;
-}
-
-void PauseStream(bool)
-{
 }
 
 int ReadLiveStream(unsigned char *, unsigned int)
--- a/src/tvheadend/HTSPVFS.cpp
+++ b/src/tvheadend/HTSPVFS.cpp
@@ -24,6 +24,9 @@ extern "C"
 {
 #include "libhts/htsmsg_binary.h"
 }
+#include <ctime>
+#include <chrono>
+#include <thread>
 
 #include "p8-platform/threads/mutex.h"
 #include "p8-platform/util/StringUtils.h"
@@ -40,7 +43,8 @@ using namespace tvheadend::utilities;
 * VFS handler
 */
 HTSPVFS::HTSPVFS ( HTSPConnection &conn )
-  : m_conn(conn), m_path(""), m_fileId(0), m_offset(0)
+  : m_conn(conn), m_path(""), m_fileId(0), m_offset(0),
+    m_eofOffsetSecs(-1), m_pauseTime(0), m_paused(false), m_isRealTimeStream(false)
 {
 }
 
@@ -73,6 +77,7 @@ bool HTSPVFS::Open ( const PVR_RECORDING &rec )
 
   /* Cache details */
   m_path = StringUtils::Format("dvr/%s", rec.strRecordingId);
+  m_fileStart = rec.recordingTime;
 
   /* Send open */
   if (!SendFileOpen())
@@ -93,30 +98,70 @@ void HTSPVFS::Close ( void )
   m_offset = 0;
   m_fileId = 0;
   m_path   = "";
+  m_eofOffsetSecs = -1;
+  m_pauseTime = 0;
+  m_paused = false;
+  m_isRealTimeStream = false;
 }
 
-ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len )
+ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len, bool inprogress )
 {
   /* Not opened */
   if (!m_fileId)
     return -1;
 
-  /* Read */
-  ssize_t read = SendFileRead(buf, len);
+  /* Tvheadend may briefly return 0 bytes when playing an in-progress recording at end-of-file
+     we'll retry 50 times with 10ms pauses (~500ms) before giving up */
+  int tries = inprogress ? 50 : 1;
+  ssize_t read = 0;
 
-  /* Update */
-  if (read > 0)
-    m_offset += read;
-
+  for (int i = 1; i <= tries; i++)
+  {
+    read = SendFileRead(buf, len);
+    if (read > 0)
+    {
+      m_offset += read;
+      return read;
+    }
+    else if (i < tries)
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+  Logger::Log(LogLevel::LEVEL_DEBUG, "vfs read failed after %d attempts", tries);
   return read;
 }
 
-long long HTSPVFS::Seek ( long long pos, int whence )
+long long HTSPVFS::Seek ( long long pos, int whence, bool inprogress )
 {
   if (m_fileId == 0)
     return -1;
 
-  return SendFileSeek(pos, whence);
+  long long ret = SendFileSeek(pos, whence);
+
+  /* for inprogress recordings see whether we need to toggle IsRealTimeStream */
+  if (inprogress)
+  {
+    int64_t fileLengthSecs = std::time(nullptr) - m_fileStart;
+    int64_t fileSize = Size();
+    int64_t bitrate = 0;
+    m_eofOffsetSecs = -1;
+
+    if (fileLengthSecs > 0)
+      bitrate = fileSize / fileLengthSecs;
+
+    if (bitrate > 0)
+      m_eofOffsetSecs = (fileSize - m_offset) > 0 ? (fileSize - m_offset) / bitrate : 0;
+
+    /* only set m_isRealTimeStream within last 10 secs of an inprogress recording */
+    m_isRealTimeStream = (m_eofOffsetSecs >=0 && m_eofOffsetSecs < 10);
+    Logger::Log(LogLevel::LEVEL_TRACE, "vfs seek inprogress recording m_eofOffsetSecs=%lld m_isRealTimeStream=%d", static_cast<long long>(m_eofOffsetSecs), m_isRealTimeStream);
+
+    /* if we've seeked whilst paused then update m_pauseTime
+       this is so we correctly recalculate m_eofOffsetSecs when returning to normal playback */
+    if (m_paused)
+      m_pauseTime = std::time(nullptr);
+  }
+
+  return ret;
 }
 
 long long HTSPVFS::Size ( void )
@@ -148,6 +193,32 @@ long long HTSPVFS::Size ( void )
   htsmsg_destroy(m);
 
   return ret;
+}
+
+void HTSPVFS::PauseStream ( bool paused )
+{
+  m_paused = paused;
+
+  if (paused)
+  {
+    m_pauseTime = std::time(nullptr);
+  }
+  else
+  {
+    if (m_eofOffsetSecs >= 0 && m_pauseTime > 0)
+    {
+      /* correct m_eofOffsetSecs based on how long we've been paused */
+      m_eofOffsetSecs += (std::time(nullptr) - m_pauseTime);
+      m_isRealTimeStream = (m_eofOffsetSecs >=0 && m_eofOffsetSecs < 10);
+      Logger::Log(LogLevel::LEVEL_TRACE, "vfs unpause inprogress recording m_eofOffsetSecs=%lld m_isRealTimeStream=%d", static_cast<long long>(m_eofOffsetSecs), m_isRealTimeStream);
+    }
+    m_pauseTime = 0;
+  }
+}
+
+bool HTSPVFS::IsRealTimeStream ( void )
+{
+  return m_isRealTimeStream;
 }
 
 /* **************************************************************************
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="4.4.16"
+  version="4.4.17"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+4.4.17
+- Fix playback when skipping an in progress recording to real-time.
+
 4.4.16
 - Fix auto close of timeshifted subscriptions if predictive tuning is activated
 - Increase packet queuedepth size for subscriptions to workaround timeshift stuttering problems
--- a/src/tvheadend/HTSPVFS.h
+++ b/src/tvheadend/HTSPVFS.h
@@ -45,9 +45,11 @@ public:
 
   bool Open(const PVR_RECORDING &rec);
   void Close();
-  ssize_t Read(unsigned char *buf, unsigned int len);
-  long long Seek(long long pos, int whence);
+  ssize_t Read(unsigned char *buf, unsigned int len, bool inprogress);
+  long long Seek(long long pos, int whence, bool inprogress);
   long long Size();
+  void PauseStream(bool paused);
+  bool IsRealTimeStream();
 
 private:
   bool SendFileOpen(bool force = false);
@@ -59,6 +61,11 @@ private:
   std::string m_path;
   uint32_t m_fileId;
   int64_t m_offset;
+  int64_t m_fileStart;
+  int64_t m_eofOffsetSecs;
+  int64_t m_pauseTime;
+  bool m_paused;
+  bool m_isRealTimeStream;
 };
 
 } // namespace tvheadend


ksooo:
@FernetMenta what is your opinion on this?

src/tvheadend/HTSPVFS.cpp
```diff
@@ -150,6 +167,12 @@ long long HTSPVFS::Size ( void )
   return ret;
 }
 
+bool HTSPVFS::IsRealTimeStream ( void )
+{
+  /* Set IsRealTimeStream when within last 10MB of an in-progress recording */
+  return (m_eofOffset > 0 && m_eofOffset < 10485760);
```
ksooo: Wouldn't it be better to calculate with a percentage here instead of a fixed (magic) byte value?

jahutchi: @ksooo

Sounds good. To be honest I was struggling to decide on what condition to trigger IsRealTimeStream.

So....what %played of an in-progress recording would you recommend for toggling IsRealTimeStream?

michaelarnauts: I wouldn't use a percentage. This will make the real-time window much shorter on a new recording and really long on a large recording.

I think fixed is the best, unless we can check the remaining seconds or something.

ksooo: Right, percentage is not a better approach; best would be 10 seconds till EOF. No idea how to calculate that, though.

jahutchi: I fully agree that time-based would be best; I originally tried to find ways to calculate this...However it seems in the context of pvr plugin there is no concept of time when dealing with recordings; just filesizes and offsets.

I also wondered whether the tipping point should be a factor of readchunksize; but dont know the VideoPlayer side of things well enough.

Guess the question is...when playing in non-real time mode then how far ahead does it need to be able to read, without delay, in order to avoid buffering. A size, number of seconds, number of readchunksizes?

If it is 10 seconds i think best we can do is agreed on a max filesize which can represent 10s.

I experimented with different sizes and the point it seemed to stop buffering after skipping beyond eof was somewhere between 3-4mb on a hd recording, and around 8mb on uhd (astra demo uhd channel). This is where the 10mb came from. In my tests I also found that simply setting IsRealTimeStream permanently for all recordings works just fine too. Not suggesting we do that but just trying to get things in context.

jahutchi: Looking further at this we can find epoch of when recording started using m_playingRecording->GetFilesStart().

Based on this, current epoch, and current file size, we can work out based on avg bitrate, how many bytes represents 10secs.

Will update PR when done.....

ksooo: 👍

FernetMenta: I would use different values for entering realtime and leaving realtime. You don't want the flag to toggle when playing pointer is on the boundary.

jahutchi: I've adjusted the PR to toggle isRealTimeStream based on whether playback is within 10 seconds (rather than 10mb) of EOF.

It only calculates this as part of Seek operations and stores the bool, so if playing pointer happens to be on the boundary then there's no danger of constantly toggling the flag.

FernetMenta: > It only calculates this as part of Seek operations

imo this is incomplete. you can pause a recording that is close to live and after a while it should become non realtime

jahutchi: @FernetMenta Good spot... clearly will need to re-work this.
I had tried to avoid these calculations after every read when playing an in-progress recording, but maybe that's unavoidable.

jahutchi: Just working on an update to the PR to recalculate m_isRealTimeStream:

1) Following a seek
2) Following each read, but only when m_isRealTimeStream is currently TRUE

If you're in non-realtime mode, then I don't see any way to move into realtime (<10s of EOF), other than a seek (forward) operation. But if you're currently in real-time, then there is obviously the possibility of pause/play to go back to non-realtime (>10s of EOF) which I'd overlooked.

Also, this is the only way I can see us avoiding the danger of constantly toggling the flag when on the boundary during playback.

jahutchi: I've adjusted the PR as above

jahutchi: Having tested my previous change I spotted a floor in the logic of triggering the re-calculation on each Read().

During playback kodi (obviously) reads ahead of the play position (updating m_offset as it goes).

Therefore, doing the same calculations (based on m_offset) after each Read() we are actually calculating how far the end of the read buffer (rather than our playback position) is from the end of the file.

The calculation (based on m_offset) at the end of Seek() is valid because we haven't yet done the first Read() following the Seek; thus m_offset correctly reflects the current playback position.

To correctly handle Pause/Play I've therefore made some changes based on the PauseStream() API call; storing the time that the stream was paused (m_timePaused), and later using this when playback resumes to add a number of seconds to the stored offset (m_eofOffsetSecs); which was calculated during previous Seek().

If a Seek() is done whilst paused then the same calculations are performed, and m_timePaused is additionally updated to the current time. Thus when playback resumes then we can update m_eofOffsetSecs to the correct value.

I've added a few extra test cases above to cover "Pause->Play" and "Pause->Seek->Play".

I plan to run through all my tests (above) again later this week, but would value feedback in the meantime.

jahutchi:
I've completed my testing using the tests outlined above, and all tests above passed successfully - now handles pauses correctly WRT setting IsRealTimeStream :-)

src/Tvheadend.cpp
```diff
@@ -1651,19 +1651,36 @@ void CTvheadend::VfsClose()
...
+void CTvheadend::VfsPauseStream(bool paused)
+{
+  /* only process this event for in-progress recordings */
+  if (m_playingRecording->GetState() == PVR_TIMER_STATE_RECORDING)
+    m_vfs->PauseStream(paused);
+}
+
+bool CTvheadend::VfsIsRealTimeStream()
+{
+  if (m_playingRecording->GetState() == PVR_TIMER_STATE_RECORDING)
```
ksooo: As you are using `m_playingRecording->GetState() == PVR_TIMER_STATE_RECORDING` a lot, you probably should introduce a private inline member function`CTvheadend::VfsIsActiveRecording() const`.

src/tvheadend/HTSPVFS.cpp
```diff
@@ -18,6 +18,7 @@
  *  http://www.gnu.org/copyleft/gpl.html
  *
  */
+#include <ctime>
```
ksooo: system headers **after** module header, please.

src/tvheadend/HTSPVFS.cpp
```diff
@@ -40,7 +41,7 @@ using namespace tvheadend::utilities;
 * VFS handler
 */
 HTSPVFS::HTSPVFS ( HTSPConnection &conn )
-  : m_conn(conn), m_path(""), m_fileId(0), m_offset(0)
+  : m_conn(conn), m_path(""), m_fileId(0), m_offset(0), m_eofOffsetSecs(-1), m_pauseTime(0), m_paused(false), m_isRealTimeStream(false)
```
ksooo: line break after m_offset init, please

src/tvheadend/HTSPVFS.cpp
```diff
@@ -108,15 +114,56 @@ ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len )
...
+    {
+      read = SendFileRead(buf, len);
+      if (read > 0)
+      {
+        m_offset += read;
+        break;
+      }
+      usleep(10000);
+    }
+  }
```
ksooo, line 128: Maybe you should log a warning when leaving after 50 unsuccesful retries?

src/tvheadend/HTSPVFS.cpp
```diff
@@ -108,15 +114,56 @@ ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len )
...
+
+  /* for inprogress recordings see whether we need to toggle IsRealTimeStream */
+  if (inprogress)
+  {
+    int64_t fileSize = Size();
+    int64_t bitrate = 0;
+    m_eofOffsetSecs = -1;
+
+    if ((time(NULL) - m_fileStart) > 0)
+      bitrate = fileSize / (time(NULL) - m_fileStart);
```
ksooo: `time` call is rather expensive. So, just call it once and reuse the value. This is even more accurate.

To put `std::time(nullptr) - m_fileStart` in a variable seems best to me.

And pleaes, use `nullptr` for new code (even if we still use `NULL ` at other places).

src/tvheadend/HTSPVFS.cpp
```diff
@@ -108,15 +114,56 @@ ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len )
...
+
+    if (bitrate > 0)
+      m_eofOffsetSecs = (fileSize - m_offset) > 0 ? (fileSize - m_offset) / bitrate : 0;
+
+    /* only set m_isRealTimeStream within last 10 secs of an inprogress recording */
+    m_isRealTimeStream = (m_eofOffsetSecs >=0 && m_eofOffsetSecs < 10);
+    Logger::Log(LogLevel::LEVEL_TRACE, "vfs seek inprogress recording m_eofOffsetSecs=%lld m_isRealTimeStream=%d", (long long)m_eofOffsetSecs, m_isRealTimeStream);
```
ksooo: c++ style cast, please.

src/tvheadend/HTSPVFS.cpp
```diff
@@ -108,15 +114,56 @@ ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len )
...
+
+    /* if we've seeked whilst paused then update m_pauseTime
+       this is so we correctly recalculate m_eofOffsetSecs when returning to normal playback */
+    if (m_paused)
+      m_pauseTime = time(NULL);
```
ksooo: nullptr

src/tvheadend/HTSPVFS.cpp
```diff
@@ -150,6 +197,32 @@ long long HTSPVFS::Size ( void )
   return ret;
 }
 
+void HTSPVFS::PauseStream ( bool paused )
+{
+  m_paused = paused;
+
+  if (paused)
+  {
+    m_pauseTime = time(NULL);
```
ksooo: nullptr

src/tvheadend/HTSPVFS.cpp
```diff
@@ -150,6 +197,32 @@ long long HTSPVFS::Size ( void )
...
+  }
+  else
+  {
+    if (m_eofOffsetSecs >= 0 && m_pauseTime > 0)
+    {
+      /* correct m_eofOffsetSecs based on how long we've been paused */
+      m_eofOffsetSecs += (time(NULL) - m_pauseTime);
```
ksooo: nullptr :-)

src/tvheadend/HTSPVFS.cpp
```diff
@@ -150,6 +197,32 @@ long long HTSPVFS::Size ( void )
...
+      m_isRealTimeStream = (m_eofOffsetSecs >=0 && m_eofOffsetSecs < 10);
+      Logger::Log(LogLevel::LEVEL_TRACE, "vfs unpause inprogress recording m_eofOffsetSecs=%lld m_isRealTimeStream=%d", (long long)m_eofOffsetSecs, m_isRealTimeStream);
```
ksooo: c++ style cast :-)

ksooo (changes requested):
minors only.

src/Tvheadend.cpp
jahutchi: Done - New inline VfsIsActiveRecording() now used in each of the 4 places this check was used.

src/tvheadend/HTSPVFS.cpp
jahutchi: Apologies.... now moved

src/tvheadend/HTSPVFS.cpp
jahutchi: Done

src/tvheadend/HTSPVFS.cpp
jahutchi: Done... adjusted break to return, and added a log message after the loop

src/tvheadend/HTSPVFS.cpp
jahutchi: Thanks for letting me know... I'm new to contributing to this and the kodi project, so picking up these things as I learn the ropes. Had only gone for time(NULL) as I could see that was how it was done elsewhere in this project.

jahutchi: Done

src/tvheadend/HTSPVFS.cpp
jahutchi: Done

src/tvheadend/HTSPVFS.cpp
jahutchi: Done

src/tvheadend/HTSPVFS.cpp
jahutchi: Done

src/tvheadend/HTSPVFS.cpp
jahutchi: Done

jahutchi:
@ksooo care to take another look?

I don't expect these changes to have caused any regressions, but please don't merge until I've retested (tonight) just to be on the safe side.

src/Tvheadend.cpp
ksooo: Nope, cannot see this change in the latest commit.

ksooo:
You have not (force-)pushed the changes so far. Cannot re-review.

jahutchi:
> 
> 
> You have not (force-)pushed the changes so far. Cannot re-review.

My apologies.... trying again

src/Tvheadend.h
```diff
@@ -185,6 +185,11 @@ class CTvheadend
   void ParseEventDelete          ( htsmsg_t *m );
   bool ParseEvent                ( htsmsg_t *msg, bool bAdd, tvheadend::entity::Event &evt );
 
+  /*
+   * VFS
+   */
+  bool VfsIsActiveRecording() const { return m_playingRecording->GetState() == PVR_TIMER_STATE_RECORDING; }
```
ksooo: I would add a null check for `m_playingRecording`before accessing it, like `return m_playingRecording && m_playingRecording->GetState() == PVR_TIMER_STATE_RECORDING;`

I know it is not necessary atm, but just to be on the safe side.

src/Tvheadend.cpp
```diff
@@ -1651,19 +1651,34 @@ void CTvheadend::VfsClose()
...
 }
 
 long long CTvheadend::VfsSize()
 {
   return m_vfs->Size();
 }
 
+void CTvheadend::VfsPauseStream(bool paused)
+{
+  /* only process this event for in-progress recordings */
```
ksooo: I would remove this comment, as the following code line should be pretty self-explaining.

jahutchi:
Now ready for re-review  ;-)

src/tvheadend/HTSPVFS.cpp
```diff
@@ -32,6 +32,8 @@ extern "C"
 #include "Settings.h"
 #include "utilities/Logger.h"
 
+#include <ctime>
```
ksooo: Well, I meant to put it after `#include "libhts/htsmsg_binary.h"` (but outside the extern C block), please.

src/tvheadend/HTSPVFS.cpp
```diff
@@ -108,15 +116,58 @@ ssize_t HTSPVFS::Read ( unsigned char *buf, unsigned int len )
...
+    {
+      read = SendFileRead(buf, len);
+      if (read > 0)
+      {
+        m_offset += read;
+        return read;
+      }
+      usleep(10000);
+    }
+    Logger::Log(LogLevel::LEVEL_ERROR, "vfs inprogress recording read failed after 50 attempts");
```
ksooo: Is this an error or a warning? I think it is more a warning.

ksooo: Could't this method be simplified to:

```c++
{
  /* Not opened */
  if (!m_fileId)
    return -1;

  /* Tvheadend may briefly return 0 bytes when playing an in-progress recording at end-of-file
     we'll retry 50 times with 10ms pauses (~500ms) before giving up */
  int tries = inprogress ? 1 : 50;

  for (int i = 1; i <= tries; i++)
  {
    ssize_t read = SendFileRead(buf, len);
    if (read > 0)
    {
      m_offset += read;
      return read;
    }
    else if (i < tries)
      usleep(10000);
  }
  Logger::Log(LogLevel::LEVEL_WARNING, "vfs inprogress recording read failed after %d attempts", tries);
  return -1;
}
```

(not tested!)

ksooo (changes requested):
almost done, I'd say.

src/tvheadend/HTSPVFS.cpp
jahutchi: That's certainly cleaner code... will adjust the PR

Had thought about doing it that way; think I was trying to make the intention of the loop clear. That said, I later added the comment which explains things.

Will adjust the log message too as this makes it (potentially) apply to normal recordings too.

Not too sure about returning -1 when (final) read could be 0 - this is a change in current behaviour.

Shouldn't we return whatever was returned by the previous SendFileRead like it does currently?

src/tvheadend/HTSPVFS.cpp
jahutchi: Done

src/Tvheadend.h
jahutchi: Done

src/Tvheadend.cpp
jahutchi: Done

src/tvheadend/HTSPVFS.cpp
jahutchi: Hows this look

```cpp
{
  /* Not opened */
  if (!m_fileId)
    return -1;

  /* Tvheadend may briefly return 0 bytes when playing an in-progress recording at end-of-file
     we'll retry 50 times with 10ms pauses (~500ms) before giving up */
  int tries = inprogress ? 1 : 50;
  ssize_t read;

  for (int i = 1; i <= tries; i++)
  {
    read = SendFileRead(buf, len);
    if (read > 0)
    {
      m_offset += read;
      return read;
    }
    else if (i < tries)
      usleep(10000);
  }
  Logger::Log(LogLevel::LEVEL_WARNING, "vfs read failed after %d attempts", tries);
  return read;
}
```

ksooo: Looking good, except the not initialized 'read' variable. If somebody changes the code later so that the loop might be not executed at all we will end with a random return value.

jahutchi: Agreed will start it out at zero.

However, not currently able to compile as LEVEL_WARNING is not declared....

```
pvr.hts/src/tvheadend/HTSPVFS.cpp: In member function 'ssize_t tvheadend::HTSPVFS::Read(unsigned char*, unsigned int, bool)':
pvr.hts/src/tvheadend/HTSPVFS.cpp:127:15: error: 'LEVEL_WARNING' is not a member of 'tvheadend::utilities::LogLevel'
```

Looks like we only have these declared in Logger.h:
      LEVEL_ERROR,
      LEVEL_INFO,
      LEVEL_DEBUG,
      LEVEL_TRACE

Shall I go for LEVEL_DEBUG / INFO or extend "enum LogLevel" in src/tvheadend/utilities/Logger.h to include LEVEL_WARNING?

ksooo: I suggest to go far DEBUG.

jahutchi: Done

jahutchi:
> 
> 
> almost done, I'd say.

Latest changes pushed... still need to test

ksooo (approved):
Nice. Tell me, when you're done with your tests.

src/tvheadend/HTSPVFS.cpp
jahutchi: Just found this obvious mistake so will correct this:
`int tries = inprogress ? 1 : 50;`

Should be:
`int tries = inprogress ? 50 : 1;`

Will update the PR shortly.

jahutchi: Now corrected

src/tvheadend/HTSPVFS.cpp
```diff
@@ -93,30 +96,70 @@ void HTSPVFS::Close ( void )
...
+  for (int i = 1; i <= tries; i++)
+  {
+    read = SendFileRead(buf, len);
+    if (read > 0)
+    {
+      m_offset += read;
+      return read;
+    }
+    else if (i < tries)
+      usleep(10000);
```
ksooo: Is usleep available on Windows? If not, we need a portable solution before we can merge this.

ksooo: We could use https://en.cppreference.com/w/cpp/thread/sleep_for

jahutchi: I was just looking at the same page.

This seems to compile
```cpp
#include <chrono>
#include <thread>
...
else if (i < tries)
 std::this_thread::sleep_for(std::chrono::milliseconds(10));
```

ksooo: yeah, I guess this will do the trick.

jahutchi: Change pushed....still need to run through my tests...

ksooo: sure, ping me if you're done

jahutchi:
@ksooo I'm done with my testing, all working as expected.

ksooo:
Here we go... thanks for your contribution.

src/client.cpp
```diff
@@ -309,6 +312,12 @@ void SetSpeed(int speed)
   tvh->DemuxSpeed(speed);
 }
 
+void PauseStream(bool paused)
```
FernetMenta, line 315: using this function is wrong. you have to recalculate the conditions in read because player can run at lower speed, i.e. 0.8.

PR #418, opened 2019-04-25
[cmake] remove usage of kodi-platform

olafhering:
The pvr.hts binary addon does not use kodi-platform.
Remove existing references from CMakeLists.txt and debian/control.


ksooo:
master is unused

PR #423, opened 2019-08-07
[pvr.hts] bump to 5.1.0

enen92:
Due to https://github.com/xbmc/xbmc/pull/16444 and GUI API breakage, a version bump is required


ksooo:
@enen92 new version should be 5.1.0, not 5.1.1

PR #443, opened 2020-03-10
Avoid using zero when season/episode information is not present

jahutchi:
The pvr.hts addon always sets the season number to 0, even when there is no season information available in Tvheadend. 

In Leia, this causes a display issue when the Tvheadend event or recording only has an Episode number but no Season number. As shown in these screenshots; kodi believes the Season Number is 0 and therefore treats this as a Special episode:

screenshot1

screenshot2

This is only a minor display issue. However, it will be made worse in Matrix owing to the following commit:

https://github.com/xbmc/xbmc/commit/3375a26606b3a6913bea18f4ef7b85a72b724915

This will mean that all events and recordings without any season/episode information in Tvheadend will always been shown as S0.

This can be easily resolved by initializing m_season and m_episode (for completeness) to -1.

### DIFF
--- a/src/tvheadend/entity/Event.h
+++ b/src/tvheadend/entity/Event.h
@@ -52,8 +52,8 @@ public:
       m_stars(0),
       m_age(0),
       m_aired(0),
-      m_season(0),
-      m_episode(0),
+      m_season(-1),
+      m_episode(-1),
       m_part(0),
       m_recordingId(0),
       m_year(0)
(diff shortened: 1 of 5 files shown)


Jalle19 (approved):
@ksooo

src/tvheadend/entity/Event.h
```diff
@@ -52,8 +52,8 @@ class Event : public Entity
       m_stars(0),
       m_age(0),
       m_aired(0),
-      m_season(0),
-      m_episode(0),
+      m_season(-1),
+      m_episode(-1),
```
ksooo, line 56: Those members are unsigned int, thus cannot carry signed values.

src/tvheadend/entity/Recording.h
```diff
@@ -72,8 +72,8 @@ class Recording : public Entity
       m_playCount(0),
       m_playPosition(0),
       m_contentType(0),
-      m_season(0),
-      m_episode(0),
+      m_season(-1),
+      m_episode(-1),
```
ksooo, line 76: Those members are unsigned int, thus cannot carry signed values.

ksooo (changes requested):
Please change season/episode members and setters/getters type to signed int and we should be good to go.

jahutchi:
> Please change season/episode members and setters/getters type to signed int and we should be good to go.

Made a start but realised a few other variables needed adjusting...will take a look tomorrow.

jahutchi:
> 
> 
> > Please change season/episode members and setters/getters type to signed int and we should be good to go.
> 
> Made a start but realised a few other variables needed adjusting...will take a look tomorrow.

I think I've identified and adjusted all the getters and setters... now ready for a further review.

ksooo (approved):
Thank you.

jahutchi:
Are backports for Leia still being accepted? Will be happy to create a pull request if so?

ksooo:
Is Leia affected by this bug? I don't think so.

ksooo:
Ah, it is. So, go ahead and open a PR for Leia.

PR #445, opened 2020-03-14
API Change for iEpisodePartNumber EPG_TAG default value

phunkyfish:
5.4.2
- Update PVR API 6.2.0

Do not merge yet

PR #446, opened 2020-03-17
Fix pointer to freed memory in EPG_TAG.strFirstAired.

ksooo:
Fixes a regression introduced by #445

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="5.4.2"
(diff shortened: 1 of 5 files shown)


phunkyfish (approved):
👍

ksooo:
Sorry, this is nonsense. The member was there before this PR, as the diff clearly shows.

ksooo:
You need to update the Kodi code you are trying to compile the add-on with. There was an incompatible Kodi API change recently.

nkichukov:
I missed your comments. I did not realize this declaration is coming from kodi includes. I will research and if I manage to identify what changed I can cherry pick into the branch I am using for my kodi platform. Thanks @ksooo

PR #449, opened 2020-03-31
[Matrix] API change update / cleanups

AlwinEsch:
Related to xbmc/xbmc#17586 and to bring in after them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,23 +1,50 @@
 # build artifacts
 build/
+pvr.*/addon.xml
+
+# Debian build files
 debian/changelog
 debian/files
-debian/kodi-pvr-hts-dbg.debhelper.log
-debian/kodi-pvr-hts-dbg.substvars
-debian/kodi-pvr-hts-dbg/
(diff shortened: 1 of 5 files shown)


ksooo (approved):

phunkyfish:
I think maybe this one was to wait for the kodi change right?

ksooo:
Oh, but looking at the XBMC PR, I cannot see any PVR add-on API change.

phunkyfish:
Is it not the header file change. Maybe I’m mistaken.

AlwinEsch:
Ups, not seen that this was gone in, was thought to bring in after the Kodi request.
There is a very big API change inside (change of function table from `.` to `->`. Not direct much to change on addons, but need a recompile after them.

See this https://github.com/xbmc/xbmc/pull/17586/commits/9c118a8da45f8b7afe0e147e3576c59c59b6edec, specially on `xbmc_pvr_dll.h`

ksooo:
Oh, sorry. My bad. Then we need another version bump for this add-on once the xmbc pr went in.

PR #453, opened 2020-05-16
[Matrix] API change update

AlwinEsch:
This changes are related to API change with xbmc/xbmc#17813

Need to bring in after them!


ksooo:
@AlwinEsch If I read the kodi change right, this PR makes the addon ABI incompatible, so increase the addon version number to 6.0.0, please.

AlwinEsch:
Will increase all PVR addon in Major version.

PR #455, opened 2020-05-25
[Matrix] change to new C++ PVR interface way

AlwinEsch:
Changes related to xbmc/xbmc#16485.

This make the addon to the new C++ addon interface.
Runtime test was OK.
Only is in Kodi Matrix a bug with is channels icons and not shown, where is in Leia OK (not related to change here). On other addons they has worked, maybe with missing file extensions there?


src/tvheadend/HTSPConnection.cpp
```diff
@@ -238,7 +240,7 @@ void HTSPConnection::SetState(PVR_CONNECTION_STATE state)
 
     /* Notify connection state change (callback!) */
     serverString = GetServerString();
-    PVR->ConnectionStateChange(serverString.c_str(), newState, nullptr);
+    m_client.ConnectionStateChange(serverString, newState, "");
```
ksooo: Please do not keep the client instance at the connection as a member, instead extend the IHTSPConnectionListener interface with a method ConnectionStateChange and call it from here (m_connlistener->ConnectionStateChange(...)). CTvheadend already implements IHTSPConnectionListener, let it also implement the new method. Thus, we can keep the software design clean.

src/tvheadend/HTSPConnection.h
```diff
@@ -115,6 +115,8 @@ class HTSPConnection : public P8PLATFORM::CThread
 
   bool m_suspended;
   PVR_CONNECTION_STATE m_state;
+
+  kodi::addon::CInstancePVRClient& m_client;
```
ksooo: No need for this member; see my other comment.

src/tvheadend/HTSPDemuxer.cpp
```diff
@@ -131,7 +131,7 @@ DemuxPacket* HTSPDemuxer::Read()
   }
   Logger::Log(LogLevel::LEVEL_TRACE, "demux read nothing");
 
-  return PVR->AllocateDemuxPacket(0);
+  return m_client.AllocateDemuxPacket(0);
```
ksooo: I'd prever not to hold the client instance as a member here, but to use an interface for decoupling, like I did with IHTSPConnectionListener. Please create a new interface IDemuxPacketHandler, let CTVHeadend implement it and hold a ref to that listener as a member here.

src/tvheadend/HTSPDemuxer.cpp
```diff
@@ -557,30 +542,28 @@ void HTSPDemuxer::ParseMuxPacket(htsmsg_t* m)
...
-    if (stream.iPID != audioIdx)
+    if (stream.GetPID() != audioIdx)
       continue;
 
     // Found the stream with the embedded RDS data. Create corresponding RDS stream.
-    const CodecDescriptor codecDescriptor = CodecDescriptor::GetCodecByName("rds");
-    const xbmc_codec_t codec = codecDescriptor.Codec();
-
-    if (codec.codec_type == XBMC_CODEC_TYPE_UNKNOWN)
+    kodi::addon::PVRCodec codec = m_client.GetCodecByName("rds");
```
ksooo: Same pattern here. Instaed of using the client instance directly, create an interface ICodecProvider, let CTvheadend implement it, keep a ref to that interface as member of this class.

src/tvheadend/HTSPDemuxer.h
```diff
@@ -39,7 +48,7 @@ class HTSPConnection;
 class HTSPDemuxer
 {
 public:
-  HTSPDemuxer(HTSPConnection& conn);
+  HTSPDemuxer(kodi::addon::CInstancePVRClient& client, HTSPConnection& conn);
```
ksooo: No need for the client, but for refs to the two new interfaces I mentioned earlier.

src/tvheadend/HTSPDemuxer.h
```diff
@@ -120,6 +129,8 @@ class HTSPDemuxer
   uint32_t m_rdsIdx;
   int32_t m_requestedSpeed = 1000;
   int32_t m_actualSpeed = 1000;
+
+  kodi::addon::CInstancePVRClient& m_client;
```
ksooo: Replace this by refs to the two new interfaces I mentioned.

ksooo (changes requested):
Wow, good job. Minors only.

AlwinEsch:
@ksooo have updated the request.
On "IHTSPConnectionListener" is the "ConnectionState" function added where now in CTvheadend
About the demux I created the "IHTSPDemuxPacketHandler" and also now as parent on CTvheadend

Hope it is correct and match your wanted way.

src/tvheadend/IHTSPDemuxPacketHandler.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+
+#include "kodi/DemuxPacket.h"
+#include "kodi/addon-instance/pvr/Stream.h"
+
+namespace tvheadend
+{
+
+/*
+ * HTSP Connection Listener interface
```
ksooo: HTSP Demux Packet Handler interface

src/tvheadend/IHTSPConnectionListener.h
```diff
@@ -27,6 +27,9 @@ class IHTSPConnectionListener
   virtual void Disconnected() = 0;
   virtual bool Connected() = 0;
   virtual bool ProcessMessage(const std::string& method, htsmsg_t* msg) = 0;
+  virtual void ConnectionState(const std::string& connectionString,
```
ksooo: I would name this method `ConnectionStateChange`.

src/tvheadend/IHTSPDemuxPacketHandler.h
AlwinEsch: Ups a c&p

src/tvheadend/IHTSPConnectionListener.h
AlwinEsch: Is renamed, was on begin not sure if same name and function on two parent classes not become conflict, but build and runtime test was OK.

PR #456, opened 2020-05-27
fix disconnect on sleep

FernetMenta:
when kodi goes to sleep all network connections should be closed. just implemented what the comment already has suggested.


ksooo:
Not 100% sure this does not introduce any bad side effects, but at the first glance the fix looks correct to me.

PR #457, opened 2020-06-13
[Matrix] fix problem with pvr instance thread, during destruct

AlwinEsch:
Also was no ATTRIBUTE_HIDDEN on instance class


ksooo (approved):
The missing call to `CTvheadend::Stop()`reintroduced by this PR is fallout from the change to the new C++ API. The call previously was made from the `ADDON_Destruct` function of the addon, which is no longer available with the new API. Thus, that call has to be moved to the dtor of `CTvheadend`.

Issue #458, opened 2020-06-26
DVR lifetime default of 3 months is not synchronized with TVH backend and can unintentionally delete recordings


m4tt075:
Last week, I started realizing, that some recordings were deleted after 3 months, although my TVH backend has always been configured to keep recordings "forever". I then learned that there is a (newish?) timer setting on recording retention in the pvr.hts addon configuration, which defaulted to 3 months, and "overrules" the backend setting. Maybe also the default of 3 months is newish (based on recent Kodi upgrades?). I don't know. I also don't know whether this is a "broader" issue. There was at least one other user on Kodinerds, who struggled with the same problem.

In any case, I don't think that having two (independent) settings on recording retention makes sense. Personally, I'd consider that setting to sit naturally in the backend (as it used to be?), as it is probably mostly about managing disk space on the server. Of course it could also be removed from the backend instead. But if you have both, I'd expect them to be synchronized.

Hope that makes sense. Any thoughts appreciated...

ksooo:
Sorry, works as designed.

ksooo:
Not possible to sync the values as backend does not offer an API for this.

ksooo:
The setting including its default value in the addon is there for years, btw.

https://github.com/kodi-pvr/pvr.hts/pull/138/commits/92b1919ec39f346e9cdb922a72b2bc439492ccb8 <= Dec 2015

ksooo:
So, cant be so much of a problem, otherwise we would have gotten many many complaints earlier.

FernetMenta:
Anyway, I can confirm that there was a change in behaviour recently. A friend of mine complained about disappearing recordings. Maybe the setting was ignored in earlier versions or the default values was "accidentally" set to infinity or disk full. IMO a default of three months is not user friendly.

ksooo:
@FernetMenta  Thanks for the feedback. Setting the default to "infinite" to avoid unexpected data loss is a good idea. However, syncing with the backend setting, which would be the best solution, is currently not possible. As far as I know there is no API for that.

m4tt075:
Wow, I'm honored! If both of you are looking into this, I know it will be alright eventually...   :-)
Thanks, guys!

m4tt075:
[...]

PR #459, opened 2020-06-27
[Settings] Change default DVR lifetime to forever

m4tt075:
The pvr.hts addon timer setting for the dvr lifetime defaults to 3 months. As it overwrites the TVH backend settings and is not synchronized with it, it can lead to unintentional data loss. This PR changes the default setting to forever. It provides a workaround, for issue #458 until addon and backend settings can be synchronized.


ksooo:
Superseded by #460

PR #460, opened 2020-06-28
v7.1.0: Change default value for 'recordings lifetime' setting

ksooo:
- Add new setting value for recordings lifetime of timers created by Kodi: 'Use backend setting"
- Change default value for 'lifetime' setting from '3 months' to 'Use backend setting' to avoid unintended loss of data.
- Enforce reset of default value for lifetime' to 'Use backend setting' to avoid unintended loss of data for existing addon configurations

As discussed here: https://github.com/kodi-pvr/pvr.hts/issues/458

@m4tt075 mind doing a runtime-test?

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="7.0.1"
+  version="7.1.0"
   name="Tvheadend HTSP Client"
   provider-name="Adam Sutton, Sam Stenvall, Lars Op den Kamp, Kai Sommerfeld">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,10 @@
+7.1.0
+- Add new setting value for recordings lifetime of timers created by Kodi: 'Use backend setting"
+- Change default value for 'lifetime' setting from '3 months' to 'Use backend setting' to
+  avoid unintended loss of data
+- Enforce reset of default value for lifetime' to 'Use backend setting' to avoid unintended loss
+  of data for existing addon configurations
+
 7.0.1
 - Fix hang during addon stop
 
--- a/pvr.hts/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.hts/resources/language/resource.language.en_gb/strings.po
@@ -382,7 +382,12 @@ msgctxt "#30389"
 msgid "Forever (tvh 4.1+)"
 msgstr ""
 
-#empty strings from id 30390 to 30399
+#. Recording lifetime representation
+msgctxt "#30390"
+msgid "Use backend setting"
+msgstr ""
+
+#empty strings from id 30391 to 30399
 
 msgctxt "#30400"
 msgid "Predictive tuning"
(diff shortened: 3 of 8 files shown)


m4tt075:
@ksooo Certainly. Could you please create a Windows build for me to test it?

ksooo:
@m4tt075 sure. => https://jenkins.kodi.tv/view/Addons/job/kodi-pvr/job/pvr.hts/view/change-requests/job/PR-460/lastSuccessfulBuild/artifact/cmake/addons/build/zips/pvr.hts+windows-x86_64/pvr.hts-7.1.0.zip

ksooo:
@m4tt075 any feedback?

m4tt075:
@ksooo  Yes, here where I am:

**Testenvironment**
- Kodi Build 19.0-ALPHA1 (18.9.701), Nightly from 2020-06-27 on Windows
- The addon build you referenced
- TVH backend 4.3-1886~g51a4c5bec (dockered) 

**Test results**
| Backend setting | Addon timer setting | Retention value in recording log | Removal value in recording log |
|-------------------|-----------------------|-------------------------------------|-----------------------------------|
| Forever | Use backend setting | 0 | 0 |
| 1 week | Use backend setting | 0 | 0 |
| 1 week | 2 weeks | 0 | 14 |
| 6 months | 2 weeks | 0 | 14 |
| 6 months | Use backend setting | 0 | 0 |
| Forever | Use backend setting | 0 | 0 |

The way I read this, specified timer settings in the addon are interpreted correctly by the backend, but the "Use backend setting" sets retention to forever instead of respecting the actual backend setting. I have to say that I did not restart the backend after changing the backend settings though, as I don't think that that should be required. Anyways, please let me know if I should test differently...

ksooo:
~> but the "Use backend setting" sets retention to forever instead of respecting the actual backend setting~

 ~Then, you have to open an issue for tvheadend. This is clearly a backend bug.~

EDIT: Not sure you are reading the backend log correctly. A "0" there imo means "use value specified in the corresponding dvr backend config". 0 does NOT mean 'forever'. 'Forever' is MAX_INT (thus a very big number). For me, the log reads like everything should work as expectd now.

m4tt075:
Not sure either, but I would have expected different backend retention values to be reflected in the `dvr/log` files. It would also mean that changing the rention time in the backend was applied retrospectively to old recordings, which seems counter-intuitive to me. I have not been able to find other, related parameters in the log files either, though... Hmh.

ksooo:
I'm afraid you have to open an issue for tvheaded if things do not work as expected. Nothing the addon can do about that. I just use an offcial tvheadend API.

ksooo:
How about, instead of trying to "decrypt" the log files, just doing a real runtime test?

ksooo:
Regarding difference between retention and removal: this addon only modifies removal. removal is about when the file (incl. db entry) gets removed, retention is when the db entry is removed without removing the recording file. A mess, if you ask me and very confusing.

So, I'm not astonished that your test reveals that retention value is always zero. The addon does not touch this value, so it always is 0 (which means "current DVR config").

ksooo:
To conclude: I think neither tvheadeend is buggy nor the implementation done with this PR. You simply misread the log files.

m4tt075:
Might well be the case, @ksooo. I'm in no position to judge. Just testing and reporting...  :-)
If you want me to test or report something else, please let me know. If you are reassured, please just move on...
I appreciate you took this on. If it works correctly, this is a clear improvement IMHO. Much appreciated...

m4tt075:
I learned something more from reading old forum posts: The "retention" value refers to the dvr LOG retention period and has nothing to do with the FILE retention. The FILE retention indeed is supposed to be stored in the "removal" variable. And I've found some old, unanswered reports that the backend settings are not always operating correctly. Maybe this IS a backend problem after all. I'll open a ticket on tvheadend.org. Maybe "Flole" knows more...

PR #461, opened 2020-07-08
7.1.1: Fix 'provider name' and 'device status' signal status values after update to PVR API 7.0.0

ksooo:
Fixes fallout from #455 

Before:
screenshot000

After:
screenshot002


Jalle19 (approved):

AlwinEsch (commented):
Thanks a lot for this fixes and sorry that it was gone away before.

PR #466, opened 2020-08-15
increase 'getDiskSpace' timeout

### LINKED ISSUE #465: frequent UI freeze on LibreELEC / OSMC
I get very frequent UI freezes when the Tvheadend plugin loads the available channels and records in the overview on LibreELEC and OSMC on a Raspberry Pi 1. I can still login via ssh etc. but the Kodi UI becomes completely frozen at 100% CPU usage.

Since I can still login via ssh and this happens before playing any live stream, I think that this issue is different from https://github.com/kodi-pvr/pvr.hts/issues/307.

I get a couple of errors:
```
ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - Command getDiskSpace failed: No response received
ERROR: GetDriveSpace: Add-on 'Tvheadend:omv.local:9982' returned an error: server error
ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - failed to read packet (Bad file descriptor)
ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
```
But I am not sure if they actually cause the freeze or if they are just a symptom of the freeze.

I have this issue since years on the Raspberry Pi, but never encountered them on more capable hardware like my Ubuntu laptop. I have the feeling that this is some kind of race condition that is more likely triggered on slower single-core systems. If I restart the Pi a sufficient amount of times, it will successfully read all the metadata and let me access live TV streams.

Any ideas how I can further debug this issue, e.g. which "server error" is produced and which "command failed"?

Full logs:
`kodi.log`: https://pastebin.com/raw/zts4gHtA
`kodi_crash.log` (triggered by `-SIGSEGV`): https://pastebin.com/raw/VHhT3d7v

christianrauch:
I encountered very frequent UI freezes in Kodi with the HTS addon on LibreELEC Raspberry Pi 1 related to timeouts of the `getDiskSpace` method.

Every time a UI freeze occurred, I found:
```
ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - Command getDiskSpace failed: No response received
ERROR: GetDriveSpace: Add-on 'Tvheadend:omv.local:9982' returned an error: server error
ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - failed to read packet (Bad file descriptor)
ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
```
in the `kodi.log`.

Increasing the timeout for `getDiskSpace` prevents these error messages and further the UI freezes. I think this is just related to the less powerful Raspberry Pi 1 hardware and the resulting longer processing times. I still feel that increasing the timeout is just a workaround. Ideally, the plugin should call `getDiskSpace` asynchronously and update the disk space once a response has been received. Locking the UI because no response has been received is not a real option.

Fixes https://github.com/kodi-pvr/pvr.hts/issues/465.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -121,7 +121,7 @@ PVR_ERROR CTvheadend::GetDriveSpace(uint64_t& total, uint64_t& used)
   CLockObject lock(m_conn->Mutex());
 
   htsmsg_t* m = htsmsg_create_map();
-  m = m_conn->SendAndWait("getDiskSpace", m);
+  m = m_conn->SendAndWait("getDiskSpace", m, std::max(30000, Settings::GetInstance().GetResponseTimeout()));
   if (!m)
     return PVR_ERROR_SERVER_ERROR;
 


Jalle19:
Could you not simply increase the request timeout to 30 seconds in the addon settings?

christianrauch:
> Could you not simply increase the request timeout to 30 seconds in the addon settings?

Probably. But why do you want to keep the Kodi UI freeze in the default configuration? This is not a reasonable choice. Every use of Kodi on a Raspberry Pi 1 has then to search the internet for the solution of their strange "UI freeze" issue.

`CTvheadend::SendDvrDelete` is using the [same minimum timeout for the `updateDvrEntry` method](https://github.com/kodi-pvr/pvr.hts/blob/b3b9d122bf6fef8f83d1edea91d03988d95db7de/src/Tvheadend.cpp#L383). So why would we increase the timeout for `updateDvrEntry` in the code, but not for `getDiskSpace`?

christianrauch:
@Jalle19 @ksooo Can this go ahead? Is there something blocking this?

ksooo:
> Is there something blocking this?
 
As you said. This does not fix the root cause. So, I guess we will not merge this.

christianrauch:
> > Is there something blocking this?
> 
> As you said. This does not fix the root cause. So, I guess we will not merge this.

It does not fix the UI freeze directly, but it does fix the `getDiskSpace` timeout.

You are free to open a PR for the UI freeze fix, and I am happy to try it out, but even then will the `getDiskSpace` timeout.

ksooo:
To quote you: " I still feel that increasing the timeout is just a workaround." I do agree to that. If we have a real fix, we will merge that.

christianrauch:
I don't see a problem/contradiction here.

Yes, this PR will not definitely solve the UI freeze issue. But I don't claim it will do. **This PR solves the timeout issue** that you will have independent from the UI freeze issue.

Since I am facing this timeout issue, and other's with less power full hardware will do too, I am proposing this PR to **fix the timeout issue**. The UI issue has to be fixed with another PR.

ksooo:
Which exact LibreELEC version are you using? debug.log files. please.

christianrauch:
> Which exact LibreELEC version are you using? debug.log files. please.

It is the same logfiles that I linked in the issue:
`kodi.log`: https://pastebin.com/raw/zts4gHtA
`kodi_crash.log` (triggered by `-SIGSEGV`): https://pastebin.com/raw/VHhT3d7v

ksooo:
> Fixes #465.
>> But I don't claim it will do. 

 You do and maybe that's confusing me.

christianrauch:
> > Fixes #465.
> > > But I don't claim it will do.
> 
> You do and maybe that's confusing me.

No, I stated that fixing the UI freeze is just a by-product of the timeout fix ("is just a workaround").

It should be clear from the diff and the title that this PR increases the `getDiskSpace` timeout. If there is still an issue with the purpose of this PR I can change the description.

ksooo:
Looking at your log:

```log
2020-08-15 15:26:51.487 T:1639969664   ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - Command getDiskSpace failed: No response received
2020-08-15 15:26:51.530 T:1639969664   ERROR: GetDriveSpace: Add-on 'Tvheadend:omv.local:9982' returned an error: server error
2020-08-15 15:26:51.533 T:1669657472   ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - failed to read packet (Bad file descriptor)
2020-08-15 15:26:51.534 T:1669657472   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - new connection requested
2020-08-15 15:26:51.535 T:1669657472   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - connecting to omv.local:9982
2020-08-15 15:26:51.681 T:1669657472   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - connected
2020-08-15 15:26:51.693 T:1606398848   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - sending hello
2020-08-15 15:26:51.706 T:1606398848   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - connected to Tvheadend / 4.3-1896~gce0907705 (HTSPv35)
2020-08-15 15:26:51.707 T:1606398848   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - sending auth
2020-08-15 15:26:51.729 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   Received permissions:
2020-08-15 15:26:51.730 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   administrator              : 0
2020-08-15 15:26:51.730 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   HTSP streaming             : 1
2020-08-15 15:26:51.731 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   HTSP DVR                   : 1
2020-08-15 15:26:51.731 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   Failed/aborted DVR         : 1
2020-08-15 15:26:51.732 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   anonymous HTSP only        : 0
2020-08-15 15:26:51.732 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   global connection limit    : 0
2020-08-15 15:26:51.744 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   DVR connection limit       : 0
2020-08-15 15:26:51.745 T:1606398848    INFO: AddOnLog: Tvheadend HTSP Client: pvr.hts -   streaming connection limit : 0
2020-08-15 15:26:51.745 T:1606398848   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - rebuilding state
2020-08-15 15:26:53.254 T:1661264768   DEBUG: Persist: Persisting channel group 'Alle Kanäle' with 26 channels
2020-08-15 15:26:54.189 T:1661264768   DEBUG: Update: Updating timers
...
2020-08-15 15:26:59.204 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:26:59.204 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:26:59.205 T:1661264768   DEBUG: Update: Updating timers
...
2020-08-15 15:27:04.206 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:27:04.207 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
...
2020-08-15 15:31:54.897 T:1661264768   DEBUG: Update: Updating timers
2020-08-15 15:31:59.898 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:31:59.898 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:36:54.028 T:1661264768   DEBUG: Update: Updating timers
2020-08-15 15:36:59.029 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:36:59.030 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:41:54.160 T:1661264768   DEBUG: Update: Updating timers
2020-08-15 15:41:59.161 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:41:59.162 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:46:54.291 T:1661264768   DEBUG: Update: Updating timers
2020-08-15 15:46:59.292 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:46:59.293 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:51:54.422 T:1661264768   DEBUG: Update: Updating timers
2020-08-15 15:51:59.423 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:51:59.424 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:56:54.554 T:1661264768   DEBUG: Update: Updating timers
2020-08-15 15:56:59.556 T:1661264768   ERROR: GetTimers: Add-on 'Tvheadend:omv.local:9982' returned an error: the command failed
2020-08-15 15:56:59.556 T:1661264768   ERROR: GetTimers: PVR client 'Tvheadend:omv.local:9982' returned an error: the command failed
```

After getDiskSpace failed, the addon reconnects to the backend, but something goes wrong, as "rebuilding state" never seems to finish and from then every call from kodi to the addon is ending with an error. I think this eventually causes the UI freeze.

ksooo:
```c++
    /* Rebuild state */
    Logger::Log(LogLevel::LEVEL_DEBUG, "rebuilding state");
    if (!m_connListener.Connected())
      goto fail;

    Logger::Log(LogLevel::LEVEL_DEBUG, "registered");
    SetState(PVR_CONNECTION_STATE_CONNECTED);
    m_ready = true;
    m_regCond.Broadcast();
    return;
  }

fail:
  if (!m_suspended)
  {
    /* Don't immediately reconnect (spare server CPU cycles)*/
    Sleep(SLOW_RECONNECT_INTERVAL);
    Disconnect();
  }
```

This is the interesting code sequence for the root cause. We do not reach "Logger::Log(LogLevel::LEVEL_DEBUG, "registered");" but jump to "fail" which triggers a reconnect.

As it seems you are a developer - so this would be the point to start debugging...

ksooo:
So, the call to  "m_connListener.Connected()" must return false, which it does only in one case:

```c++
  msg = m_conn->SendAndWait0("enableAsyncMetadata", msg);
  if (!msg)
  {
    m_asyncState.SetState(ASYNC_NONE);
    return false;
  }
```

... which could be caused by a timeout of "enableAsyncMetadata" backend call? Another one you want to increase the timeout value for, if this is the case? ;-)

ksooo:
Could you please check the value of "m_suspended" when the "fail:" block in "HTSPConnection::Register" gets called? If it for some reason is true (and which would be a bug), we would not trigger a reconnect, which would lead to the bad behavior we're observing.

What also would be helpful is a new log with "trace" enabled in pvr.hts addon settings.

ksooo:
An interesting test case would be to keep the short timeout value for "getDiskSpace" to run into the reconnect case, but to increase the timeout value for the "enableAsyncMetadata" backend call to check whether this fixes the UI freeze (which I guess it will do.) IMO, the UI freeze is caused by pvr.hts unable to reconnect to the backend after failed "getDiskSpace" call and we need to find out why the reconnect fails.

christianrauch:
Increasing the timeout for `enableAsyncMetadata` does not prevent the freeze and in fact I see the `async updates requested` in the log which means that it does not `return false;` before.

christianrauch:
> This is the interesting code sequence for the root cause. We do not reach "Logger::Log(LogLevel::LEVEL_DEBUG, "registered");" but jump to "fail" which triggers a reconnect.

How do you get to this conclusion?

I see ``2020-08-15 15:26:42.443 T:1661264768   DEBUG: AddOnLog: Tvheadend HTSP Client: pvr.hts - registered``
in the log. Isn't this the debug output you were looking for?

ksooo:
There are ofc two backend connects in your log. e.g. two times "rebuilding state", but only one "registered". Everything before 

2020-08-15 15:26:51.487 T:1639969664   ERROR: AddOnLog: Tvheadend HTSP Client: pvr.hts - Command getDiskSpace failed: No response received

is not interesting for our case.

However, meanwhile I found out that the real problem most probably is that the call to "m_connectionListener.Connected()" will never return.

BTW: I do not really like your tone. I'm investing hours of my live to solve your (!) problem. Maybe it's just the language barrier, if not, I would like you to act a bit less "aggressive".

ksooo:
This:

```log
Thread 35 (Thread 0x5fbfb380 (LWP 331)):
#0  __lll_lock_wait (futex=0x72e693c0, private=0) at lowlevellock.c:46
#1  0x76eaadcc in __GI___pthread_mutex_lock (mutex=0x72e693c0) at pthread_mutex_lock.c:115
#2  0x76d8c534 in P8PLATFORM::CMutex::Lock() () from /usr/lib/libcec.so.4
#3  0x683353a4 in CTvheadend::Connected() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#4  0x68343f48 in tvheadend::HTSPConnection::Register() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#5  0x68344bd0 in tvheadend::HTSPConnection::HTSPRegister::Process() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#6  0x76d8c6e8 in P8PLATFORM::CThread::ThreadHandler(void*) () from /usr/lib/libcec.so.4
#7  0x76ea83cc in start_thread (arg=0x5fbfb380) at pthread_create.c:486
#8  0x74da48d8 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /usr/lib/libc.so.6
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

-> this call does not return. Seems to be a deadlock. THIS is the root cause.

ksooo:
**We have a classical deadlock between two threads and need to fix this:**

```log
Thread 27 (Thread 0x64050380 (LWP 320)):

has tvh mtx -> CTvheadend::Process()
wants conn mtx

#0  __lll_lock_wait (futex=0x72e66390, private=0) at lowlevellock.c:43
#1  0x76eaadcc in __GI___pthread_mutex_lock (mutex=0x72e66390) at pthread_mutex_lock.c:115
#2  0x76d8c534 in P8PLATFORM::CMutex::Lock() () from /usr/lib/libcec.so.4
#3  0x683426dc in tvheadend::HTSPConnection::GetProtocol() const () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#4  0x68337508 in CTvheadend::ParseEvent(htsmsg*, bool, tvheadend::entity::Event&) () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#5  0x683382f0 in CTvheadend::ParseEventAddOrUpdate(htsmsg*, bool) () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#6  0x6833c028 in CTvheadend::Process() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#7  0x76d8c6e8 in P8PLATFORM::CThread::ThreadHandler(void*) () from /usr/lib/libcec.so.4
#8  0x76ea83cc in start_thread (arg=0x64050380) at pthread_create.c:486
#9  0x74da48d8 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /usr/lib/libc.so.6
Backtrace stopped: previous frame identical to this frame (corrupt stack?)

===================================================

Thread 35 (Thread 0x5fbfb380 (LWP 331)):

has conn mtx -> HTSPConnection::Register()
wants tvh mtx

#0  __lll_lock_wait (futex=0x72e693c0, private=0) at lowlevellock.c:46
#1  0x76eaadcc in __GI___pthread_mutex_lock (mutex=0x72e693c0) at pthread_mutex_lock.c:115
#2  0x76d8c534 in P8PLATFORM::CMutex::Lock() () from /usr/lib/libcec.so.4
#3  0x683353a4 in CTvheadend::Connected() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#4  0x68343f48 in tvheadend::HTSPConnection::Register() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#5  0x68344bd0 in tvheadend::HTSPConnection::HTSPRegister::Process() () from /storage/.kodi/addons/pvr.hts/pvr.hts.so.4.4.21
#6  0x76d8c6e8 in P8PLATFORM::CThread::ThreadHandler(void*) () from /usr/lib/libcec.so.4
#7  0x76ea83cc in start_thread (arg=0x5fbfb380) at pthread_create.c:486
#8  0x74da48d8 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /usr/lib/libc.so.6
Backtrace stopped: previous frame identical to this frame (corrupt stack?)```

christianrauch:
> BTW: I do not really like your tone. I'm investing hours of my live to solve your (!) problem. Maybe it's just the language barrier, if not, I would like you to act a bit less "aggressive".

Sorry if I sounded aggressive. I just tried to be neutral here about what this PR does (fixing the timeout) and the orthogonal issue of the race conditions that leads to UI freezes.

I am in no way a Kodi developer and I would also argue that is not "my" issue, but an issue in the PVR plugin. I hoped that reporting this issue and giving a hint where this might originate from might help this project and make the code "better" (in "more robust to random freezes").
You can imagine that I also spend my personal time here to fix the timeout issue for everybody that uses the PVR addon on a Raspberry Pi 1. This is an open-source project after all, so we are all here in our free time to contribute to a public good. The timeout issue and the UI freeze have nagged me for a long time and I finally took the time to look into the code and submit a PR. I am running some public repos too and I am generally grateful about every issue report and PR that helps me to identify underlying issues.

Still, I appreciate your efforts in fixing the race condition and I very much prefer to have this race condition fixed in addition to the timeout.

ksooo:
@christianrauch regarding this PR: I'm a bit hesitating with your fix approach, because I think it is not correct to pollute pvr.hts with assumptions regarding how long several tvh backend calls may last. Wouldn't it be better to fix tvheadend to ensure getDiskSpace response within a reasonable time?

christianrauch:
> I think it is not correct to pollute pvr.hts with assumptions regarding how long several tvh backend calls may last.

I see your point that is not a nice solution since it is hard-coded. But this way of ensuring a minimum timeout is used in other parts of the code too and I don't see a downside, other than having to wait for 30s in the worst case before the function proceeds.

> Wouldn't it be better to fix tvheadend to ensure getDiskSpace response within a reasonable time?

I doubt that this can be fixed on the server-side. My laptop, which is also connected via WiFi, does not have troubles with the timeout. I might get less timeouts with a wired Ethernet connection, but generally I assume that this timeout is related to the computational power that is available on the client-side (e.g. a thread will just block for too long because the CPU is saturated).

I think the proper solution would be to wait indefinitely for a response. This could be done synchronously or asynchronously, e.g. the plugin will have to wait until a method (e.g. `getDiskSpace`) returns, or it registers callbacks and updates information (e.g. the disk space) when it becomes available.

ksooo:
tvheadend could setup a thread which periodically calculates free space. getDiskSpace implementation simply would return the last value calculated by that thread.

ksooo:
Hmm, just had a quick look at tvheadend sources. It seems that tvh is aleady doing what I suggested: https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr_vfsmgr.c#L525

No idea how the long response time can be explained, then.

christianrauch:
> tvheadend could setup a thread which periodically calculates free space. getDiskSpace implementation simply would return the last value calculated by that thread.

I am not sure how this is implemented on the server-side (i.e. the server might already cache this value) but even having this value available immediately on the server can still result in timeouts on the client. Since the timeout is based on `CLOCK_MONOTONIC` it can easily happen that a thread blocks for more than the timeout duration because other threads take the CPU for this time.

Given that my laptop does not report this timeout issue supports the theory that this is not a server-side issue. Given more cores and processing power just makes it less likely that the thread passively misses its deadline.

ksooo:
The longer I think about this and the more we know about the problem, I guess the response timeout addon setting is exactly there for what you encounter on your low computing power device.

ksooo:
I guess we can close this one.

Issue #386, opened 2018-11-30
Add support for the HTTP digest authentication (for image / icon requests)


perexg:
https://tvheadend.org/issues/5370

PR #471, opened 2020-09-21
add http streaming capability

FernetMenta:


src/Tvheadend.cpp
```diff
@@ -367,6 +367,34 @@ PVR_ERROR CTvheadend::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet&
...
+  std::string url = "http://";
+  url += Settings::GetInstance().GetUsername() + ":" + Settings::GetInstance().GetPassword() + "@";
+  url += Settings::GetInstance().GetHostname() + ":";
+  url += std::to_string(Settings::GetInstance().GetPortHTTP());
+  url += "/stream/channelid/";
+  url += std::to_string(it->first);
+
+  std::string streamingProfile = Settings::GetInstance().GetStreamingProfile();
+  if (!streamingProfile.empty())
+    url += "?profile=streamingProfile";
```
ksooo: this looks wrong.

FernetMenta: why

FernetMenta: right 😁

FernetMenta: fixed

src/Tvheadend.cpp
```diff
@@ -367,6 +367,34 @@ PVR_ERROR CTvheadend::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet&
...
+  CLockObject lock(m_mutex);
+
+  if (!Settings::GetInstance().GetStreamingHTTP())
+    return PVR_ERROR_NO_ERROR;
+
+  auto it = m_channels.find(channel.GetUniqueId());
+  if (it == m_channels.end())
+    return PVR_ERROR_FAILED;
+
+  std::string url = "http://";
```
ksooo, line 381: std::string url = Settings::GetInstance().GetUseHTTPS() ? "https://" : "http://";

src/Tvheadend.cpp
```diff
@@ -367,6 +367,34 @@ PVR_ERROR CTvheadend::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet&
...
+  url += Settings::GetInstance().GetUsername() + ":" + Settings::GetInstance().GetPassword() + "@";
```
ksooo, line 382: please handle empty username, password correctly.

src/Tvheadend.cpp
FernetMenta: video streaming via https is nonsense. in particular in a home network

src/Tvheadend.cpp
FernetMenta: if username is empty, we won't get to this stage

src/Tvheadend.cpp
ksooo: Why is https nonsense? Why is this limited to home network?

src/Tvheadend.cpp
ksooo: why?

FernetMenta: how would the addon connect to the backend without valid username / pw?

ksooo: Currently it might be the case that kodi calls 'GetChannelStreamProperties' only after successful backend connection was established by the addon (although I'm not sure about this). Think about async connect. May I suggest to make the code a bit more robust?

src/Tvheadend.cpp
FernetMenta: > Why is this limited to home network?

do you remember what the 'p' stands for in pvr?

> Why is https nonsense?

it introduces significant overhead

ksooo: > do you remember what the 'p' stands for in pvr?

 Sure, but why does "personal" imply something technical related to network? Personally I sometimes use a HTSP server located at the other end of europe and it works very well.

src/Tvheadend.cpp
FernetMenta: if username or password are empty or wrong, the channel switch will fail later when opening the url.  this is the desired behavior.

src/Tvheadend.cpp
FernetMenta: if you think https is anything of value, feel free to implement later. I won't do this nonsense

ksooo: I do not need http streaming.

Jalle19: @FernetMenta nonsense or not, there's an option to use HTTPS and breaking that contract with the user would be pretty bad.

src/Tvheadend.cpp
Jalle19: @FernetMenta it's possible to configure tvheadend to accept no or any username/password, so you can't AFAIK rely on execution not reaching this place if either username or password is empty.

Not sure what the most robust solution would be here. Maybe omit the credentials part completely if either username or password is empty.

src/Tvheadend.cpp
```diff
@@ -367,6 +367,34 @@ PVR_ERROR CTvheadend::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet&
...
+  url += Settings::GetInstance().GetHostname() + ":";
+  url += std::to_string(Settings::GetInstance().GetPortHTTP());
+  url += "/stream/channelid/";
+  url += std::to_string(it->first);
```
Jalle19, line 386: What does `td::to_string(it->first)` evaluate to? I'm guessing the channel ID? Maybe make it more explicit?

src/Tvheadend.cpp
Jalle19: When I think about this some more, I think the reason for the HTTPS option is so people don't have to send their username/password in plain text back and forth. Even if we'd skip HTTPS for streaming because it's heavy, we'd still leak the username and password.

The sanest solution would be to implement Digest authentication (https://github.com/kodi-pvr/pvr.hts/issues/386).

Jalle19 (changes requested):

src/Tvheadend.cpp
FernetMenta: https isn't even implemented properly here because there is no method to import the server's certificate. makes https pretty useless here. better not having it at all than having it implemented the way it is.

> The sanest solution would be to implement Digest authentication

feel free to do so. I don't see any need for this because I consider it nonsense to expose a tvh server to the public internet. if I want to connect to my tvh backend from remote, I use vpn.

src/Tvheadend.cpp
FernetMenta: imo this is pretty obvious

src/Tvheadend.cpp
Jalle19: There's no need to import any certificate, most people nowadays (I assume) use legitimate certificates and not self-signed.

Anyway, copying the snippet @ksooo started this thread with is enough, we don't need to get philosophical over this feature right now.

src/Tvheadend.cpp
Jalle19: Yeah you're right, I'm rusty.

src/Tvheadend.cpp
FernetMenta: Who would pay 80€ per year for a trusted certificate? I bet you don't know a single person doing this for their tvh server. I won't add this incomplete https nonsens. If this stops you from merging this, let me know.

FernetMenta: Who would pay 80€ per year for a trusted certificate? I bet you don't know a single person doing this for their tvh server. I won't add this incomplete https nonsens. If this stops you from merging this, let me know.

Jalle19: Let's Encrypt gives anyone free certificates nowadays. I don't personally know anyone doing this, but obviously some do since the HTTPS option was added in the first place.

I'm not sure why you can't just change it to one-liner @ksooo gave you?

a1rwulf: https://letsencrypt.org
No need to pay for trusted certs these days.

FernetMenta: > I'm not sure why you can't just change it to one-liner

because I don't do nonsense. this https code in this repo is crap and I won't add to this. why don't you add this line instead of forcing others to do this.

EDIT: apart from this I consider it bad style to submit code that is untested. I don't use https and can't test it.

ksooo: If you only could stay polite - "crap", "nonsense", ... this all does not help a bit, but upsets people, at least those who have spent their time implementing this (not me, btw.).

hagaygo:
While there is no "perfect" software or "perfect" code but in my experience its better to have 80% of a feature than none.

As far i can see from the code , HTSP streaming is still the default option and only a user that enables the HTTP streaming will be affected from the new code.

Maybe all that is missing is warning message when enabling the HTTP streaming. (experimental feature warning).

Let me know if i can do anything to help make this happen (also hopefully it will be backported/supported on kodi 18.8)

hagaygo:
@FernetMenta 

Hi, Tried "blindly retro fitting" your pull request code to the Leia branch so i can compile it and try using it on my libreelec rpi/windows installs to see if it works and solves my problems.

I manged with a few changes with the xml stuff but the code it self wont compile for start because of 

```
Error (active)	E0135	namespace "kodi::addon" has no member "PVRChannel"	pvr.hts	C:\Temp\kodi\xbmc\cmake\addons\build\pvr.hts\src\Tvheadend.cpp	337	
Error (active)	E0135	namespace "kodi::addon" has no member "PVRStreamProperty"	pvr.hts	C:\Temp\kodi\xbmc\cmake\addons\build\pvr.hts\src\Tvheadend.cpp	337	
Error (active)	E0135	namespace "kodi::addon" has no member "PVRChannel"	pvr.hts	C:\Temp\kodi\xbmc\cmake\addons\build\pvr.hts\src\Tvheadend.h	95	
Error (active)	E0135	namespace "kodi::addon" has no member "PVRStreamProperty"	pvr.hts	C:\Temp\kodi\xbmc\cmake\addons\build\pvr.hts\src\Tvheadend.h	96	
```

Any chance you can make the changes on a fork for 4.4.21 (Leia) versions ?

Thanks for your work so far.

PR #475, opened 2020-10-06
7.2.1: Fixes and p8-platform removal

ksooo:
- Fix 'pvr.hts' appearing twice in every addon log message
- Fix calculation of start time, stop time and file size for multi-file recordings
- Remove p8-platform build dependency

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ project(pvr.hts)
 
 find_package(Kodi REQUIRED)
 
+set(CMAKE_CXX_STANDARD 17)
+
 include_directories(${KODI_INCLUDE_DIR}/.. # Hack way with "/..", need bigger Kodi cmake rework to match right include ways
                     ${PROJECT_SOURCE_DIR}/lib)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,9 @@ set(HTS_SOURCES_TVHEADEND_UTILITIES
                 src/tvheadend/utilities/LifetimeMapper.h
                 src/tvheadend/utilities/AsyncState.cpp
                 src/tvheadend/utilities/AsyncState.h
-                src/tvheadend/utilities/SyncedBuffer.h)
+                src/tvheadend/utilities/SyncedBuffer.h
+                src/tvheadend/utilities/TCPSocket.h
+                src/tvheadend/utilities/TCPSocket.cpp)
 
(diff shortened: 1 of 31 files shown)


src/tvheadend/utilities/TCPSocket.h
```diff
@@ -0,0 +1,120 @@
...
+    auto endpoints = asio::ip::tcp::resolver(m_context).resolve(m_host, std::to_string(m_port));
+
+    std::error_code error;
+    asio::async_connect(
+        m_socket, endpoints,
+        [&](const std::error_code& result_error, const asio::ip::tcp::endpoint&) {
+          error = result_error;
+        });
+
+    if (error)
```
AlwinEsch: About here, as I understand his code of ASIO correct is the error not set until his work was done where is in independent thread, the `RunFor()` then only blocks until he is finished or timeout.

AlwinEsch (commented):
Except the error check it looks good to me :+1:

src/tvheadend/utilities/TCPSocket.h
ksooo: Sorry, I do not understand what you mean.

ksooo: Fixed.

CMakeLists.txt
```diff
@@ -98,7 +96,7 @@ list(APPEND HTS_SOURCES
 
 add_subdirectory(lib/libhts)
 
-set(DEPLIBS ${p8-platform_LIBRARIES} hts)
+set(DEPLIBS hts)
 if(WIN32)
   list(APPEND DEPLIBS ws2_32)
```
AlwinEsch: This also no more needed as the `ASIO` makes `#   pragma comment(lib, "ws2_32.lib")`

ksooo: Okay, will remove, then.

ksooo:
Please note that pvr.hts now uses kissnet socket implementation (https://github.com/Ybalrid/kissnet) , which requires c++ 17.

phunkyfish:
Nice!

ksooo:
@AlwinEsch any idea why Travis build still fails? Errors indicate that the compiler does not understand C++17...

lrusak:
@ksooo the ubuntu version is too old, gcc 5.x doesn't support c++17. You need at least gcc 7 (ubuntu 18.04 bionic beaver)

PR #479, opened 2020-10-22
7.2.2: Fix toggling of HTTP streaming setting

jahutchi:
The HTTP Streaming toggle button in the addon settings does not take effect immediately after toggling the setting.

A change to this setting only takes effect after you either (for example) disable/re-enable the addon, or restart kodi.

You can confirm this by checking in the Status area of the Tvheadend admin interface after toggling this setting and playing a channel:
image

In this case I had ticked the box to use HTTP Streaming, but the htsp protocol was still being used instead.

Looking in the code it looks like #476 missed adding a line to Settings::SetSetting in src/tvheadend/Settings.cpp

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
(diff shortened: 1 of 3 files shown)


ksooo (approved):
Thanks.

PR #485, opened 2020-11-29
8.1.2: More IPv6 fixes

### LINKED ISSUE #483: IPv6 connection to TVHeadend doesn't seem to work with Matrix/pvr.hts 8.1
I just upgraded my Kodi installation to the newly-released Matrix beta (and pvr.hts 8.1 from the add-on repository).  Now that I have done this, pvr.hts will only connect to TVHeadend using IPv4.  Previously (with Kodi 18.x) it would use IPv6 if it was available.  Now with a domain name that has both A and AAAA it will use IPv4 and with a domain name that has only AAAA it refuses to connect entirely.  I don't have an advancedsettings.xml.  I am running Windows 10 19042.630 on the client.

ksooo:
Fixes #483 (needs kissnet fixes).

### DIFF
--- a/lib/kissnet/kissnet.hpp
+++ b/lib/kissnet/kissnet.hpp
@@ -711,7 +711,6 @@ namespace kissnet
 		}
 
 		///sockaddr struct
-		sockaddr_storage socket_output = {};
 		sockaddr_storage socket_input = {};
 		socklen_t socket_input_socklen{};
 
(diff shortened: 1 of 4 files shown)


ksooo:
@Jalle19 how to proceed with this? @Ybalrid stopped responding, so the required kissnet changes are not yet merged upstream. Do we want to merge this PR anyway or wait for kissnet?

Jalle19:
I'd wait until someone with IPv6 is able to test

mamarley:
We already did quite a bit of testing with IPv6.  @ksooo tested it with pvr.hts and Kodi on a Mac and I tested it with pvr.hts and Kodi on Linux and alone on Linux and Windows.  I also added a unit test and expanded several more to ensure that IPv6 works on both the listening and connecting side.

ksooo:
I also tested on Android.

mamarley:
It looks like all of this stuff did get accepted upstream, so we should be good to go either way!

ksooo:
> It looks like all of this stuff did get accepted upstream, so we should be good to go either way!

 I want to have https://github.com/Ybalrid/kissnet/pull/35 merged first. This PR contains an important bug fix.

PR #486, opened 2020-12-27
8.1.3: Small fixes

### LINKED ISSUE #464: Streaming profile not set when startup is set to "Play TV"
When Kodi is set to automatically start Live TV (Settings->Interface->Startup->Perform on startup=Play TV) the profile used in the streaming is the "default" profile and not the one set in the Addon config.

I tried to check the logs but there's no reference to the profile in use.

ksooo:
Fixes #464


ksooo:
Jenkins build error is unrelated.

PR #487, opened 2020-12-29
8.1.3: Fix: Error handling for socket read/write operations.

ksooo:
Forgot to add this commit to #486. Sorry. This is needed for proper reconnect after network outage.


ksooo:
Jenkins error is unrelated

PR #501, opened 2021-03-22
8.3.0: New setting: Regex vs. 'contains' search for Autorecs

ksooo:
Feature suggested in the Forum: https://forum.kodi.tv/showthread.php?tid=361542

screenshot00002

### DIFF
--- a/src/tvheadend/AutoRecordings.cpp
+++ b/src/tvheadend/AutoRecordings.cpp
@@ -17,6 +17,7 @@
 
 #include <cstring>
 #include <ctime>
+#include <regex>
 
 using namespace tvheadend;
 using namespace tvheadend::entity;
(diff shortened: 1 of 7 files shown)


src/tvheadend/AutoRecordings.cpp
```diff
@@ -174,8 +175,15 @@ PVR_ERROR AutoRecordings::SendAutorecAddOrUpdate(const kodi::addon::PVRTimer& ti
...
   htsmsg_add_str(m, "name", timer.GetTitle().c_str());
 
-  /* epg search data match string (regexp) */
-  htsmsg_add_str(m, "title", timer.GetEPGSearchString().c_str());
+  /* epg search data match string */
+  std::string searchString = timer.GetEPGSearchString();
+  if (!Settings::GetInstance().GetAutorecUseRegEx())
+  {
+    // escape regex special chars
+    static const std::regex specialChars(R"([-[\]{}()*+?.,\^$|#])");
```
Jalle19, line 183: Have you validated that this regular expression does what you want it to do?

ksooo: Yes. Do you have concerns?

ksooo: I found it on stackoverflow.

Jalle19: I had some, but when I looked closer I think they're irrelevant. Mainly about escaping `-`, but I think it's okay in its current form.

Jalle19 (approved):

src/tvheadend/AutoRecordings.cpp
ksooo: You may be right. Shall I remove `-` from the list?

ksooo: And what about the `,`?

ksooo: Reading a bit further, I think I will let it like it is now.

PR #506, opened 2021-05-11
8.3.1: Fix: Demuxer must reset start time when reconnecting

ksooo:
Fixes broken timeshift visualization and control after a reconnect.


ksooo:
wip, do not merge

ksooo:
does not work

PR #520, opened 2021-07-18
8.3.4: Fixed 'Use HTTPS' setting init/write.

### LINKED ISSUE #519: Not loading icons when using https reverse proxy
Hello,
i have set up my TVH to be accessed via internet and reverse proxy.
It works fine, but for some reason i dont get any icons / picons when i use it that way.
The channels are all loading and playing, but just the icons are missing.
When i check the logs, it says: "CCurlFile::Open failed with code 400 for http://tvh.domain.de:443/tvheadend/imagecache/411"

Of course i have checked the "use https" inside the addon. But it seems to find the icon with http.
When i simply add the "s" manual, it works in my browser. Just the http produces Error 400.

Is it a bug?

ksooo:
Should fix #519.

Interestingly, this feature never could have worked. The related settings value never was read / set on add-on init, so HTTPS actually never was used.

### DIFF
--- a/pvr.hts/addon.xml.in
+++ b/pvr.hts/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.hts"
-  version="8.3.3"
+  version="8.3.4"
   name="Tvheadend HTSP Client"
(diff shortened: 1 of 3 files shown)


ksooo:
Fix confirmed working by the issue submitter.

PR #556, opened 2021-11-20
fixed: Unable to change existing timer rule to use any time when rela…

### LINKED ISSUE #548: Editing existing timer rule and enabling "Start any time" doesn't work
When you edit an already existing timer rule with "Start any time" disabled and enabling "Start any time" for it, it doesn't work. It will keep "Start any time" disabled after finishing editing the timer rule. When directly changing it in the tvheadend webinterface it does work. I've enabled debug tracing but the log didn't give any clue why this doesn't work. Please let me know if you need any additional info.

arnova:
…xed timer setting is enabled (fixes #548)

### DIFF
--- a/src/tvheadend/AutoRecordings.cpp
+++ b/src/tvheadend/AutoRecordings.cpp
@@ -254,6 +254,11 @@ PVR_ERROR AutoRecordings::SendAutorecAddOrUpdate(const kodi::addon::PVRTimer& ti
       htsmsg_add_s32(m, "start", startWindowBegin);
       htsmsg_add_s32(m, "startWindow", startWindowEnd);
     }
+    else
+    {
+      htsmsg_add_s32(m, "start", -1);
+      htsmsg_add_s32(m, "startWindow", -1);
+    }
   }
   else
   {


ksooo:
Please explain *why* this fixes the issue and how you have tested, especially for unwanted side effects.

src/tvheadend/AutoRecordings.cpp
```diff
@@ -254,6 +254,11 @@ PVR_ERROR AutoRecordings::SendAutorecAddOrUpdate(const kodi::addon::PVRTimer& ti
       htsmsg_add_s32(m, "start", startWindowBegin);
       htsmsg_add_s32(m, "startWindow", startWindowEnd);
     }
+    else
+    {
+      htsmsg_add_s32(m, "start", -1);
+      htsmsg_add_s32(m, "startWindow", -1);
+    }
```
ksooo, line 261: As the note in line 238 says, **not sending** these values (like we did before), will set start and startwindow to "any" time.

Respective tvh code:

```c
      if(htsmsg_get_s32(in, "start", &start))
        start = -1;
      if(htsmsg_get_s32(in, "startWindow", &start_window))
        start_window = -1;
```

So, from my point of view, your code change changes actually nothing on server side. Not sure how this could fix something, sorry.

ksooo: And related code and logic for both the add-on and server is identical for Matrix/Nexus and tvh 4.2/4.3, I've just checked that.

arnova:
> Please explain _why_ this fixes the issue and how you have tested, especially for unwanted side effects.

Well the problem is that with relaxed "start time window calculation" enabled and when we're updating(!) an existing timer rule, the old htsmsg_t-structure is reused/updated. So if it previously didn't have "any start time" enabled, and you change the timer rule to "any start time", the start/startwindow fields will never be updated and still use the old start/startwindow. You understand? It also took me a while to figure out the logic why this was happening. But please be assured: this really fixes the issue. I've extensively tested this with adding/updating different timer rules.

src/tvheadend/AutoRecordings.cpp
arnova: What you're stating is true for new(!) timers, not when updating existing timers (see my comment below) which don't have "any start time"-enabled.

ksooo:
You may be right. I will take a closer look and do a runtime-test.

src/tvheadend/AutoRecordings.cpp
ksooo: correct, just spotted this as well

ksooo (approved):

PR #557, opened 2021-11-20
fixed: Unable to change existing timer rule to use any time when rela…

### LINKED ISSUE #548: Editing existing timer rule and enabling "Start any time" doesn't work

arnova:
…xed timer setting is enabled (fixes #548) (Backport)


ksooo (approved):
Thanks.

ksooo:
@AlwinEsch what does the failed check mean and how to fix? => https://jenkins.kodi.tv/job/kodi-pvr/job/pvr.hts/job/PR-557/1/display/redirect

PR #585, opened 2022-07-18
[Nexus] Increase version to 20.4.0

CastagnaIT:
Bump version for inpustream API change
due to:
https://github.com/xbmc/xbmc/pull/21390
https://github.com/xbmc/xbmc/pull/21319

### DIFF
--- a/pvr.hts/changelog.txt
+++ b/pvr.hts/changelog.txt
@@ -1,3 +1,6 @@
+v20.4.0
+- Kodi inputstream API update to version 3.2.0
+
 v20.3.0
 - Inputstream API bump.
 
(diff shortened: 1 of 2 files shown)


pvr.hts/changelog.txt
```diff
@@ -1,3 +1,6 @@
+v20.3.1
```
ksooo: Sorry, for incompatible API bumps, add-on's minor version number needs to be increased, not micro. Micro is for bug fixes only, which keep API compatibility and add no new functionality.

This holds not only for pvr.hts, but for any binary add-on. So please also adjust the other PRs.

ksooo (changes requested):

ksooo (approved):

ksooo:
Superseded by #592

Issue #361, opened 2018-04-03
Backup server (as second option where first one is unavailable)


nickiman:
Hi,
Recently I had to disable my tvheadend setup (it was running on a QNAP NAS) to configure it in a Raspberry Pi where my DVB-T device was recognized correctly. So, at the end, I have two tvheadend servers running fully operational with
a) SAT>IP device
b) DVB-T device connected right now in the raspberry pi due to the problem before mentioned
c) and several streams that I get from Internet processed with ffmpeg.

I was wondering if could be a good idea to implement in the tvheadend addon the possibility to configure a second IP address where it could be called if the first one server it´s unavailable.

This could be useful in order to have the oportunity to restart the device where tvheadends runs in the first instance and meanwhile the clients can swith to the second tvheadend server and can keep watching the TV.

What´s your opinion?
Keep waiting your feedback...
Regards.

ksooo:
Will not happen any time soon. Too much effort, sorry.

ksooo:
Will be available in Kodi Nexus. See #591

Jalle19:
Closing this

PR #591, opened 2022-09-03
Add support for multiple backends

ksooo:
See https://github.com/xbmc/xbmc/pull/20860


ksooo:
BTW, this fixes #361

src/tvheadend/utilities/SettingsMigration.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  if (mig.Changed())
+  {
+    // Set a title for the new instance settings
+    std::string title;
+    target.CheckInstanceSettingString("host", title);
+    if (title.empty())
+      title = "Migrated Add-on Config";
+
+    target.SetInstanceSettingString("kodi_addon_instance_name", title);
```
Jalle19, line 83: Would perhaps be better to use a dedicated setting for this, e.g. `setting_migration_done`, that way we don't run into unexpected issues if for whatever reason the `kodi_addon_instance_name` setting is later renamed or removed.

ksooo: Yes, definitely better. I will change that.

ksooo: On second thought, I think we actually don't need another setting. The only thing that might go wrong is that if somebody removes the instance title by hand from intsance-settings-1.xml (Kodi GUI prevents this), we migrate the settings for this instance a second time, overwriting the previously user-set values with the old values from settings.xml. Not nice, but if users manipulate files in a way they are not supposed to to, well ...

Jalle19: Alright, sure!

PR #627, opened 2023-11-04
[Omega] Fix TV channels stuttering after wake from suspend

ksooo:
Fixes a regression introduced with #615. 

We need to rebuild demuxer and vfs state early. Doing this when async update is done is to late, because async update can take several seconds and Kodi will try to resume playback early after wakeup from suspend.


Jalle19:
Thank you @ksooo for continuing to put in the effort year after year 🙏

ksooo:
@Jalle19 good to see that you are also still around.

PR #629, opened 2023-11-05
[Nexus] GitHub Actions/Azure build pipeline fixes/updates

ksooo:
* GitHub Actions: ubuntu-18.04 runner is eol. https://github.com/actions/runner-images/issues/6002
* GitHub Actions: Debian build job: Switch xbmc repo to Nexus branch 
* Azure pipelines: Switch xbmc repo to Nexus branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,17 +11,10 @@ jobs:
       matrix:
         include:
         - name: "Debian package test"
-          os: ubuntu-18.04
+          os: ubuntu-latest
           CC: gcc
           CXX: g++
           DEBIAN_BUILD: true
-        #- os: ubuntu-18.04
-          #CC: gcc
-          #CXX: g++
-        #- os: ubuntu-18.04
-          #CC: clang
-          #CXX: clang++
-        #- os: macos-10.15
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -34,9 +27,9 @@ jobs:
       uses: actions/checkout@v2
       with:
         repository: xbmc/xbmc
-        ref: master
+        ref: Nexus
         path: xbmc
-    - name: Checkout pvr.hts repo
+    - name: Checkout add-on repo
```
phunkyfish, line 32: Curious, why is this now add-on instead of pvr.hts?

ksooo: Because the script can be reused for the other addons and then there is one place less to adapt.

ksooo:
I will look at the Jenkins errors later. Nothing that can be fixed here and not introduced by this PR.

.github/workflows/build.yml
phunkyfish: 👍

ksooo:
> I will look at the Jenkins errors later. Nothing that can be fixed here and not introduced by this PR.

Seems nothing that happens always. Run after merge went without errors. Hmm.

PR #636, opened 2023-11-14
[Omega] Fix TV channels stuttering after wake from suspend - take 2

ksooo:
See #634


smp79:
With this fix It reconnects fine now 👍

ksooo:
> With this fix It reconnects fine now

Thanks for reporting back.

PR #663, opened 2024-04-22
Predictive tuning: Fix crash because of an invalid comparator

ace20022:
The compare function violated the antisymmetry. (Counter)example:
grafik


ksooo:
Please provide instructions how to reproduce the crash.

EDIT: Alright, got it.

ksooo (changes requested):
Code change is okay. Thanks. 
Please update minor version number in addon.xml.in and changelog.txt, then we can release a new add-on version.

ksooo (approved):
Thanks for your contribution.

PR #677, opened 2024-08-23
Add support for PVR Providers (HTSPv38+)

ksooo:
Requires a decent tvheadend build, at least [267aef151ec30fa9c5469500100ab7f59092d39a](https://github.com/tvheadend/tvheadend/commit/267aef151ec30fa9c5469500100ab7f59092d39a) (merged Aug 23rd, 2024)


src/Tvheadend.cpp
```diff
@@ -223,6 +224,50 @@ bool CTvheadend::HasStreamingProfile(const std::string& streamingProfile) const
...
+    {
+      kodi::addon::PVRProvider provider;
+      provider.SetUniqueId(entry.second.GetId());
+      provider.SetName(entry.second.GetName());
+      //SetType
+      //SetIconPath
+      //SetCountries
+      //SetLanguages
+
+      providers.emplace_back(std::move(provider));
```
Jalle19, line 254: Why not `std::move()` directly to `results`?

src/Tvheadend.cpp
```diff
@@ -2256,10 +2348,29 @@ void CTvheadend::ParseChannelDelete(htsmsg_t* msg)
...
+  const auto it = m_channels.find(u32);
+  if (it != m_channels.cend())
+    providerUid = (*it).second.GetProviderUid();
+
+  /* Erase channel */
   m_channels.erase(u32);
   m_channelTuningPredictor.RemoveChannel(u32);
   TriggerChannelUpdate();
+
+  if (providerUid > 0)
```
Jalle19: Should this be `providerUid != PVR_PROVIDER_INVALID_UID`?

src/Tvheadend.cpp
```diff
@@ -2224,6 +2291,31 @@ void CTvheadend::ParseChannelAddOrUpdate(htsmsg_t* msg, bool bAdd)
       /* CAID */
       if (caid == 0)
         htsmsg_get_u32(&f->hmf_msg, "caid", &caid);
+
+      /* Service provider */
+      str = htsmsg_get_str(&f->hmf_msg, "providername");
+      if (str)
+      {
+        const uint32_t uid{utilities::hash_str_uint32(str)};
```
Jalle19: Empty provider names should probably be ignored?

src/Tvheadend.cpp
ksooo: `results` is not a simple container. `Add` actually calls back to Kodi to immediately deliver the result. We use the concept "collect results mutex ptrotected, then release mutex and deliver results to Kodi" everywhere in the add-on, also for channels, groups, ...

src/Tvheadend.cpp
ksooo: Yep. Thanks.

src/Tvheadend.cpp
ksooo: Yep. Thanks.

src/Tvheadend.cpp
ksooo: Wait! PVR_PROVIDER_INVALID_UID is signed and defined as -1. providerUid is unsigned.

ksooo: Will change type accordingly.

ksooo: done

src/Tvheadend.cpp
ksooo: done

src/Tvheadend.cpp
Jalle19: Right :+1:

src/Tvheadend.cpp
```diff
@@ -2256,10 +2344,29 @@ void CTvheadend::ParseChannelDelete(htsmsg_t* msg)
   }
   Logger::Log(LogLevel::LEVEL_DEBUG, "delete channel %u", u32);
 
-  /* Erase */
+  /* We need to check and update providers. May be the deleted channel is the last channel with this provider */
+  int32_t providerUid{0};
```
Jalle19: Should this too be `PVR_PROVIDER_INVALID_UID`?

ksooo: Yes. I will fix that.

ksooo: Done.

Jalle19 (approved):
Nice feature, thanks!

PR #688, opened 2024-10-22
PVR Add-on API 9.2.0 (multiple recorded streams, title extra info).

ksooo:
Requires https://github.com/xbmc/xbmc/pull/25867 to be merged.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -36,14 +36,12 @@ CTvheadend::CTvheadend(const kodi::addon::IInstanceInfo& instance)
     m_customTimerProps(
         {CUSTOM_PROP_ID_DVR_CONFIGURATION, CUSTOM_PROP_ID_DVR_COMMENT}, *m_conn, m_dvrConfigs),
     m_streamchange(false),
-    m_vfs(new HTSPVFS(m_settings, *m_conn)),
     m_queue(static_cast<size_t>(-1)),
     m_asyncState(m_settings->GetResponseTimeout()),
     m_timeRecordings(*m_conn, m_dvrConfigs),
     m_autoRecordings(m_settings, *m_conn, m_dvrConfigs),
     m_epgMaxDays(EpgMaxFutureDays()),
-    m_playingLiveStream(false),
-    m_playingRecording(nullptr)
+    m_playingLiveStream(false)
 {
   m_dmx.reserve(m_settings->GetTotalTuners());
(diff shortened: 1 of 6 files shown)


src/Tvheadend.cpp
```diff
@@ -1723,54 +1728,128 @@ PVR_ERROR CTvheadend::OnSystemWake()
...
 {
   if (!m_asyncState.WaitForState(ASYNC_EPG))
+    return PVR_ERROR_SERVER_TIMEOUT;
+
+  const auto vfs{std::make_shared<HTSPVFS>(m_settings, *m_conn)};
+  if (!vfs->Open(recording))
     return false;
 
-  bool ret = m_vfs->Open(rec);
+  streamId = vfs->GetFileId();
```
Jalle19, line 1740: Since the map key (`streamId`) is deduced from the map value (`vfs`), would it be simpler to store a `std::vector<std::shared_ptr<tvheadend::HTSPVFS>>` and find the correct instance with a lambda?

src/Tvheadend.h
```diff
@@ -254,12 +246,14 @@ class ATTR_DLL_LOCAL CTvheadend : public kodi::addon::CInstancePVRClient,
   /*
    * VFS (pass-thru)
```
Jalle19: "pass-thru" isn't really valid anymore, could be removed

src/Tvheadend.cpp
ksooo: What do you mean exactly? Not sure I understand, sorry.

src/Tvheadend.h
ksooo: will do

src/Tvheadend.cpp
ksooo: Now I got it. I added the streamId later to HTSPVFS... Anyway, wouldn't this have a negative performance impact (linear seek on read of every chunk, for example)?

Jalle19: In theory it will probably be slower, but for the most part there will only be one item in the vector/map so it shouldn't matter much (if I understand correctly, the only time there will be multiple streams read is when Kodi is generating thumbnails for a list of recordings).

If changing to a vector doesn't simplify the code then I don't think it's worth changing what you have now.

ksooo: Fair enough. Anyway, i think I will go with what I have now.

ksooo:
jenkins build this please

PR #703, opened 2025-03-16
Fix async state

ksooo:
Partly reverts https://github.com/kodi-pvr/pvr.hts/pull/615/commits/b08c11b831c76e1de57f220955816dbe7bf5d581 as this caused problems: https://forum.kodi.tv/showthread.php?tid=381259.


lib/libhts/hts_strtab.h
```diff
@@ -1,6 +1,6 @@
 /*
  *  tvheadend
- *  Copyright (C) 2007 Andreas �man
+ *  Copyright (C) 2007 Andreas �man
```
Jalle19: This looks wrong, at least in my browser ( `...man`)

src/tvheadend/InstanceSettings.cpp
```diff
@@ -39,7 +39,7 @@ const int DEFAULT_DVR_DUPDETECT = DVR_AUTOREC_RECORD_ALL;
 const bool DEFAULT_DVR_PLAYSTATUS = true;
 const int DEFAULT_STREAM_CHUNKSIZE = 64; // KB
 const bool DEFAULT_DVR_IGNORE_DUPLICATE_SCHEDULES = true;
-const bool DEFAULT_STREAM_STALLED_THRESHOLD = 10; // seconds
+const int DEFAULT_STREAM_STALLED_THRESHOLD = 10; // seconds
```
Jalle19, line 42: Has this ever worked correctly?

Jalle19 (approved):

src/tvheadend/InstanceSettings.cpp
ksooo: Not really. But without serious functional consequences. Default was 1, not 10 as intended.

lib/libhts/hts_strtab.h
ksooo: Same here. Was just a try, with the last commit, to deal with:

```
C:\jenkws\workspace\binary-addons\kodi-windows-i686-Piers\tools\depends\target\binary-addons\pvr.hts\lib\libhts\htsbuf.c(1): warning C4828: The file contains a character starting at offset 0x44 that is illegal in the current source character set (codepage 65001).
```

ksooo: I guess I will withdraw those codepage related fixes from the PR. It's comments only, no functional change.

Issue #656, opened 2024-03-03
cstdint needs to be included in HTSPVFS.h


dsiggi:
Hi,
whenn I try to build for Matrix 19.5 I get these errors.

```
[ 60%] Building CXX object CMakeFiles/pvr.hts.dir/src/tvheadend/HTSPDemuxer.cpp.o
[ 63%] Building CXX object CMakeFiles/pvr.hts.dir/src/tvheadend/HTSPVFS.cpp.o
In Datei, eingebunden von /usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp:8:
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.h:52:3: Fehler: »uint32_t« bezeichnet keinen Typ
   52 |   uint32_t m_fileId;
      |   ^~~~~~~~
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.h:11:1: Anmerkung: »uint32_t« ist im Header »<cstdint>« definiert; haben Sie vergessen, »#include <cstdint>« zu schreiben?
   10 | #include <string>
  +++ |+#include <cstdint>
   11 | 
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp: In Konstruktor »tvheadend::HTSPVFS::HTSPVFS(tvheadend::HTSPConnection&)«:
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp:35:5: Fehler: Klasse »tvheadend::HTSPVFS« hat keinen Feldnamen »m_fileId«
   35 |     m_fileId(0),
      |     ^~~~~~~~
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp: In Elementfunktion »void tvheadend::HTSPVFS::RebuildState()«:
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp:51:7: Fehler: »m_fileId« wurde in diesem Gültigkeitsbereich nicht definiert
   51 |   if (m_fileId != 0)
      |       ^~~~~~~~
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp: In Elementfunktion »void tvheadend::HTSPVFS::Close()«:
/usr/src/pvr.hts/src/tvheadend/HTSPVFS.cpp:88:7: Fehler: »m_fileId« wurde in diesem Gültigkeitsbereich nicht definiert
   88 |   if (m_fileId != 0)
      |       ^~~~~~~~
[...]

PR #728, opened 2025-12-03
Fix compilation in armv7 with gcc15

hbiyik:
Fixes compilation in armv7 Kodi Omega / pvr.hts Omega branch (both on latest)

Full log:
[log.txt](https://github.com/user-attachments/files/23898276/log.txt)

Same issue in Piers as well, @ksooo @garbear 

Fixes: https://github.com/kodi-pvr/pvr.hts/issues/656

### DIFF
--- a/src/tvheadend/HTSPVFS.h
+++ b/src/tvheadend/HTSPVFS.h
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <memory>
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


garbear (approved):

ksooo:
> Same issue in Piers as well

@hbiyik could you please open a PR also for Piers? Thanks.

PR #729, opened 2025-12-05
Fix compilation in with gcc15

hbiyik:
noticed that this is also a problem on x86 amd64, here is the piers pr


hbiyik:
@garbear @ksooo pining just in case things gets lost

PR #736, opened 2026-01-06
Fix hang of Kodi on unpause

ksooo:


ksooo:
jenkins build this please

ksooo:
Runtime-tested for couple of weeks. Works as designed, hang is fixed.

PR #745, opened 2026-07-12
New setting: Use backend settings for recording folder creation

### LINKED ISSUE #740: TVheadend HTSP Client Timer Settings - Defaults for new timers to include 'Folders' to allow Folder to be blank
Hi team.  Thanks for all your brilliant work.
I just have a feature request to solve what is probably an unusual problem...

When I add a timer, the 'Folder' option is preconfigured with the programme Title.

Image

This is normally fine but UK TV Guides regularly use the colon (:) character in the title, as in the example above. 
TVHeadend does a great job of stripping out illegal characters when creating folders and filenames but this 'Folder' field in the client overrides all the settings in TVHeadend.

This results in a folder being created with a character unrecognisable to TVHeadend itself and thus the recording being seen as 'Removed - File Missing'. 

My workaround is to remove  the content of the 'Folder' field when creating a new timer but I often forget to do this.  (Removing the content ensures TVheadends own folder settings are used and this ensures the character substitution takes place to remove the illegal character).

Please can we have an option in the PVR Addon Timer Settings which adds a new default for new timers to essentially make the 'Folder' entry blank.

Image

Many thanks
John

ksooo:
Fixes #740 

> What would work is a client setting that disabled the recording folder timer settings field completely, resulting in what you want, server settings being used, but in case you want to intentionally override server side behavior you will not be able to do so (due to no input field for the name being available)

I decided to go with this approach. 

Introduces a new add-on setting to control recording folder creation. 

If deactivated, behavior is unchanged. Kodi will create a folder name and send it to the backend when a timer rule is created/edited. User can change the folder name using the respective text input field in the timer rule settings.

If activated, timer rule settings will not contain above mentioned text input field for recording folder name. folders will be created according to the backend's recording profile settings.

### DIFF
--- a/src/Tvheadend.cpp
+++ b/src/Tvheadend.cpp
@@ -39,7 +39,7 @@ CTvheadend::CTvheadend(const kodi::addon::IInstanceInfo& instance)
     m_streamchange(false),
     m_queue(static_cast<size_t>(-1)),
     m_asyncState(m_settings->GetResponseTimeout()),
-    m_timeRecordings(*m_conn, m_dvrConfigs),
+    m_timeRecordings(m_settings, *m_conn, m_dvrConfigs),
     m_autoRecordings(m_settings, *m_conn, m_dvrConfigs),
     m_epgMaxDays(EpgMaxFutureDays()),
     m_playingLiveStream(false)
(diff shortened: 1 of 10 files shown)


src/Tvheadend.cpp
```diff
@@ -1096,7 +1096,10 @@ PVR_ERROR CTvheadend::GetTimerTypes(std::vector<kodi::addon::PVRTimerType>& type
           PVR_TIMER_TYPE_SUPPORTS_ENABLE_DISABLE | PVR_TIMER_TYPE_SUPPORTS_CHANNELS |
           PVR_TIMER_TYPE_SUPPORTS_START_TIME | PVR_TIMER_TYPE_SUPPORTS_END_TIME |
           PVR_TIMER_TYPE_SUPPORTS_WEEKDAYS | PVR_TIMER_TYPE_SUPPORTS_PRIORITY |
-          PVR_TIMER_TYPE_SUPPORTS_LIFETIME | PVR_TIMER_TYPE_SUPPORTS_RECORDING_FOLDERS,
+          PVR_TIMER_TYPE_SUPPORTS_LIFETIME |
+          (m_settings->GetDvrUseBackendFolderCreation()
+               ? 0
+               : PVR_TIMER_TYPE_SUPPORTS_RECORDING_FOLDERS),
```
Jalle19, line 1102: Maybe it would be better to always announce this capability even if we wouldn't be sending the directory to the backend? E.g. what happens to existing timers with a recording directory specified if the user changes the setting?

ksooo: > Maybe it would be better to always announce this capability even if we wouldn't be sending the directory to the backend? 

Always announcing the capability means Kodi shows an editable folder field whose input is silently discarded — that's worse UX than hiding it, and it defeats the point of the setting (the field being gone is the feature. 

> what happens to existing timers with a recording directory specified if the user changes the setting?

 They keep existing directory settings. The only drawback is that the user cannot use pvr.hts GUI to change custom recording directory names anymore.

@Jalle19 is that acceptable?

Jalle19: Thanks for the explanation!

Jalle19 (approved):

johnandally:
Thanks Kai .  I really appreciate it.  You are a star. :)

PR #752, opened 2026-09-10
HTSPDemuxer: re-announce stream change on subscription recovery

smp79:
### Problem

When a live channel's signal drops and recovers (`subscriptionStatus` reporting `badSignal` and then clearing), Tvheadend does **not** send a new`subscriptionStart` — it's the same HTSP subscription throughout. If Kodi had dropped the video stream while the signal was out, **it may stay stuck in audio-only (Radio) mode indefinitely** after the signal comes back, because nothing tells it a video stream may be available again.

### Fix

`Subscription::ParseSubscriptionStatus()` already tracks the transition between error states (`SUBSCRIPTION_NOSIGNAL`, `SUBSCRIPTION_SCRAMBLED`, etc.) and `SUBSCRIPTION_RUNNING`, but only updates internal state. This adds a thin
wrapper, `HTSPDemuxer::ProcessSubscriptionStatus()`, that detects the error → `SUBSCRIPTION_RUNNING` transition and pushes the same `DEMUX_SPECIALID_STREAMCHANGE` marker that a real channel switch already sends via `ParseSubscriptionStart()`. This routes through Kodi's stream-change handling in `CVideoPlayer::Process()`, which re-queries the current stream set and reselects defaults — recovering video if it's actually present again.

### DIFF
--- a/src/tvheadend/HTSPDemuxer.cpp
+++ b/src/tvheadend/HTSPDemuxer.cpp
@@ -447,13 +447,31 @@ void HTSPDemuxer::ResetStatus(bool resetSubscriptionData /* = true */)
  * Parse incoming data
  * *************************************************************************/
 
+void HTSPDemuxer::ProcessSubscriptionStatus(htsmsg_t* m)
+{
+  const eSubsriptionState prevState = m_subscription.GetState();
+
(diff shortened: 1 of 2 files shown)


ksooo (approved):
LGTM, thanks

